PR #17, opened 2025-10-13
feat(contoller/views): validate that all handlers have `@rest` decorator

### LINKED ISSUE #16: Validate that all handlers have `@rest` decorator
https://github.com/wemake-services/django-modern-rest/blob/86f9f9934bfd50e85295d53dac4c18e91a4863ee/django_modern_rest/controller.py#L94

Peopl3s:
### 1. Classification of validation  in `django_modern_rest/controller.py `:

- Added logic in the method of "existing" http_methods"to improve the work of the doctor"@rest`
- Validation checks `__endpoint__`, which is the editor of `@rest`
- When creating the editor, a clear value is selected: `"Handler {Class name}.{method}() must be decorated with the @rest"` symbol
- Validation only works for advanced methods that play for Django View

### 2. Added missing `@rest`decorators in all files:

- `UserCreateController.post()` - added `@rest(return_type=_UserOutput)`
- `UserListController.get()` - added `@rest(return_type=list[_UserInput])`
- `UserUpdateController.patch()` - added `@rest(return_type=_UserInput)`
- `AsyncController.get()` - added `@rest(return_type=str)`
- `SyncController.post()` - added `@rest(return_type=str)`

Closes #16

### DIFF
--- a/tests/test_unit/test_controller/test_rest_decorator_validation.py
+++ b/tests/test_unit/test_controller/test_rest_decorator_validation.py
@@ -1,0 +1,90 @@
+from typing import final
+
+import pytest
+
+from django_modern_rest import Controller, rest
+from django_modern_rest.plugins.pydantic import PydanticSerializer
+
+
+@final
+class ValidController(Controller[PydanticSerializer]):
+    @rest(return_type=str)
+    def get(self) -> str:
+        return 'success'
+
+    @rest(return_type=dict)
+    def post(self) -> dict:
+        return {'data': 'success'}
+
+
+def test_invalid_controller_raises_validation_error() -> None:
+    pattern = r'Handler InvalidController\.get\(\) must be decorated with @rest'
+    with pytest.raises(ValueError, match=pattern):
+
+        @final
+        class InvalidController(Controller[PydanticSerializer]):
+            def get(self) -> str:
+                return 'success'
+
+            @rest(return_type=dict)
+            def post(self) -> dict:
+                return {'data': 'success'}
+
+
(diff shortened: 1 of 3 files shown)


django_modern_rest/openapi/views.py
```diff
@@ -48,6 +47,12 @@ def as_view(  # type: ignore[override]
 
     @classmethod
     def _default_config(cls) -> OpenAPIConfig:
+        # Local import to avoid circular import
```
kondratevdev: Describe in more detail in which cases you received circular import.

Peopl3s: Just run tests

`poetry run python -m pytest tests/test_integration/ -v --tb=
short`

django_modern_rest/controller.py
```diff
@@ -91,9 +91,15 @@ def dispatch(
...
-            for method in cls.http_method_names
-            if getattr(cls, method, None) is not None
-        }
+        implemented_methods = set(cls.http_method_names) & vars(cls).keys()
+
+        # Validate that all handlers have `@rest` decorator
+        for method in implemented_methods:
+            method_handler = getattr(cls, method)
+            if not hasattr(method_handler, '__endpoint__'):
+                class_name = getattr(cls, '__name__', str(cls))
```
kondratevdev, line 100: Can we have a case where` __name__` is not defined in a class?

Peopl3s: @kondratevdev No, __name__ is always defined for classes in Python. It's a fundamental attribute that Python automatically sets when a class is created.

PR #37, opened 2025-10-14
DRAFT combine controllerparser models

milssky:
I had to implement raw data extraction in the components, I haven't come up with a better one yet.

### DIFF
--- a/django_modern_rest/controller.py
+++ b/django_modern_rest/controller.py
@@ -1,5 +1,5 @@
 from http import HTTPStatus
-from typing import Any, ClassVar, Generic, TypeAlias, TypeVar
+from typing import Any, ClassVar, Generic, TypeAlias, TypeVar, get_args
 
 from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpRequest, HttpResponse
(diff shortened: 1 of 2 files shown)


django_modern_rest/controller.py
```diff
@@ -150,25 +145,32 @@ def _handle_request(
         if endpoint is not None:
```
kondratevdev, line 165: Suggest to use early return

```python
if endpoint is None:
    raise MethodNotAllowedError
```

django_modern_rest/serialization.py
```diff
@@ -77,7 +81,115 @@ def deserialize_hook(
...
+    def _build_component_specs(self) -> dict[str, type[Any]]:
+        specs = {}
+        for component in self.components:
+            type_args = get_args(component)
+            if type_args:
+                context_name = component._provide_context_name()  # noqa: SLF001
+                specs[context_name] = type_args[0]
+        return specs
+
+    def _get_or_create_combined_model(
```
kondratevdev: Can there be a situation when the model is there initially?

sobolevn (changes requested):
I see a problem with this approach. `create_model` is a very slow function. We can't create a model on each request.

We need to create a model on import. And validate on request.

milssky:
I rewrite it in new brach over new code. 
See https://github.com/wemake-services/django-modern-rest/pull/81

PR #54, opened 2025-10-16
Implement `TypedDict` typed for `PydanticSerializer.model_dump_kwargs` class variable

Nifacy:
I ran into problem when defining dictionary fields: `pydantic` doesn't allow importing the alias `IncEx`. Therefore, I had to make a local copy of this alias `_IncEx` in the project.

### DIFF
--- a/django_modern_rest/plugins/pydantic.py
+++ b/django_modern_rest/plugins/pydantic.py
@@ -8,6 +8,7 @@ from typing import (
     TypeAlias,
     TypedDict,
     Union,
+    final,
 )
 
 try:
@@ -39,6 +40,8 @@ if TYPE_CHECKING:
     )
 
 
+# pydantic does not allow to import this,
+# so we have to duplicate this type.
 _IncEx: TypeAlias = (
     set[int]
     | set[str]
@@ -47,6 +50,7 @@ _IncEx: TypeAlias = (
 )
 
 
+@final
 class ModelDumpKwargs(TypedDict, total=False):
     """Keyword arguments for pydantic's model dump method."""
 


django_modern_rest/plugins/pydantic.py
```diff
@@ -30,6 +39,32 @@
...
 
+_IncEx: TypeAlias = (
+    set[int]
+    | set[str]
+    | Mapping[int, Union['_IncEx', bool]]
+    | Mapping[str, Union['_IncEx', bool]]
+)
+
+
+class ModelDumpKwargs(TypedDict, total=False):
```
sobolevn, line 54:
```suggestion
@final
class ModelDumpKwargs(TypedDict, total=False):
```

django_modern_rest/plugins/pydantic.py
```diff
@@ -1,5 +1,14 @@
...
 from functools import cache
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    ClassVar,
+    Literal,
+    TypeAlias,
+    TypedDict,
+    Union,
```
sobolevn, line 10:
```suggestion
    Union,
    final,
```

django_modern_rest/plugins/pydantic.py
```diff
@@ -30,6 +39,32 @@
     )
 
 
+_IncEx: TypeAlias = (
```
sobolevn, line 45:
```suggestion
# pydantic does not allow to import this,
# so we have to duplicate this type.
_IncEx: TypeAlias = (
```

sobolevn:
Please, create one more PR for the other typed-dict.

Issue #43, opened 2025-10-15
Do not depend on `msgspec` directly


sobolevn:
Right now we always install `msgspec` directly as the main dependency: https://github.com/wemake-services/django-modern-rest/blob/25513fdddc16702fcbce64578b40586e19f5addb/pyproject.toml#L49-L54

However, `msgspec` is written in C and it can not build on some platforms / setups.
Since we don't use anything from `msgspec` except the json parsing / dumping, I guess we can make it an optional dependency.

How this should be:
1. Add `plugins/msgspec/json.py` file
2. Move currect json logic there
3. Check if `msgspec` is installed. If yes, use: https://github.com/wemake-services/django-modern-rest/blob/25513fdddc16702fcbce64578b40586e19f5addb/django_modern_rest/settings.py#L17-L21
4. If not, fallback to regular `json` module (it is slow, but always exists)
5. Add docs about `[msgspec]` extra and that it is required for faster json parsing
6. Test it :)

sobolevn:
Friendly ping :)
I really need this to move forward. Do you have an ETA? :)

PR #56, opened 2025-10-16
Add `3.14t` to the test

sobolevn:


sobolevn:
ok, msgspec does not support 3.14t at the moment, exactly as I said in https://github.com/wemake-services/django-modern-rest/issues/43

Issue #68, opened 2025-10-17
Add `slotscheck` to the CI


sobolevn:
There's an amazing tool called https://github.com/ariebovenberg/slotscheck
We use `__slots__` in this project by default. So, we would love to test their defition.

List of things to do:
- Add `slotscheck` to test deps
- Add the configuration to `pyproject.toml`
- Add the call to `Makefile`

This should be a really small task for several minutes :)

F0RRZZ:
I would like to work on this

PR #71, opened 2025-10-17
Add `slotscheck` lib to the CI #68

F0RRZZ:


django_modern_rest/headers.py
```diff
@@ -5,6 +5,7 @@
 from django_modern_rest.types import Empty, EmptyObj
 
 
+@dataclasses.dataclass(frozen=True, slots=True, kw_only=True)
```
sobolevn: I don't think that this is correct. `_BaseResponseHeader` should not be constructed. I propose adding `__slots__ = ('description', 'deprecated', 'example')` instead

sobolevn (approved):
Thank you!

PR #81, opened 2025-10-18
feat: add combined model validate, refs #8

### LINKED ISSUE #8: Speed up the validation
Right now we do many individual validation steps: `headers`, `body`, `query`, `security,` etc.
We need to think about a way to create common context for the validation.

Steps:
1. Create a dummy example with several models (1, 2, 3, and 4) and measure times for creation + validation
2. Create a dummy example with one big model and measure times for creation + validation

Do it for `pydantic` and `msgspec`

If the second option is better by a lot, we need to change the logic.
We would construct one big object for validation and use it just once.

This way we would have:
- All errors at once (which is great on its own)
- Better performance
- Better debug experience

milssky:
Third try for implement task. 

Closes #8

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ version = "0.1.0"
 authors = [
   "Nikita Sobolev <mail@sobolevn.me>",
   "Alexander Kondratev <mail@kondratev.dev>",
-  # TODO: add Alexey
+  "Alexey Potapov <me@black-as-ink.ru>",
 ]
 
 readme = "README.md"


django_modern_rest/controller.py
```diff
@@ -87,6 +88,11 @@ def __init_subclass__(cls) -> None:
         }
         cls._validate_endpoints()
 
+        cls._serializer_context = SerializerContext.build_for_class(
```
sobolevn: Please, make sure that `SerializerContext` is easy to customize. Use similar pattern that we use for `Endpoint`

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
             metadata: Endpoint metadata to optimize.
 
         """
+
+
+@dataclass(slots=True, frozen=True)
```
sobolevn:
```suggestion
@final
@dataclass(slots=True, frozen=True, kw_only=True)
```

it is not meant to be subclassed.

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
+        """Collect, validate, and bind component data to the controller."""
+        context = self._collect_context(controller, request, *args, **kwargs)
+        validated = self._validate_context(context)
+        self._bind_parsed(controller, validated)
+
+    @classmethod
+    def for_controller(
```
sobolevn: Sorry, not clear why we would need this method. Controller already has `_serializer_context` prop. Just use it.

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+    @classmethod
+    def build_for_class(
+        cls,
+        controller_cls: type[Any],
+        serializer: type[BaseSerializer],
+    ) -> 'SerializerContext':
+        """Eagerly build context for a given controller and serializer."""
+        specs: list[_ComponentSpec] = []
+        names: list[str] = []
+        type_map = _build_type_map(controller_cls, serializer, specs, names)
```
sobolevn: I propose this API instead. There's no need to mutate lists we create in-place.

```suggestion
        specs, names, type_map = _build_type_map(controller_cls, serializer, specs, names)
```

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+
+        combined_name = f'_{controller_cls.__name__}ContextModel'
```
sobolevn:
```suggestion
        combined_name = f'_{controller_cls.__qualname__}@ContextModel'
```

`@` means that this name is generated.

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+        return cast(
+            dict[str, Any],
+            self._serializer.from_python(
+                context,
+                self._combined_model,
+                strict=False,
+            ),
+        )
+
+    def _try_strip_component_prefix(self, errors: list[Any]) -> list[Any]:
```
sobolevn: We don't need to do that. Any error message we produce is good enough. This happens in runtime, so please prefer speed over nice little features.

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+    names: list[str],
+) -> dict[str, Any]:
+    """Build mapping name -> model and fill specs/names lists."""
+    type_map: dict[str, Any] = {}
+    for component_cls, type_args in getattr(
+        controller_cls,
+        '_component_parsers',
+        [],
+    ):
+        data_getter = component_cls._provide_context_data  # noqa: SLF001
```
sobolevn: Make it public, since you use it from another place.

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+        model = type_args[0] if len(type_args) >= 1 else Any
```
sobolevn: This can be a footgun in the long run. Just get the `[0]` item. It is better to fail on parsing than to allow any data it.

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+        strict = getattr(component_cls, 'strict_validation', False)
+
+        name = component_cls._provide_context_name()  # noqa: SLF001
```
sobolevn: This is also public.

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+    def _validate_context(self, context: dict[str, Any]) -> dict[str, Any]:
+        """Validate the combined payload using the cached TypedDict model."""
+        try:  # noqa: WPS229
+            return self._validate_via_serializer(context)
+        except self._serializer.validation_error as exc:
+            errors = self._try_strip_component_prefix(
+                self._serializer.error_to_json(exc),
+            )
+            raise RequestSerializationError(errors) from None
+        except Exception as exc:
```
sobolevn: We can't catch all here. Just catch `self._serializer.validation_error`

django_modern_rest/serialization.py
```diff
@@ -124,3 +126,177 @@ def optimize_endpoint(cls, metadata: 'EndpointMetadata') -> None:
...
+
+    def __init__(
+        self,
+        specs: tuple[_ComponentSpec, ...],
+        serializer: type[BaseSerializer],
+        combined_model: Any,
+        names: tuple[str, ...],
+    ) -> None:
+        """Initialize the context with a compiled plan and combined model."""
+        self._specs = specs
```
sobolevn: This can be a dataclass.

django_modern_rest/components.py
```diff
@@ -135,24 +135,18 @@ class Headers(ComponentParser, Generic[_HeadersT]):
 
     parsed_headers: _HeadersT
 
+    @classmethod
     @override
-    def parse_component(
+    def _provide_context_name(cls) -> str:
```
sobolevn: This can be a attribute, not a method.

django_modern_rest/controller.py
```diff
@@ -67,6 +70,7 @@ class Controller(View, Generic[_SerializerT_co]):  # noqa: WPS214
     # Internal API:
     _component_parsers: ClassVar[list[_ComponentParserSpec]]
     _is_async: ClassVar[bool]
+    _serializer_context: ClassVar[Any]
```
sobolevn:
```suggestion
    _serializer_context: ClassVar[SerializerContext]
```

django_modern_rest/controller.py
```diff
@@ -98,6 +102,11 @@ def __init_subclass__(cls) -> None:
         }
         cls._validate_endpoints()
 
+        cls._serializer_context = cls.serializer_context_cls.build_for_class(
```
sobolevn: Please, create it before endpoints validation

sobolevn (approved):
Thanks! Great work!

Issue #85, opened 2025-10-18
Write tests for "double validation" problem


sobolevn:
We need to be sure that double validation does not happen. Ever.
See what "double validation" is: https://github.com/fastapi/fastapi/issues/3021

To be sure, we need to write tests for it.

Tests would be rather easy to write:
- One for `@modify`
- One for `@validate`
- One for raw data return
Also test disabled response validation mode to do 0 validations.

See the attached issue to learn how to make sure that double validation happened.
No matter what - we need to do a 1 or 0 validations.

F0RRZZ:
I'd like to work on it :)

sobolevn:
Thank you!

PR #101, opened 2025-10-19
Add tests for double validation problem #85

F0RRZZ:
- Issue: https://github.com/wemake-services/django-modern-rest/issues/85


sobolevn (approved):
Great stuff! Thanks a lot!

PR #127, opened 2025-10-21
fix: replace options with meta method to resolve typing errors

### LINKED ISSUE #100: `options` method is a problem, let's design it
When `options` method is defined right now on a controller:

```python
from http import HTTPStatus

from django.http import HttpResponse

from django_modern_rest import (
    Controller,
    HeaderDescription,
    ResponseDescription,
    validate,
)
from django_modern_rest.plugins.pydantic import PydanticSerializer


class _WithOptionsController(Controller[PydanticSerializer]):
    @validate(ResponseDescription(list[str], status_code=HTTPStatus.ACCEPTED))
    def get(self) -> HttpResponse:
        raise NotImplementedError

    @validate(
        ResponseDescription(
            None,
            status_code=HTTPStatus.NO_CONTENT,
            headers={'Allow': HeaderDescription()},
        ),
    )
    @override
    def options(self) -> HttpResponse:
        return self.to_response(
            None,
            status_code=HTTPStatus.NO_CONTENT,
            # You can extend this method and add any other headers
            # like `Access-Control-Allow-Origin` and others.
            headers={
                'Allow': ', '.join(
                    sorted(
                        method.upper()
                        for method in self.existing_http_methods()
                    ),
                ),
            },
        )
```

this happens:

```python
tests/test_unit/test_controllers/test_options_method.py:45: error: Signature of "options" incompatible with supertype "django.views.generic.base.View"  [override]
tests/test_unit/test_controllers/test_options_method.py:45: note:      Superclass:
tests/test_unit/test_controllers/test_options_method.py:45: note:          def options(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponseBase
tests/test_unit/test_controllers/test_options_method.py:45: note:      Subclass:
tests/test_unit/test_controllers/test_options_method.py:45: note:          def options(self) -> HttpResponse
```

This happens because `View.options` is defined already. What we can do?
1. Force users to write `# type: ignore[override]` – very ugly
2. Use another method name – ugly but ok
3. Use different way of defining `option` methods – very ugly
4. Use special decorator for it like `@meta` – ugly

Second question: how should `OPTIONS` be enabled?
1. Enabled by default
2. Disabled by default

If so, we should provide per-controller and global switches to configure `OPTIONS` avalability.
We should also handle `compose_controller` with `OPTION` method.

This is a rather big task. But, very important one.

aazmv:
## Fix: Replace `options` with `meta` method to resolve typing errors

The `options` method has incompatible signature with Django's `View.options`, causing mypy/pyright errors.

**Solution:**
- Add `meta` method as replacement for `options`
- Deprecate `options` with clear error message
- Automatic OPTIONS handler for backward compatibility

**Usage:**
```python
# Before (typing error)
def options(self) -> HttpResponse:  # ❌

# After (clean solution)  
def meta(self) -> HttpResponse:  # ✅
```

Fixes #100


django_modern_rest/validation.py
```diff
@@ -608,6 +608,18 @@ class _ValidationContext:
...
+        return HTTPMethod.OPTIONS
+
+    # Special case for internal default options handler
+    if func_name == '_default_options_handler':
+        return HTTPMethod.OPTIONS
+
+    # Special case for internal default options wrapper
+    if func_name == 'default_options_wrapper':
+        return HTTPMethod.OPTIONS
+
```
```suggestion
```

we don't need any of this :)

tests/test_unit/test_controllers/test_options_method.py
```diff
@@ -0,0 +1,139 @@
...
+        return self.to_response(
+            None,
+            status_code=HTTPStatus.NO_CONTENT,
+            headers={
+                'Allow': 'GET, OPTIONS',
+            },
+        )
+
+
+def test_options_method_typing_error() -> None:
```
sobolevn: Why is this test empty?

django_modern_rest/controller.py
```diff
@@ -112,11 +113,44 @@ def __init_subclass__(cls) -> None:
...
         cls.serializer_context = cls.serializer_context_cls(cls)
-        cls.api_endpoints = {
-            meth: cls.endpoint_cls(func, controller_cls=cls)
-            for meth in cls.existing_http_methods()
-            if (func := getattr(cls, meth)) is not getattr(View, meth, None)
-        }
+
+        # Build API endpoints
+        api_endpoints = {}
+        for meth in cls.existing_http_methods():
```
sobolevn: This does not look right. Several reasons:
- you can't compose `Controller` types with the similar methods. right now there are no tests for it, see `routing.py`
- There are no `settings` to enable / disable `OPTIONS` generation on per-controller level

django_modern_rest/controller.py
```diff
@@ -90,7 +91,7 @@ class Controller(View, Generic[_SerializerT_co]):  # noqa: WPS214
     _is_async: ClassVar[bool]
 
     @override
-    def __init_subclass__(cls) -> None:
+    def __init_subclass__(cls) -> None:  # noqa: C901
```
sobolevn:
```suggestion
    def __init_subclass__(cls) -> None:
```

Please, do not ever ignore complexity errors, refactor instead.

django_modern_rest/controller.py
```diff
@@ -112,11 +113,44 @@ def __init_subclass__(cls) -> None:
...
+        for meth in cls.existing_http_methods():
+            func = getattr(cls, meth, None)
+            if func is not None and func is not getattr(View, meth, None):
+                # Skip our deprecated options method
+                if meth == 'options' and func is cls.options:
+                    continue
+                api_endpoints[meth] = cls.endpoint_cls(func, controller_cls=cls)
+
+        # Special handling for meta method -> options mapping
+        meta_func = getattr(cls, 'meta', None)
```
sobolevn: I would just suggest to change the function name here from `meta` to `options` and pass it as-is further. Everything should just work.

sobolevn (changes requested):

PR #147, opened 2025-10-23
Work on docs

### LINKED ISSUE #138: Provide usage examples for all components
Several components have not-so-great docs with some missing examples: 

https://github.com/wemake-services/django-modern-rest/blob/489038ff0f17db815f69e9662169cd95b3c11a48/django_modern_rest/components.py#L102-L110

https://github.com/wemake-services/django-modern-rest/blob/489038ff0f17db815f69e9662169cd95b3c11a48/django_modern_rest/components.py#L140-L148

https://github.com/wemake-services/django-modern-rest/blob/489038ff0f17db815f69e9662169cd95b3c11a48/django_modern_rest/components.py#L165-L170

We need to add examples on how to use these components.

wimble3:
# Work on docs
I just tried to write some examples for components, literally i dont know, how much right it is, i saw this project one week ago and i dont understand so many things, butttt i should start from something =)

## Checklist

## Related issues
- Closes #138

### DIFF
--- a/django_modern_rest/components.py
+++ b/django_modern_rest/components.py
@@ -244,7 +244,7 @@ class Path(ComponentParser, Generic[_PathT]):
         ...     ),
         ... ]
 
-    Will parse a url path like ``/user_id/100`` 
+    Will parse a url path like ``/user_id/100``
     which will be translated into ``{'user_id': 100}``
     into ``UserPath`` model.
 


django_modern_rest/components.py
```diff
@@ -166,7 +186,15 @@ class Path(ComponentParser, Generic[_PathT]):
     """
     Parses the url part of the request.
 
-    # TODO: example
+    For example:
```
sobolevn, line 216:
```suggestion
    For example:

```

django_modern_rest/components.py
```diff
@@ -166,7 +186,15 @@ class Path(ComponentParser, Generic[_PathT]):
...
+    .. code:: python
+
+        >>> import pydantic
+
+        >>> class UserPath(pydantic.BaseModel):
```
sobolevn, line 225: Please, provide `path()` call as well.

wimble3:
@sobolevn okay, I was relying on the way Query is described now, is it worth adding an example of how to add Query to the controller too?

django_modern_rest/components.py
wimble3: @sobolevn Im so sorry, can you explain, do you wanna see something like this?
```python
class Query(ComponentParser, Generic[_QueryT]):
    """
    Parses query params of the request.

    For example:

    .. code:: python

        >>> import pydantic
        >>> from django_modern_rest import Query, Controller
        >>> from django_modern_rest.plugins.pydantic import PydanticSerializer
        >>> from django_modern_rest import Router
        >>> from django.urls import path, include

        >>> class ProductQuery(pydantic.BaseModel):
        ...     category: str
        ...     reversed: bool

        >>> class ProductListController(
        ...     Query[ProductQuery],
        ...     Controller[PydanticSerializer],
        ... ): ...

        >>> router = Router([
        ...     path(
        ...         'product/',
        ...         ProductListController.as_view(),
        ...         name='products',
        ...     ),
        ... ])

        >>> urlpatterns = [
        ...     path(
        ...         'api/', include((router.urls, 'rest_app'), namespace='api')
        ...     ),
        ... ]

    Will parse a request like ``?category=cars&reversed=true``
    into ``ProductQuery`` model.

    If your controller class inherits from ``Query`` - then you can access
    parsed query model as ``self.parsed_query`` attribute.
    """
```

sobolevn: Yes, but only for `Path`, not for `Query` :)

wimble3: Copy that

django_modern_rest/components.py
```diff
@@ -166,7 +213,43 @@ class Path(ComponentParser, Generic[_PathT]):
...
+        ...     user_id: int
+
+        >>> class UserUpdateInput(pydantic.BaseModel):
+        ...     email: str
+        ...     age: int
+
+        >>> class UserUpdateController(
+        ...     Body[UserUpdateInput],
```
sobolevn:
```suggestion
```

I propose to remove parts that show other components.

django_modern_rest/components.py
```diff
@@ -166,7 +213,43 @@ class Path(ComponentParser, Generic[_PathT]):
...
+        >>> from django_modern_rest import Body, Path, Controller, Router
+        >>> from django_modern_rest.plugins.pydantic import PydanticSerializer
+        >>> from django.urls import path, include
+
+        >>> class UserPath(pydantic.BaseModel):
+        ...     user_id: int
+
+        >>> class UserUpdateInput(pydantic.BaseModel):
+        ...     email: str
+        ...     age: int
```
sobolevn:
```suggestion
```

django_modern_rest/components.py
```diff
@@ -166,7 +213,43 @@ class Path(ComponentParser, Generic[_PathT]):
...
+        ...     ),
+        ... ])
+
+        >>> urlpatterns = [
+        ...     path(
+        ...         'api/', include((router.urls, 'rest_app'), namespace='api')
+        ...     ),
+        ... ]
+
+    Will parse a url path like ``/user_id/100`` into ``UserPath`` model.
```
sobolevn:
```suggestion
    Will parse a url path like ``/user_id/100`` 
    which will be translated into ``{'user_id': 100}``
    into ``UserPath`` model.
```

PR #155, opened 2025-10-23
Add QUERY HTTP method support

### LINKED ISSUE #140: Test `query` HTTP method support
Recently I added https://github.com/wemake-services/django-modern-rest/blob/489038ff0f17db815f69e9662169cd95b3c11a48/django_modern_rest/controller.py#L91-L93 which does not have `query` method in it. Docs: https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html
TLDR: it is not official yet, but still can be used.

We need to test that adding `'query'` to `http_methods` class variable and providing `def query` method will work.

acostyle:
This PR adds support for the `QUERY` HTTP method in Controllers. The `QUERY` method is defined in the HTTP Working Group draft specification (https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html).

## Changes made
- Controllers can now define a `query()` method to handle QUERY requests
- Added test coverage for the QUERY method support

## Checklist

- [ ] I have updated the documentation for the changes I have made (not needed for this change)

## Related issues
- Closes #140

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -144,6 +144,7 @@ html_theme_options = {
     'accent_color': 'green',
     'light_logo': '_static/images/logo-light.svg',
     'dark_logo': '_static/images/logo-dark.svg',
+    'og_image_url': 'https://repository-images.githubusercontent.com/1072817092/f0ab70e3-c165-485b-b591-e860c16f7c4f',
 }
 
 html_context = {
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,6 +1,15 @@
 :layout: landing
 
-django-modern-rest
+.. figure:: /_static/images/logo-light.svg
+   :figclass: light-only
+   :width: 700
+   :align: center
+
+.. figure:: /_static/images/logo-dark.svg
+   :figclass: dark-only
+   :width: 700
+   :align: center
+
 ==================
 
 .. container:: badges


django_modern_rest/controller.py
```diff
@@ -89,7 +89,8 @@ class Controller(View, Generic[_SerializerT_co]):  # noqa: WPS214
     responses: ClassVar[list[ResponseDescription]] = []
     responses_from_components: ClassVar[bool] = True
     http_methods: ClassVar[frozenset[str]] = frozenset(
-        {method.name.lower() for method in HTTPMethod} - {'options'} | {'meta'},
+        {method.name.lower() for method in HTTPMethod} - {'options'}
```
sobolevn: No, sorry, we don't want to support it by default just yet. But, we want to provide this option for users.

tests/test_unit/test_controllers/test_query_method.py
```diff
@@ -0,0 +1,28 @@
...
+from django_modern_rest import (
+    Controller,
+    ResponseDescription,
+    validate,
+)
+from django_modern_rest.plugins.pydantic import PydanticSerializer
+from django_modern_rest.test import DMRRequestFactory
+
+
+class _QueryController(Controller[PydanticSerializer]):
```
sobolevn, line 15:
```suggestion
class _QueryController(Controller[PydanticSerializer]):
    http_methods = frozenset((*Controller.http_methods, 'query'))
```

that's all we need :)
Please, remove other non-test changes.

acostyle:
> Thanks a lot, I now see some major pain points for this implementation, I will take this PR from here :)

Thank you for your feedback!
If you need my assistance, please let me know. I'm here to help

PR #161, opened 2025-10-24
Add mixin validations for Controller

### LINKED ISSUE #141: Validate that `MetaMixin` can only be used with sync controller and the opposite
We have a TODO task: https://github.com/wemake-services/django-modern-rest/blob/489038ff0f17db815f69e9662169cd95b3c11a48/django_modern_rest/validation.py#L233-L237

It should not be possible to create

```python
class My(MetaMixin, Controller[PydanticSerializer]):
    async def get(self) -> int: ...
```

and

```python
class My(AsyncMetaMixin, Controller[PydanticSerializer]):
    def get(self) -> int: ...
```

Please, add a validation for this.

AlexanderZharyuk:
# I have made things!

I have added validation for controller mixins. In my implementation, the code checks that:
- a sync controller does not include AsyncMetaMixin,
- an async controller does not include MetaMixin,
- and a controller cannot include both at the same time.

To implement this, I introduced two new base classes: `BaseMeta` and `BaseAsyncMeta` in `types.py` module for two reasons:
1. Circular imports: It is not possible to import `MetaMixin` and `AsyncMetaMixin` into `validation.py` cleanly, except at the function level, which leads to the next issue.
2. Layered architecture contract: Importing `MetaMixin` and `AsyncMetaMixin` from `options_mixins.py` would break the existing layered architecture.

Therefore, I decided to introduce a couple of lightweight base entities to serve as markers.
I’m not entirely sure about this design choice, so I’d be happy to hear your feedback!

## Checklist

## Related issues
Closes #141 

Also, I’d like to note that the `_validate_endpoints(controller)` function already validates the mixin logic indirectly - meaning it’s currently impossible to trigger a validation error for a sync controller with an async mixin (or vice versa), since such cases are caught earlier at the endpoint validation stage.

However, I’ve still added dedicated tests to cover this behavior explicitly, in case the internal validation logic changes in the future.

### DIFF
--- a/django_modern_rest/validation.py
+++ b/django_modern_rest/validation.py
@@ -45,8 +45,6 @@ from django_modern_rest.settings import (
     resolve_setting,
 )
 from django_modern_rest.types import (
-    BaseAsyncMeta,
-    BaseMeta,
     Empty,
     EmptyObj,
     infer_bases,
@@ -247,29 +245,21 @@ class ControllerValidator:
         *,
         is_async: bool = False,
     ) -> None:
-        if not controller.api_endpoints:
-            return
-
-        is_subclass_of_meta_mixin = issubclass(controller, BaseMeta)
-        is_subclass_of_async_meta_mixin = issubclass(
-            controller,
-            BaseAsyncMeta,
+        from django_modern_rest.options_mixins import (  # noqa: PLC0415
+            AsyncMetaMixin,
+            MetaMixin,
         )
-        err_header = f'Controller {controller} contains incompatible mixins'
-        if is_subclass_of_meta_mixin and is_subclass_of_async_meta_mixin:
-            raise EndpointMetadataError(
-                f'{err_header}. Use only MetaMixin or AsyncMetaMixin, '
-                f'not both.',
-            )
 
-        if is_async and is_subclass_of_meta_mixin:
-            raise EndpointMetadataError(
-                f'{err_header}. Use AsyncMetaMixin in async Controller.',
+        if (
+            issubclass(controller, MetaMixin)
+            and issubclass(controller, AsyncMetaMixin)  # type: ignore[unreachable]
+        ):
+            suggestion = (  # type: ignore[unreachable]
+                'AsyncMetaMixin' if is_async else 'MetaMixin'
             )
-
-        if not is_async and is_subclass_of_async_meta_mixin:
             raise EndpointMetadataError(
-                f'{err_header}. Use MetaMixin in sync Controller.',
+                f'Use only {suggestion!r}, '
+                f'not both meta mixins in {controller!r}',
             )
 
     def _validate_components(
--- a/tests/test_unit/test_controllers/test_controller_validation.py
+++ b/tests/test_unit/test_controllers/test_controller_validation.py
@@ -65,7 +65,7 @@ def test_controller_have_either_mixins() -> None:
     """Ensure that controllers does not have both mixins."""
     with pytest.raises(
         EndpointMetadataError,
-        match='incompatible mixin',
+        match="'AsyncMetaMixin'",
     ):
 
         class _MixedController(  # type: ignore[misc]
@@ -74,4 +74,17 @@ def test_controller_have_either_mixins() -> None:
             Controller[PydanticSerializer],
         ):
             async def post(self) -> list[str]:
+                raise NotImplementedError
+
+    with pytest.raises(
+        EndpointMetadataError,
+        match="'MetaMixin'",
+    ):
+
+        class _MixedController2(  # type: ignore[misc]
+            MetaMixin,
+            AsyncMetaMixin,
+            Controller[PydanticSerializer],
+        ):
+            def post(self) -> list[str]:
                 raise NotImplementedError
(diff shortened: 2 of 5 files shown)


django_modern_rest/validation.py
```diff
@@ -234,10 +236,40 @@ class ControllerValidator:
...
+
+        if has_async_mixin and has_sync_mixin:
+            raise EndpointMetadataError(
+                f'Controller {controller} can only have one of '
+                f'MetaMixin or AsyncMetaMixin, not both.',
+            )
+
+        async_controller_with_sync_mixin = is_async and has_sync_mixin
+        sync_controller_with_async_mixin = not is_async and has_async_mixin
+        if async_controller_with_sync_mixin or sync_controller_with_async_mixin:
```
```suggestion
        if (
            (is_async and issubclass(controller, MetaMixin))
            or (not is_async and issubclass(controller, AsyncMetaMixin)
        ):
```

tests/test_unit/test_controllers/test_controller_validation.py
```diff
@@ -54,3 +60,77 @@ class _MixedController(Controller[PydanticSerializer]):
 
             def get(self) -> str:  # needs at least one endpoint to validate
                 raise NotImplementedError
+
+
+def test_controller_have_either_mixins() -> None:
+    """Ensure that controllers does not have both mixins."""
+    with pytest.raises(
+        EndpointMetadataError,
+        match='can only have one of',
```
sobolevn:
```suggestion
        match='incompatible mixin',
```

tests/test_unit/test_controllers/test_controller_validation.py
```diff
@@ -54,3 +60,77 @@ class _MixedController(Controller[PydanticSerializer]):
...
+        class _MixedController(
+            AsyncMetaMixin,
+            MetaMixin,
+            Controller[PydanticSerializer],
+        ):
+            async def post(self) -> list[str]:
+                raise NotImplementedError
+
+
+def test_controller_have_valid_mixins() -> None:
```
sobolevn: We already have tests for that. You can remove this test.

tests/test_unit/test_controllers/test_controller_validation.py
```diff
@@ -54,3 +60,77 @@ class _MixedController(Controller[PydanticSerializer]):
...
+        AsyncMetaMixin,
+        Controller[PydanticSerializer],
+    ):
+        async def post(self) -> list[str]:
+            raise NotImplementedError
+
+
+def test_sync_controller_have_invalid_mixin(monkeypatch) -> None:
+    """Ensure that sync controller have an invalid mixin."""
+    monkeypatch.setattr(
```
sobolevn: Why do you need this? Please, remove this code.

tests/test_unit/test_controllers/test_controller_validation.py
```diff
@@ -54,3 +60,77 @@ class _MixedController(Controller[PydanticSerializer]):
...
+        ControllerValidator,
+        '_validate_endpoints',
+        lambda self, controller: False,
+    )
+    with pytest.raises(
+        EndpointMetadataError,
+        match='contains incompatible mixin',
```
sobolevn:
```suggestion
        match='incompatible mixin',
```

tests/test_unit/test_controllers/test_controller_validation.py
```diff
@@ -54,3 +60,77 @@ class _MixedController(Controller[PydanticSerializer]):
...
+            AsyncMetaMixin,
+            Controller[PydanticSerializer],
+        ):
+            def post(self) -> list[str]:
+                raise NotImplementedError
+
+
+def test_async_controller_have_invalid_mixin(monkeypatch) -> None:
+    """Ensure that async controller have an invalid mixin."""
+    monkeypatch.setattr(
```
sobolevn: The same here.

tests/test_unit/test_controllers/test_controller_validation.py
```diff
@@ -54,3 +60,77 @@ class _MixedController(Controller[PydanticSerializer]):
...
+        ControllerValidator,
+        '_validate_endpoints',
+        lambda self, controller: True,
+    )
+    with pytest.raises(
+        EndpointMetadataError,
+        match='contains incompatible mixin',
```
sobolevn:
```suggestion
        match='incompatible mixin',
```

tests/test_unit/test_controllers/test_controller_validation.py
AlexanderZharyuk: Hey, @sobolevn !

Thanks for your comments!
I have a question about this lines of code: I wrote this patch because the `_validate_endpoints(controller)` function already validates the mixin logic indirectly - meaning it’s currently impossible to trigger a validation error for a sync controller with an async mixin (or vice versa), since such cases are caught earlier at the endpoint validation stage. So I decided that it will may be usefull just for test a validate mixin for future.

What do you think? Maybe delete this test so?

sobolevn: Oh, right! In this case we only need to test that you can't provide both mixins to regular and composed controllers. Other logic is covered :)

PR #165, opened 2025-10-25
Issue #153. Add fix to resolve canonical names of types to names that resolve in intersphinx

### LINKED ISSUE #153: Figure out why `sphinx` does not link `HttpResponse` as a return type
See https://github.com/wemake-services/django-modern-rest/blob/7aa23316219d4226d7c49435f9b028130f8c5dbc/docs/conf.py#L98-L99

When `make -C docs clean html` is generated without this line, `sphinx` produces:

```
django-modern-rest/django_modern_rest/errors.py:docstring of django_modern_rest.errors.global_error_handler:1: WARNING: py:class reference target not found: django.http.response.HttpResponse [ref.class]
django-modern-rest/django_modern_rest/endpoint.py:docstring of django_modern_rest.endpoint.Endpoint.__call__:1: WARNING: py:class reference target not found: django.http.response.HttpResponse [ref.class]
django-modern-rest/django_modern_rest/endpoint.py:docstring of django_modern_rest.endpoint.Endpoint.handle_async_error:1: WARNING: py:class reference target not found: django.http.response.HttpResponse [ref.class]
django-modern-rest/django_modern_rest/endpoint.py:docstring of django_modern_rest.endpoint.Endpoint.handle_error:1: WARNING: py:class reference target not found: django.http.response.HttpResponse [ref.class]
```

This does not seem right, because everything is configure: intersphinx, imports, etc.

- Links in the form of inline `:class:` roles are rendered correctly

Image

- The link in the annotations is not generated

Image

asidowner:
## Checklist

## Related issues

Closes #153

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -102,7 +102,7 @@ nitpick_ignore = [
     (PY_CLASS, 'django_modern_rest.endpoint._ModifyAnyCallable'),
     (PY_CLASS, 'django_modern_rest.endpoint._ModifyAsyncCallable'),
     (PY_CLASS, 'django_modern_rest.endpoint._ModifySyncCallable'),
-    (PY_CLASS, 'django_modern_rest.endpoint._ParamT'),
+    (PY_CLASS, '_ParamT'),
     (PY_CLASS, 'django_modern_rest.internal.middleware_wrapper._TypeT'),
 ]
 


docs/conf.py
```diff
@@ -157,3 +166,43 @@ def _get_project_meta() -> dict[str, str]:
...
+
+        * https://github.com/sphinx-doc/sphinx/issues/4826 - solves this
+            with the "canonical" directive
+        * https://github.com/pyca/cryptography/pull/7938 - where this
+            was fixed for cryptography
+        * https://www.sphinx-doc.org/en/master/extdev/appapi.html#events
+        * https://stackoverflow.com/a/62301461 - source of this hack
+
+    """
+    pending_xrefs: Generator[pending_xref] = doctree.findall(
```
sobolevn:
```suggestion
    pending_xrefs: Iterable[pending_xref] = doctree.findall(
```

docs/conf.py
```diff
@@ -12,9 +12,14 @@
 
 import sys
 import tomllib
+from collections.abc import Generator
```
sobolevn:
```suggestion
from collections.abc import Iterator
```

PR #166, opened 2025-10-25
Extend `EndpointMetadata`

milssky:
# I have made things!
I extend the EndpointMetadata class for uses in OpenAPI generators.

## Checklist

### DIFF
--- a/django_modern_rest/endpoint.py
+++ b/django_modern_rest/endpoint.py
@@ -521,6 +521,13 @@ def modify(
     validate_responses: bool | None = None,
     extra_responses: list[ResponseDescription] | Empty = EmptyObj,
     allow_custom_http_methods: bool = False,
+    summary: str | None = None,
+    description: str | None = None,
+    tags: list[str] | None = None,
+    operation_id: str | None = None,
+    deprecated: bool = False,
+    security: list[SecurityRequirement] | None = None,
+    external_docs: ExternalDocumentation | None = None,
 ) -> _ModifyAsyncCallable: ...
 
 
@@ -533,6 +540,13 @@ def modify(
     validate_responses: bool | None = None,
     extra_responses: list[ResponseDescription] | Empty = EmptyObj,
     allow_custom_http_methods: bool = False,
+    summary: str | None = None,
+    description: str | None = None,
+    tags: list[str] | None = None,
+    operation_id: str | None = None,
+    deprecated: bool = False,
+    security: list[SecurityRequirement] | None = None,
+    external_docs: ExternalDocumentation | None = None,
 ) -> _ModifySyncCallable: ...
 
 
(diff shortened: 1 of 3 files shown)


django_modern_rest/validation.py
```diff
@@ -297,6 +310,15 @@ class ModifyEndpointPayload:
     error_handler: SyncErrorHandlerT | AsyncErrorHandlerT | Empty
     allow_custom_http_methods: bool
 
+    # OpenAPI documentation fields:
+    summary: str | None = None
```
sobolevn: Please, use a common base class for common fields. We don't want to duplicate this.

django_modern_rest/endpoint.py
```diff
@@ -333,13 +337,20 @@ def validate(
 ]: ...
 
 
-def validate(  # pyright: ignore[reportInconsistentOverload]
+def validate(  # noqa: WPS211  # pyright: ignore[reportInconsistentOverload]
     response: ResponseDescription,
     /,
     *responses: ResponseDescription,
     validate_responses: bool | Empty = EmptyObj,
     error_handler: SyncErrorHandlerT | AsyncErrorHandlerT | Empty = EmptyObj,
```
sobolevn: Please, fix `error_handler` and `validate_responses` as well, they should be `None` by default, not `Empty`.

docs/conf.py
```diff
@@ -104,6 +104,9 @@ def _get_project_meta() -> dict[str, str]:
     (PY_CLASS, 'django_modern_rest.endpoint._ModifySyncCallable'),
     (PY_CLASS, '_ParamT'),
     (PY_CLASS, 'django_modern_rest.internal.middleware_wrapper._TypeT'),
+    # OpenAPI types used in TYPE_CHECKING blocks:
+    (PY_CLASS, 'SecurityRequirement'),
```
sobolevn, line 112: They should be documented, please add the `autoclass` docs in `openapi.rst`

django_modern_rest/validation.py
kondratevdev: @sobolevn Should such attributes be `Empty` by default, not `None`?

sobolevn: I only need `Empty` in `return_type`, actually. Because there `None` is a valid value. In other places, I just copied that design without thinking twice. Now I see the problem :)

django_modern_rest/endpoint.py
```diff
@@ -525,10 +552,17 @@ def modify(  # noqa: WPS211
...
-    error_handler: SyncErrorHandlerT | AsyncErrorHandlerT | Empty = EmptyObj,
+    error_handler: SyncErrorHandlerT | AsyncErrorHandlerT | None = None,
     allow_custom_http_methods: bool = False,
+    summary: str | None = None,
+    description: str | None = None,
+    tags: list[str] | None = None,
+    operation_id: str | None = None,
+    deprecated: bool = False,
+    security: list[SecurityRequirement] | None = None,
+    external_docs: ExternalDocumentation | None = None,
```
sobolevn, line 586: You need to update all `modify` overloads with this, these fields are always allowed. To catch this, please provide one optionl field like `description` in `typesafety/`

django_modern_rest/validation.py
```diff
@@ -277,24 +281,35 @@ def _validate_endpoints(
 
 
 @dataclasses.dataclass(slots=True, frozen=True, kw_only=True)
```
```suggestion
```

It is not a dataclass by itself, it cannot be created, it is only a base class. Just add `__slots__`

sobolevn (approved):

PR #179, opened 2025-10-26
Fix `_BaseResponseHeader`

milssky:
# I have made things!

I changed `_BaseResponseHeader` to have consistent inheritance, like `_OpenAPIPayload'.

## Checklist


sobolevn (changes requested):
Please, add a test case :)

PR #181, opened 2025-10-26
Extend EnpointMetadata attributes

kondratevdev:
# I have made things!

## Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -111,6 +111,11 @@ nitpick_ignore = [
     # OpenAPI types used in TYPE_CHECKING blocks:
     (PY_CLASS, 'SecurityRequirement'),
     (PY_CLASS, 'ExternalDocumentation'),
+    (PY_CLASS, 'Callback'),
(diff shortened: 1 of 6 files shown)


docs/conf.py
```diff
@@ -111,6 +111,11 @@ def _get_project_meta() -> dict[str, str]:
     # OpenAPI types used in TYPE_CHECKING blocks:
     (PY_CLASS, 'SecurityRequirement'),
     (PY_CLASS, 'ExternalDocumentation'),
+    (PY_CLASS, 'Callback'),
```
sobolevn, line 114: Not sure why this is required?
It should be clear from the annotation, these objects do exist in the docs.

kondratevdev: `Callback` is a `TypeAlias` (a dict alias), not a class. Sphinx’s nitpick mode tries to resolve it but can’t create a link because there’s no class/object, only a dict equivalent. Adding it to `nitpick_ignore` avoids the warning and correctly skips it.

PR #183, opened 2025-10-26
feat: add prefix-optimized routes

deliro:
Added prefix-based route matching optimization to improve routing performance. The new `path()` function from `django_modern_rest` uses `_PrefixRoutePattern` for faster URL resolution by checking route prefixes before full regex pattern matching.

## Changes:
- Added `path()` function from `django_modern_rest` to use the new prefix-optimized pattern

## Checklist

### DIFF
--- a/django_modern_rest/routing.py
+++ b/django_modern_rest/routing.py
@@ -181,7 +181,7 @@ class _PrefixRoutePattern(RoutePattern):
             if not self._is_endpoint and path.startswith(self._prefix):
                 return path[len(self._prefix) :], (), {}
         elif path.startswith(self._prefix):
-            return cast(_RouteMatch | None, super().match(path))  # type: ignore[redundant-cast]
+            return super().match(path)  # pyright: ignore
         return None
 
 


README.md
```diff
@@ -92,8 +92,8 @@ The shortest example:
 And then route this controller in your `urls.py`:
 
 ```python
->>> from django.urls import include, path
->>> from django_modern_rest import Router
+>>> from django.urls import include
+>>> from django_modern_rest import Router, path
```
sobolevn: I would prefer to keep a readme example as simple and as familiar as possible.

django_modern_rest/components.py
```diff
@@ -218,9 +218,9 @@ class Path(ComponentParser, Generic[_PathT]):
     .. code:: python
 
         >>> import pydantic
-        >>> from django_modern_rest import Body, Path, Controller, Router
+        >>> from django_modern_rest import Body, Path, Controller, Router, path
```
sobolevn: Again, let's test everything with the default. Keeping our stuff as a separate drop-in feature, not a requirement.

docs/pages/routing.rst
```diff
@@ -16,8 +16,8 @@ behaviours into a single URL? For this we use
 
 .. code:: python
```
sobolevn, line 17: Please, document this feature here :)

django_modern_rest/components.py
deliro: If prefix routes were an opt-in feature, what should we call it? `path` seems confusing. `fast_path` looks silly, and `prefix_path` feels a bit like exposing internals. Any ideas?

deliro:
What should I use for benchmarking?

django_modern_rest/components.py
sobolevn: it should be just `path`. here's the main idea of the feature:

```diff
--- from django.urls import path
+++ from django_modern_rest import path
```

tests/test_unit/test_routing/test_prefix_route_pattern.py
```diff
@@ -0,0 +1,79 @@
...
+            False,
+        ),
+    ],
+)
+def test_prefix_route_pattern_matching(
+    *,
+    route: str,
+    path_to_match: str,
+    is_endpoint: bool,
+    should_match: bool,
```
sobolevn: One more thing: can we please assert the `matched` dict instead of just `True`?

We need to be sure that `api/v1/users/<int:user_id>/posts/<int:post_id>'` matches `'api/v1/users/2/posts/123'` as `{'user_id': 2, 'post_id': 123}`

deliro: Yeah, makes sense, thank you!

django_modern_rest/routing.py
```diff
@@ -140,3 +151,89 @@ def _build_method_mapping(
...
+        self,
+        path: str,
+    ) -> _RouteMatch | None:
+        if self._is_static:
+            if self._is_endpoint and path == self._prefix:
+                return '', (), {}
+            if not self._is_endpoint and path.startswith(self._prefix):
+                return path[len(self._prefix) :], (), {}
+        elif path.startswith(self._prefix):
+            return cast(_RouteMatch | None, super().match(path))  # type: ignore[redundant-cast]
```
sobolevn:
```suggestion
            return super().match(path)  
```

deliro: This was done on purpose. While mypy accepts just `super().match(path)`, pyright doesn't.

sobolevn: Then we should use `# pyright: ignore`, because `cast` and `Union` contruction take time, we don't want to do that in the hot path

README.md
deliro: done

django_modern_rest/components.py
deliro: done

docs/pages/routing.rst
deliro: documented

tests/test_unit/test_routing/test_prefix_route_pattern.py
deliro: tests are rewritten in one "more real world" test

django_modern_rest/routing.py
deliro: done

sobolevn (approved):
Great stuff, thanks a lot!

PR #184, opened 2025-10-26
Run code examples

F0RRZZ:
I've implemented a plugin for automatic code execution using `# run` comments.
You can see how to use it in `docs/examples/getting_started/single_file_asgi.py`

To run some code, just add this:
```
# run: {"controller": "ControllerName", "method": "http_method", "body": {}, "headers": {}, "url": "/api/user/"}
```
The `url` is optional - only use it if you need to hit a specific endpoint (see the `single_file_asgi.py` example).
`body` and `headers` are also optional.

The Django app gets auto-generated, so in your examples you only need to write your controller.

The output will look like this:

Screenshot 2025-10-27 at 00 57 06

### DIFF
--- a/tools/sphinx_ext/run_examples.py
+++ b/tools/sphinx_ext/run_examples.py
@@ -389,10 +389,7 @@ class LiteralInclude(LiteralIncludeOverride):  # type: ignore[misc]
         clean_content: str,
     ) -> None:
         cwd = Path.cwd()
-        if cwd.name == 'docs':
-            docs_dir = cwd
-        else:
-            docs_dir = cwd / 'docs'
+        docs_dir = cwd if cwd.name == 'docs' else cwd / 'docs'
         tmp_file = (
             cwd
             / PATH_TO_TMP_EXAMPLES


docs/pages/micro-framework.rst
```diff
@@ -8,4 +8,5 @@ You don't need microframeworks to build small APIs, because
 Django is a microframework itself. With the only difference: it scales!
 
 .. literalinclude:: /examples/getting_started/single_file_asgi.py
+   :language: python
```
sobolevn: Isn't it autodetected from `.py` extension?

tools/sphinx_ext/missing_references.py
```diff
@@ -0,0 +1,225 @@
...
+    from collections.abc import Generator
+    from typing import Any, TypeAlias
+
+    from docutils.nodes import Element, Node
+    from sphinx.addnodes import pending_xref
+    from sphinx.application import Sphinx
+    from sphinx.environment import BuildEnvironment
+
+
+IgnoreRefs: TypeAlias = dict[
```
sobolevn: Sorry, not clear what this file does and why is it needed?

tools/sphinx_ext/run_examples.py
```diff
@@ -0,0 +1,401 @@
+from __future__ import annotations
```
sobolevn, line 7: Please, add a note that we took some parts of it from `Litestar`. Let's credit where credit is due.

sobolevn (commented):
Please, fix the CI problem :)
The idea looks great!

sobolevn (approved):
AM-AZ-ING! Thanks a lot, we can continue from here in the next PRs 👏

PR #211, opened 2025-10-29
Fix make comment indent

Tapeline:
# I have made things!

Existing Makefile breaks on Windows (maybe on other platforms too).

As per [Makefile documentation](https://www.gnu.org/software/make/manual/html_node/Makefile-Contents.html):

> Comments within a recipe are passed to the shell, just as with any other recipe text. The shell decides how to interpret it: whether or not this is a comment is up to the shell.

On Windows 10, GNU Make 4.4.1 for Windows32 `smoke` recipe fails with:

```
> make smoke
# Checks that it is possible to import the base package without django.setup
process_begin: CreateProcess(NULL, # Checks that it is possible to import the base package without django.setup, ...) failed.
make (e=2): �� ������� ����� ��������� ����.
make: *** [Makefile:30: smoke] Error 2
```

Comment has been unindented to prevent that.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ unit:
 
 .PHONY: smoke
 smoke:
-	# Checks that it is possible to import the base package without django.setup
+# Checks that it is possible to import the base package without django.setup
 	poetry run python -c 'from django_modern_rest import Controller'
 
 .PHONY: example


Makefile
```diff
@@ -27,7 +27,7 @@ unit:
 
 .PHONY: smoke
 smoke:
-	# Checks that it is possible to import the base package without django.setup
+# Checks that it is possible to import the base package without django.setup
```
kondratevdev, line 30: Maybe is it better or will it not work?:

```suggestion
smoke:  # Checks that it is possible to import the base package without django.setup
    poetry run python -c 'from django_modern_rest import Controller'
```

kondratevdev: Oh, sorry. I read the documentation, my mistake :)

kondratevdev:
LGTM, thank you!

PR #224, opened 2025-10-30
Concretise error_serialize docstring

Tapeline:
# I have made things!

## Checklist

### DIFF
--- a/django_modern_rest/serialization.py
+++ b/django_modern_rest/serialization.py
@@ -115,7 +115,7 @@ class BaseSerializer:
     @abc.abstractmethod
     def error_serialize(cls, error: Exception | str) -> Any:
         """
-        Serialize a serializer exception to json the best way possible.
+        Convert serialization or deserialization error to json format.
 
         Args:
             error: A serialization exception like a validation error or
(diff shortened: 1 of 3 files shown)


django_modern_rest/serialization.py
```diff
@@ -114,7 +114,16 @@ def from_python(
     @classmethod
     @abc.abstractmethod
     def error_serialize(cls, error: Exception | str) -> Any:
-        """Serialize an exception to json the best way possible."""
+        """
+        Serialize a serializer exception to json the best way possible.
```
sobolevn:
```suggestion
       Convert related error to the json format.
       
       What errors are "related"? Ones that we can
       expect when serializing or deserializing data.
       
       ...
```

Tapeline: > What errors are "related"? Ones that we can expect when serializing or deserializing data.

I think documentation should have more strict wording.

How about this?
```
Convert serialization or deserialization error to json format.
Args:
    error: A serialization exception like a validation error or
        a ``django_modern_rest.exceptions.DataParsingError``.
 ```

sobolevn: Nice!

PR #228, opened 2025-10-31
feat: validate error handlers

### LINKED ISSUE #214: Validate `async_error_handler` with sync endpoint and vice-a-versa
Right now this code does not raise a validation error:

```python
class _AsyncErrorHandlingController(Controller[PydanticSerializer]):
    blueprints: ClassVar[Sequence[_BlueprintT]] = [_AsyncErrorHandlingBlueprint]

    @override
    async def handle_async_error(
        self,
        endpoint: Endpoint,
        exc: Exception,
    ) -> HttpResponse:
        if isinstance(exc, KeyError):
            return self.to_response('controller', status_code=HTTPStatus.OK)
        raise exc

     def post(self) -> str:
         return 'a'
```

But, the thing is: it is probably an error. You should use sync error handler with sync routes.

@milssky will you please add this validation rule?

milssky:
# I have made things!

## Checklist

## Related issues

Closes #214

### DIFF
--- a/tests/test_unit/test_controllers/test_controller_validation.py
+++ b/tests/test_unit/test_controllers/test_controller_validation.py
@@ -6,13 +6,11 @@ import pytest
 from typing_extensions import override
 
 from django_modern_rest import (
-    AsyncMetaMixin,
     Blueprint,
     Controller,
-    Endpoint,
-    MetaMixin,
     ResponseDescription,
 )
+from django_modern_rest.endpoint import Endpoint
 from django_modern_rest.exceptions import EndpointMetadataError
 from django_modern_rest.options_mixins import AsyncMetaMixin, MetaMixin
 from django_modern_rest.plugins.pydantic import PydanticSerializer


django_modern_rest/validation.py
```diff
@@ -306,6 +307,31 @@ def _validate_endpoints(
             )
         return is_async
 
+    def _validate_error_handlers(
+        self,
+        blueprint: 'type[Blueprint[BaseSerializer]]',
+        *,
+        is_async: bool,
+    ) -> None:
+        handle_error_overridden = 'handle_error' in blueprint.__dict__
```
sobolevn, line 319:
```suggestion
        if not blueprint.api_endpoints:
            return
            
        handle_error_overridden = 'handle_error' in blueprint.__dict__
```

sobolevn (commented):
We should allow empty blueprints to have any error handlings as they want.

tests/test_unit/test_controllers/test_controller_validation.py
```diff
@@ -88,3 +95,239 @@ class _MixedController2(  # type: ignore[misc]
...
+                self,
+                endpoint: Endpoint,
+                exc: Exception,
+            ) -> Any:
+                raise NotImplementedError
+
+
+def test_sync_bp_with_async_handler_fails() -> None:
+    """Ensure controllers with blueprints cannot use async error handler."""
+    with pytest.raises(  # noqa: PT012
```
sobolevn:
```suggestion
    with pytest.raises(
```

PR #252, opened 2025-11-02
Sphinx chartjs

kondratevdev:
# I have made things!

## Checklist

### DIFF
--- a/docs/tools/sphinx_ext/__init__.py
+++ b/docs/tools/sphinx_ext/__init__.py
@@ -14,10 +14,12 @@ def _register_directives(app: Sphinx) -> None:
         override=True,
     )
     app.add_directive('code-block', CodeBlockOverride, override=True)
+    app.add_directive('chartjs', chartjs.ChartJSDirective)
 
 
 def setup(app: Sphinx) -> dict[str, bool]:
     """Initialize Sphinx extensions and return configuration."""
     app.connect('builder-inited', _register_directives)
     chartjs.setup(app)
-    return run_examples.setup(app)
+    run_examples.setup(app)
+    return {'parallel_read_safe': True, 'parallel_write_safe': True}
--- a/docs/tools/sphinx_ext/chartjs.py
+++ b/docs/tools/sphinx_ext/chartjs.py
@@ -1,10 +1,10 @@
 import json
-from typing import TYPE_CHECKING, Any, Final, final
-from typing_extensions import override
+from typing import TYPE_CHECKING, Final, final
 
 from docutils import nodes
 from sphinx.application import Sphinx
 from sphinx.util.docutils import SphinxDirective
+from typing_extensions import override
 
 if TYPE_CHECKING:
     from sphinx.writers.html5 import HTML5Translator
@@ -32,6 +32,7 @@ _SCRIPT: Final = """
 }})();
 </script>
 """
+
 
 @final
 class ChartJSNode(nodes.General, nodes.Element):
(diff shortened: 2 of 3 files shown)


tests/test_unit/test_plugins/test_msgspec/test_serializers.py
kondratevdev, line 1: emm, after merge from master i have this diff, trying to fix that

.github/workflows/test.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       run: |
         poetry config virtualenvs.in-project true
         poetry run pip install -U pip
-        poetry install --all-extras
+        poetry install --all-extras --all-groups
```
kondratevdev, line 46: @sobolevn FYI, required for correct testing

docs/docs/_build/_tmp_example/examples_getting_started_single_file_asgi.py
sobolevn, line 1: Why do we need this file?

docs/tools/sphinx_ext/chartjs.py
```diff
@@ -0,0 +1,116 @@
...
+                new Chart(ctx, config);
+            }}
+        }});
+    }}
+}})();
+</script>
+"""
+
+
+class ChartJSNode(nodes.General, nodes.Element):
```
sobolevn, line 38:
```suggestion
@final
class ChartJSNode(nodes.General, nodes.Element):
```

and in other places as well :)

docs/tools/sphinx_ext/chartjs.py
```diff
@@ -0,0 +1,116 @@
...
+    self.body.append(canvas)
+    self.body.append(
+        _SCRIPT.format(
+            chart_id=node['chart_id'],
+            chart_config=node['chart_config'],
+        ),
+    )
+
+
+def depart_chartjs_html(self: 'HTML5Translator', node: ChartJSNode) -> None:
```
sobolevn: Please, make sure that you only have public names for just public apis. Other names should be protected.

docs/tools/sphinx_ext/__init__.py
```diff
@@ -19,4 +19,5 @@ def _register_directives(app: Sphinx) -> None:
 def setup(app: Sphinx) -> dict[str, bool]:
     """Initialize Sphinx extensions and return configuration."""
     app.connect('builder-inited', _register_directives)
+    chartjs.setup(app)
     return run_examples.setup(app)
```
sobolevn: Why do we return just the second dict? What about `chartjs.setup(app)` dict?

.github/workflows/test.yml
sobolevn: What happens without this line? Because the only group we don't install is `docs`. Why do we need `docs` for testing?

kondratevdev: mypy cannot correct find sphinx types, because we do not install this depend

sobolevn: why does it work right now for https://github.com/wemake-services/django-modern-rest/tree/master/docs/tools ?

kondratevdev: I think because we has `shpinx` types only at https://github.com/wemake-services/django-modern-rest/blob/4f57a7a50cd4e61174231b4e0caceb3dcc64694d/docs/tools/sphinx_ext/run_examples.py#L402C11-L402C22 line, this is an unused variable

In my code, I directly inherit from sphinx classes and use their methods, so `mypy` throws an error when it cannot check the interface due to the dependency being unavailable and it defines it as `Any`

docs/docs/_build/_tmp_example/examples_getting_started_single_file_asgi.py
kondratevdev: My bad :(

docs/tools/sphinx_ext/__init__.py
```diff
@@ -14,9 +14,12 @@ def _register_directives(app: Sphinx) -> None:
...
+    app.add_directive('chartjs', chartjs.ChartJSDirective)
 
 
 def setup(app: Sphinx) -> dict[str, bool]:
     """Initialize Sphinx extensions and return configuration."""
     app.connect('builder-inited', _register_directives)
-    return run_examples.setup(app)
+    chartjs.setup(app)
+    run_examples.setup(app)
+    return {'parallel_read_safe': True, 'parallel_write_safe': True}
```
kondratevdev, line 25: I've moved the configuration to this location. We shouldn't have multiple sources of truth.

.github/workflows/test.yml
sobolevn: Ok, got it!

sobolevn (approved):

PR #257, opened 2025-11-02
feat: use <lru_cache> for <_is_validation_enabled>

### LINKED ISSUE #230: Optimize `_is_validation_enabled` by using cache
This method is in hot-path, it is called for every response: https://github.com/wemake-services/django-modern-rest/blob/c6071d9907c72a54258bc7f644822e91640559d4/django_modern_rest/validation.py#L137-L162

So, we need to optimize it to the max. I propose to:
1. Make a function out of this method
2. Use `@lru_cache(maxsize=MAX_CACHE_SIZE)` as we do in other places to cache the function result (since the result can't change for a given endpoint / controller)
3. Profit!

ApostolFet:
# I have made things!

## Checklist

## Related issues
- Closes #230


django_modern_rest/validation.py
```diff
@@ -266,6 +247,36 @@ def _validate_response_cookies(  # noqa: WPS210
...
+    Should we run response validation?
+
+    Priority:
+    - We first return any directly specified *validate_responses*
+        argument to endpoint itself
+    - Second is *validate_responses* on the blueprint, if it exists
+    - Then we return *validate_responses* from controller if specified
+    - Lastly we return the default value from settings
+    """
+    if isinstance(metadata_validate_responses, bool):
```
ApostolFet: Maybe use none check instead of `isinstance`?
```suggestion
    if metadata_validate_responses is not None:
```

sobolevn: yes, should be faster!

django_modern_rest/validation.py
```diff
@@ -266,6 +247,36 @@ def _validate_response_cookies(  # noqa: WPS210
...
+        return metadata_validate_responses
+    if controller.blueprint and isinstance(
+        controller.blueprint.validate_responses,
+        bool,
+    ):
+        return controller.blueprint.validate_responses
+    if isinstance(controller.validate_responses, bool):
```
sobolevn: here

django_modern_rest/validation.py
```diff
@@ -266,6 +247,36 @@ def _validate_response_cookies(  # noqa: WPS210
...
+    Priority:
+    - We first return any directly specified *validate_responses*
+        argument to endpoint itself
+    - Second is *validate_responses* on the blueprint, if it exists
+    - Then we return *validate_responses* from controller if specified
+    - Lastly we return the default value from settings
+    """
+    if isinstance(metadata_validate_responses, bool):
+        return metadata_validate_responses
+    if controller.blueprint and isinstance(
```
sobolevn: and here

Issue #259, opened 2025-11-02
Validate that we can't set `Set-Cookie` header and should use `cookies=` instead


sobolevn:
When using `@validate` and `@modify` we must check that no `ResponseSpec.headers` contain `Set-Cookie` header description / modification.

If so, we need to raise an error that `cookies=` parameter must be used instead.

PR #261, opened 2025-11-02
validate Set-Cookie header usage in `@modify` and `@validate` decorat…

MukievMukhammad:
#259

### DIFF
--- a/django_modern_rest/validation.py
+++ b/django_modern_rest/validation.py
@@ -562,7 +562,9 @@ class _ResponseListValidator:
         endpoint: str,
     ) -> None:
         for response in responses:
-            if response.headers is not None and any(
+            if response.headers is None:
+                continue
+            if any(
                 header_name.lower() == 'set-cookie'
                 for header_name in response.headers
             ):


django_modern_rest/validation.py
```diff
@@ -827,7 +835,17 @@ def _validate_new_headers(
         *,
         endpoint: str,
     ) -> None:
-        if payload.headers is not None and any(
+        if payload.headers is None:
```
sobolevn: Let's have this check in only one place. Let's move it near `_validate_http_spec`, there's a todo called `# TODO: validate cookie descriptions`. So, this logic will be in a single place.

tests/test_unit/test_endpoint/test_modify_decorator.py
```diff
@@ -232,3 +232,77 @@ async def async_endpoint_error(
...
+    """@modify with Set-Cookie in headers= raise EndpointMetadataError."""
+    with pytest.raises(EndpointMetadataError, match='Set-Cookie'):
+
+        class _SetCookieHeaderController(Controller[PydanticSerializer]):
+            @modify(
+                status_code=HTTPStatus.OK,
+                headers={'Set-Cookie': NewHeader(value='session=abc123')},
+            )
+            def post(self) -> dict[str, str]:
+                return {'result': 'done'}  # pragma: no cover
```
sobolevn:
```suggestion
                raise NotImplementedError
```

tests/test_unit/test_endpoint/test_modify_decorator.py
```diff
@@ -232,3 +232,77 @@ async def async_endpoint_error(
             )
             def get(self) -> int:
                 raise NotImplementedError
+
+
+def test_modify_with_set_cookie_in_headers_raises_error() -> None:
```
sobolevn:
```suggestion
def test_modify_with_set_cookie() -> None:
```

tests/test_unit/test_endpoint/test_modify_decorator.py
```diff
@@ -232,3 +232,77 @@ async def async_endpoint_error(
...
+
+
+def test_modify_with_set_cookie_case_insensitive() -> None:
+    """@modify with Set-Cookie in headers= raise EndpointMetadataError."""
+    with pytest.raises(EndpointMetadataError, match='Set-Cookie'):
+
+        class _SetCookieHeaderController(Controller[PydanticSerializer]):
+            @modify(
+                status_code=HTTPStatus.OK,
+                headers={'set-cookie': NewHeader(value='session=abc123')},
```
sobolevn: Just make the previous test parameterized on the header name. And do not duplicate the test.

tests/test_unit/test_endpoint/test_modify_decorator.py
```diff
@@ -232,3 +232,77 @@ async def async_endpoint_error(
...
+    """Testing @modify without headers."""
+
+    @modify(status_code=HTTPStatus.CREATED)
+    def post(self) -> dict[str, str]:
+        return {'result': 'done'}  # pragma: no cover
+
+
+def test_modify_without_headers() -> None:
+    """@modify without headers should work."""
+    assert 'POST' in _ModifyWithoutHeadersController.api_endpoints
```
sobolevn: We already have tests for other headers, we don't need them.

```suggestion
```

sobolevn: Please, follow the same logic in `@validate` tests :)

django_modern_rest/validation.py
MukievMukhammad: i am confusing, cuz we have separate validation classes `EndpointMetadataValidator` and `_ResponseListValidator` which is the member of first calling from `_from_modify` and `_from_validate`. How can i move this check in a single place? also i have thought maybe it is good to move header validation into separate class `SomeHeaderValidationMixin` so we can call it whether in  `EndpointMetadataValidator` or `_ResponseListValidator`, following DRY principle or it is just overengineering?

MukievMukhammad: we have 2 options to set headers:
1. using `response` prop
- `ValidateEndpointPayload.responses.header`
- `EndpointMetadataValidator.responses.header`
2. Directly in `ModifyEndpointPayload.headers`

So we validate both cases:
1. first case [here](https://github.com/MukievMukhammad/django-modern-rest/blob/a01036acb5f21f4ee6293d31794291115b86a962/django_modern_rest/validation.py#L540)
2. second [here](https://github.com/MukievMukhammad/django-modern-rest/blob/a01036acb5f21f4ee6293d31794291115b86a962/django_modern_rest/validation.py#L838)

am i understand correctly that we need move checks `headers is not None` from both cases to single place? but how?)

sobolevn: We combine them into `_AllResponses`, that's the proper place

PR #267, opened 2025-11-03
Fix routing examples

gimntut:
In #231 global API has changed, but some of the documentation remains the same.
1. Fixed it
2. Added a test of the README.md file
3. Extracted example code to python file

### DIFF
--- a/docs/pages/routing.rst
+++ b/docs/pages/routing.rst
@@ -103,15 +103,9 @@ Example Workflow
 
 Let's say you have this URL configuration:
 
-.. code:: python
-
-    from django_modern_rest.routing import Router, path
-
-    router = Router([
-        path('api/v1/users/', views.UserList.as_view()),
-        path('api/v1/posts/', views.PostList.as_view()),
-        path('api/v1/users/<int:id>/', views.UserDetail.as_view()),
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -108,7 +108,7 @@ And then route this controller in your `urls.py`:
 
 ```python
 >>> from django.urls import include, path
->>> from django_modern_rest import Router
+>>> from django_modern_rest.routing import Router
```
sobolevn, line 112: Can you please explore, why this example is not runned in CI? We must check the readme with `doctest`

docs/pages/routing.rst
```diff
@@ -105,7 +105,7 @@ Let's say you have this URL configuration:
 
 .. code:: python
 
-    from django_modern_rest import Router, path
+    from django_modern_rest.routing import Router, path
```
sobolevn: can you please move this example to its own file and make sure that it is checked with mypy?

you would need to import some views from any other module.

sobolevn (commented):
Please, convert these examples to a runnable ones, so we can find this in CI :)

sobolevn (approved):
Thanks!

PR #271, opened 2025-11-04
Introduce HeaderDict class for handling http headers

### LINKED ISSUE #256: Consider using better suited data structure for HTTP headers
Currently, headers in the response appear to be stored and handled using a plain `dict`.

This can lead to ambiguity when multiple headers with the same key are added using different casing, or when a header is meant to have multiple values. For example:

```python
headers["content-length"] = 1
headers["Content-Length"] = 2
# which one will be used in the final response?
```

or:

```python
headers["x-my-list-like-header"] = 1
headers["x-my-list-like-header"] = 2
# both values should ideally be included in the response
```

Some examples where this pattern appears:

* https://github.com/wemake-services/django-modern-rest/blob/e67856d0bea3f6801935b844ecc36609ef50e277/django_modern_rest/response.py#L188
* https://github.com/wemake-services/django-modern-rest/blob/e67856d0bea3f6801935b844ecc36609ef50e277/django_modern_rest/response.py#L132
* https://github.com/wemake-services/django-modern-rest/blob/e67856d0bea3f6801935b844ecc36609ef50e277/django_modern_rest/controller.py#L181

(And possibly other places as well.)

Since HTTP headers are **case-insensitive** and may contain **multiple values per key** (e.g., `Set-Cookie`), using a plain `dict` can cause subtle bugs or data loss.

To make header handling more robust, it would be better to use a data structure designed for this purpose, such as:

* [CIMultiDict](https://github.com/aio-libs/multidict) from `aio-libs`, or
* a list of `(key, value)` tuples.
* a dict where each value is always a list of header values.

`multidict` seems like the most natural fit here, but it’s a binary dependency and might not have wheels available for all platforms.
If that’s a concern, a simpler alternative could be to implement a small custom `HeaderDict` wrapper class around a list of tuples  (or dict of lists) that implements the `Mapping` protocol.

Also I want to mention pure python implementation of `Header` d.s. that I personally worked with: 
https://github.com/scrapy/scrapy/blob/master/scrapy/http/headers.py and it's issues: https://github.com/scrapy/scrapy/pull/5146

albertedwardson:
# I have made things!

In this PR I introduce a custom `HeaderDict` class tailored to HTTP header semantics, rather than relying on Django’s `HttpHeaders` or `MultiValueDict` classes.

The Django `HttpHeaders` class is immutable and expects `HTTP_`-prefixed keys, making it unsuitable for a general header-map container.

The `MultiValueDict` is built for form/query parameter use-cases, returns only the last value in lists via `__getitem__`, and supports list-oriented manipulation rather than comma-joined header strings, its semantics don’t align with standard HTTP header usage.

By implementing HeaderDict as a UserDict subclass, I tried to ensure:
- consistent key normalization (Title-Case),
- multiple values stored as comma-joined strings (not lists),
- merge semantics (duplicate keys combine into single string),
- simpler, more friendly semantics for header handling in our context.
- `Set-Cookie` headers are **explicitly forbidden** in this implementation, since `Set-Cookie` semantics differ from other headers, where it is not combinable, not comma-joinable, and may appear multiple times independently in a response.

## Checklist

## Related issues

- Closes #256

### DIFF
--- a/django_modern_rest/headers.py
+++ b/django_modern_rest/headers.py
@@ -1,5 +1,4 @@
 import dataclasses
-import string
 import typing
 from collections import UserDict
 from collections.abc import Iterable, Mapping
@@ -13,7 +12,8 @@ if typing.TYPE_CHECKING:  # pragma: no cover
 HeaderLike: typing.TypeAlias = typing.Union['HeaderDict', dict[str, str]]
 StrOrIterstr: typing.TypeAlias = str | Iterable[str]
 
-_header_strip_chars: str = string.whitespace + ','  # noqa: WPS336
+# ref: https://infra.spec.whatwg.org/#ascii-whitespace
+_ASCII_WHITESPACE = '\t\n\x0c\r '  # noqa: WPS336
 
 
 @dataclasses.dataclass(frozen=True, slots=True, kw_only=True, init=False)
@@ -100,7 +100,7 @@ class HeaderDict(UserDict[str, str]):  # noqa: WPS214
     ``"content-type"`` -> ``"Content-Type"``). Values are stored as strings
     separated by a comma to allow multi-valued headers.
 
-    Use `add(key, val)` or `update(key, vals)`
+    Use `add(key, value)` or `update(values)`
     to append values without replacing.
 
     Examples:
@@ -165,7 +165,7 @@ class HeaderDict(UserDict[str, str]):  # noqa: WPS214
     @typing.override
     def __delitem__(self, key: str) -> None:  # noqa: WPS603
         """Delete key from headers case insensetively."""
-        self.data.pop(self._make_key(key), None)
+        self.data.pop(self._make_key(key))
 
     @typing.override
     def __or__(self, other: typing.Any) -> typing.NoReturn:
@@ -245,9 +245,9 @@ class HeaderDict(UserDict[str, str]):  # noqa: WPS214
                 return [
                     striped
                     for subvalue in input_value  # pyright: ignore[reportUnknownVariableType]
-                    if (striped := subvalue.strip(_header_strip_chars))  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+                    if (striped := subvalue.strip(_ASCII_WHITESPACE))  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
                 ]
-            except Exception as exc:
+            except (TypeError, AttributeError) as exc:
                 raise TypeError(
                     'All elements of header iterable must be `str`',
                 ) from exc


django_modern_rest/response.py
```diff
@@ -209,14 +210,13 @@ def build_response(  # noqa: WPS211
             'Cannot pass both `method=None` and `status_code=Empty`',
         )
 
-    response_headers = {} if headers is None else headers
-    if 'Content-Type' not in response_headers:
-        response_headers['Content-Type'] = serializer.content_type
+    headers = HeaderDict() if headers is None else headers
```
albertedwardson: revert

tests/test_unit/test_headers/test_headers_consistency.py
```diff
@@ -0,0 +1,182 @@
+import pytest
```
albertedwardson, line 3: dedublicate what ai generated

django_modern_rest/headers.py
```diff
@@ -68,13 +70,94 @@ class HeaderSpec(_BaseResponseHeader):
...
+    if modification.headers:
+        result_headers.update({
+            header_name: response_header.value
+            for header_name, response_header in modification.headers.items()
+        })
     return result_headers
+
+
+@final
+class HeaderDict(UserDict[str, Any]):
```
albertedwardson: implement building from sequence of (key, value) pairs with repeated keys

django_modern_rest/headers.py
```diff
@@ -68,13 +70,94 @@ class HeaderSpec(_BaseResponseHeader):
...
+    def __delitem__(self, key: str) -> None:  # noqa: WPS603
+        """Delete key from headers case insensetively."""
+        return super().__delitem__(self._make_key(key))
+
+    @override
+    def __or__(
+        self,
+        _: Any,
+    ) -> NoReturn:
+        raise NotImplementedError
```
albertedwardson: implement update() where we will forbid bulk merging entirely or handle multiple keys correctly

search for related rfc where multiple values of a header are separated by comma

https://datatracker.ietf.org/doc/html/rfc2616#section-4.2

https://www.rfc-editor.org/rfc/rfc9651 ?

https://www.rfc-editor.org/rfc/rfc9110.html#name-field-names

django_modern_rest/response.py
```diff
@@ -162,7 +163,7 @@ def build_response(
     *,
     raw_data: Any,
     method: HTTPMethod | str,
-    headers: dict[str, str] | None = None,
```
albertedwardson, line 165: allow dict[str, str] as input and then convert it to HeaderDict?

tests/test_unit/test_controllers/test_to_response.py
```diff
@@ -26,7 +27,7 @@ def get(self) -> HttpResponse:
         """Tests that `.to_response` works correctly."""
         return self.to_response(
             ['a', 'b'],
-            headers={'X-Custom': 'value'},
+            headers=HeaderDict({'X-Custom': 'value'}),
```
sobolevn: Nope, this is clearly not the design we want. We want to allow users to use just `dict` if it works for them. We want to provide `HttpHeaders` for cases when it is needed.

django_modern_rest/headers.py
```diff
@@ -68,13 +74,179 @@ class HeaderSpec(_BaseResponseHeader):
...
+        Traceback (most recent call last):
+            ...
+        TypeError: Header keys must be `str`
+        >>> h['x-my-header'] = 1
+        Traceback (most recent call last):
+            ...
+        TypeError: Header values must be `str` or `Iterable[str]`
+    """
+
+    _titled_keys_cache: typing.ClassVar[dict[str, str]] = {}
```
albertedwardson: need lru, requests with arbitrary headers could lead to oom

tests/test_unit/test_controllers/test_to_response.py
```diff
@@ -41,7 +41,10 @@ def post(self) -> HttpResponse:
         """Tests that `.to_response` works with extra headers."""
         return self.to_response(
             ['a', 'b'],
-            headers={'X-Custom': 'value', 'Content-Type': 'application/json5'},
```
albertedwardson, line 44: revert

django_modern_rest/headers.py
```diff
@@ -68,13 +78,181 @@ class HeaderSpec(_BaseResponseHeader):
...
+            input_value = input_value.split(',')
+
+        if isinstance(input_value, (list, tuple, Iterable)):
+            try:
+                return [
+                    striped
+                    for subvalue in input_value  # pyright: ignore[reportUnknownVariableType]
+                    if (striped := subvalue.strip(_header_strip_chars))  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+                ]
+            except Exception as exc:
```
albertedwardson: possibly could be narrowed to TypeError (b"".strip("")) or AttributeError (when trying to call .strip on non-string)

django_modern_rest/headers.py
```diff
@@ -68,13 +78,181 @@ class HeaderSpec(_BaseResponseHeader):
...
+@typing.final
+class HeaderDict(UserDict[str, str]):  # noqa: WPS214
+    """
+    Dictionary-like container for HTTP headers with case-normalized keys.
+
+    Header keys are automatically converted to Title-Case format (e.g.
+    ``"content-type"`` -> ``"Content-Type"``). Values are stored as strings
+    separated by a comma to allow multi-valued headers.
+
+    Use `add(key, val)` or `update(key, vals)`
```
albertedwardson: I need to correct that, update have different semantics

sobolevn (commented):
Thank you! Looks great, I will update some small pieces and merge after #274 :)

django_modern_rest/headers.py
```diff
@@ -68,13 +78,181 @@ class HeaderSpec(_BaseResponseHeader):
...
+
+    @typing.override
+    def __contains__(self, key: typing.Any) -> bool:
+        """Check if key is present case insensetively."""
+        return self._make_key(key) in self.data
+
+    @typing.override
+    def __delitem__(self, key: str) -> None:  # noqa: WPS603
+        """Delete key from headers case insensetively."""
+        self.data.pop(self._make_key(key), None)
```
albertedwardson: I made a mistake here, __del__ shouldn't behave like .pop with default argument

django_modern_rest/headers.py
```diff
@@ -1,10 +1,20 @@
...
+from django_modern_rest.internal.strings import str_title_cached_interned
+
+if typing.TYPE_CHECKING:  # pragma: no cover
     from django_modern_rest.response import ResponseModification
     from django_modern_rest.serialization import BaseSerializer
 
+HeaderLike: typing.TypeAlias = typing.Union['HeaderDict', dict[str, str]]
+StrOrIterstr: typing.TypeAlias = str | Iterable[str]
+
+_header_strip_chars: str = string.whitespace + ','  # noqa: WPS336
```
albertedwardson: actually this is wrong too. this needs to be something like "".join(chr(x) for x in range(33))

Issue #275, opened 2025-11-05
Test that we support new styled type aliases as request / return types


sobolevn:
We have a special test case for different types that we support https://github.com/wemake-services/django-modern-rest/blob/998715103c375edf4270cbf89cb68d5fad10365e/tests/test_unit/test_validation/test_type_validation.py#L48-L64

This test is missing a case like

```python
type MyInt = int
```

Please, add it :)
Note, that `type X` is 3.12+ syntax, while we also support 3.11
So, this needs to be added with a guard and probably `exec('type MyInt = int')`

MaximGit1:
Hello, I can help

sobolevn:
Thanks a lot!

PR #278, opened 2025-11-05
Updated test_type_validation.py

MaximGit1:
Test that we support new styled type aliases as request / return types #275 

- Added a test case for the alias type.
- Added the `test_type_alias_defined` test to check coverage.

### DIFF
--- a/tests/test_unit/test_validation/test_type_validation.py
+++ b/tests/test_unit/test_validation/test_type_validation.py
@@ -1,7 +1,7 @@
 import sys
 from collections.abc import Callable
 from http import HTTPStatus
-from typing import Any, Literal, TypeAlias
+from typing import Any, Literal
 
 import pytest
 from django.http import HttpResponse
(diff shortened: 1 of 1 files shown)


tests/test_unit/test_validation/test_type_validation.py
```diff
@@ -144,3 +153,11 @@ class _Controller(Controller[serializer]):  # type: ignore[valid-type]
...
             endpoint.metadata.responses[HTTPStatus.OK],
         )
+
+
+def test_type_alias_defined() -> None:
+    """Ensure that `MyInt` type alias is defined.
+
+    MyInt: can be either a plain `int`, or a TypeAlias (Python 3.12+).
+    """
+    assert MyInt is int or getattr(MyInt, '__value__', None) is int
```
sobolevn:
```suggestion
```

tests/test_unit/test_validation/test_type_validation.py
```diff
@@ -14,6 +14,14 @@
 
 serializers: list[Any] = [PydanticSerializer]
 
+MyInt: Any = int  # for Pyright
+
+if sys.version_info >= (3, 12):
+    exec('type MyInt = int')  # noqa: S102, WPS421
+else:  # pragma: no cover
+    MyInt: TypeAlias = int
```
sobolevn:
```suggestion
```

PR #293, opened 2025-11-09
Add more examples run to the documentation #292

F0RRZZ:
I added more run commands for examples in the documentation.

Also added new parameter to `# run` command: `fail-with-body`. Before it was always enabled, because of that examples where need to show error didn't build.

Usage:
```
# run: {"fail-with-body": False}
```
true by default.

### DIFF
--- a/docs/examples/core_concepts/glossary.py
+++ b/docs/examples/core_concepts/glossary.py
@@ -28,5 +28,5 @@ class UserController(  # <- `Controller` definition
         return UserModel(uid=uuid.uuid4(), email=self.parsed_body.email)
 
 
-# run: {"controller": "UserController", "method": "get", "body": {"email": "user@wms.org"}, "url": "/api/user/", "curl_args": ["-D", "-"]}  # noqa: ERA001, E501
-# run: {"controller": "UserController", "method": "post", "body": {"email": "user@wms.org"}, "url": "/api/user/", "curl_args": ["-D", "-"]}  # noqa: ERA001, E501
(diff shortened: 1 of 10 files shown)


docs/examples/core_concepts/glossary.py
```diff
@@ -26,3 +26,7 @@ def get(self) -> UserModel:  # <- `Endpoint` definition
 
     def post(self) -> UserModel:  # <- `Endpoint` definition
         return UserModel(uid=uuid.uuid4(), email=self.parsed_body.email)
+
+
+# run: {"controller": "UserController", "method": "get", "body": {"email": "user@wms.org"}, "url": "/api/user/", "curl_args": ["-D", "-"]}  # noqa: ERA001, E501
```
sobolevn: We don't need `-D -` in default examples, we only need it when headers are important.

docs/pages/middleware.rst
```diff
@@ -163,7 +163,8 @@ Use with ``wrap_middleware``:
 
 .. literalinclude:: /examples/middleware/rate_limit.py
   :linenos:
-  :lines: 33-
+  :lines: 33-49
+  :no-run:
```
sobolevn, line 167: If we don't run it, maybe let's remove the command instead? Why don't we run it though?

F0RRZZ: This code was already run [on line 57](https://github.com/wemake-services/django-modern-rest/blob/master/docs/pages/middleware.rst?plain=1#L57), so I thought there was no need to re-run it. And the commands can't be deleted precisely because we're running this code before that.

sobolevn: oh, right :)

sobolevn (approved):
Thanks a lot, really nice!

PR #310, opened 2025-11-16
feat: extract metadata.{summary, description} from docstring if unset in payload

### LINKED ISSUE #289: feat: extract `EndpointMetadata.summary` and `EndpointMetadata.description` from `endpoint.__doc__` in `EndpointMetadataValidator` by default.
If the `EndpointMetadata.summary`  and `EndpointMetadata.description`  aren't specified by the current logic, get the field data from `endpoint.__doc__`. 
The docstring parsing function from django can be used. https://github.com/django/django/blob/c4e07f94ebc1f9eaa3dae7b3dc6a2b9832182a10/django/contrib/admindocs/utils.py#L32-L57

Also need to remove the todo. https://github.com/wemake-services/django-modern-rest/blob/36cbb7b33cbeee8631e468fede713ac7809a7665/django_modern_rest/validation/endpoint_metadata.py#L517-L518

Assign this issue to me, please)

ApostolFet:
# I have made things!
Add fallback to docstring for EndpointMetadata.summary/description

- Implemented docstring parsing using Django’s utility logic (adapted from django.contrib.admindocs.utils).
- Added logic in EndpointMetadataValidator to extract summary and description from `endpoint.__doc__` by default.
- Removed the existing TODO comment about this functionality.
- Ensured fallback only triggers when summary and description aren’t provided via other means.

## Checklist

- [X] I have created at least one test case for the changes I have made

## Related issues
- Closes #289

### DIFF
--- a/tests/test_unit/test_metadata/test_resolve_description.py
+++ b/tests/test_unit/test_metadata/test_resolve_description.py
@@ -16,6 +16,45 @@ class _PaylodDescriptionController(
         raise NotImplementedError
 
 
+class _PaylodDocDescriptionController(
+    Controller[PydanticSerializer],
+):
+    @modify(
+        summary='Summary from payload.',
+        description='Description from payload.',
+    )
+    def get(self) -> list[int]:
+        """Summary from docstring.
+
+        Description from docstring.
+        """
+        raise NotImplementedError
+
+
+class _PaylodDocOnlySummaryController(
+    Controller[PydanticSerializer],
+):
+    @modify(summary='Summary from payload.')
+    def get(self) -> list[int]:
+        """Summary from docstring.
+
+        Description from docstring.
+        """
+        raise NotImplementedError
+
+
+class _PaylodDocOnlyDescriptionController(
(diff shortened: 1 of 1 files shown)


tests/test_unit/test_metadata/test_resolve_description.py
```diff
@@ -0,0 +1,99 @@
+import pytest
+
+from django_modern_rest import modify
+from django_modern_rest.controller import Controller
+from django_modern_rest.plugins.pydantic import PydanticSerializer
+
+
+class _PaylodDescriptionController(
```
sobolevn, line 8: In these tests you never actually test priorities of the summery generation :)

Please, provide clear values so we can be sure that all priorities work.

sobolevn (commented):
Thanks!

tests/test_unit/test_metadata/test_resolve_description.py
ApostolFet: Added a test case to check priority

Issue #367, opened 2026-01-16
Support `pyrefly`


sobolevn:
https://pyrefly.org is already quite stable.

We use it in https://github.com/typeddjango/django-stubs/blob/master/pyrefly.toml with no major complains.
I really want to add this in CI as soon as possible.

PR #381, opened 2026-01-24
Add pyrefly check #367

F0RRZZ:
- Issue: https://github.com/wemake-services/django-modern-rest/issues/367

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,6 @@ repos:
     hooks:
       - id: shellcheck
         args: ["--severity=style"]
-
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.14.13
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,6 +25,7 @@ repos:
     hooks:
       - id: shellcheck
         args: ["--severity=style"]
+
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.14.13
     hooks:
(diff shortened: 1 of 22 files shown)


Makefile
```diff
@@ -9,6 +9,7 @@ lint:
 	poetry run ruff check --exit-non-zero-on-fix
 	poetry run ruff format --check --diff
 	poetry run flake8 .
+	poetry run pyrefly check .
```
sobolevn: Please, move this to `type-check` target

.pre-commit-config.yaml
```diff
@@ -25,6 +25,7 @@ repos:
     hooks:
       - id: shellcheck
         args: ["--severity=style"]
+
```
sobolevn: Unrelated change

pyproject.toml
```diff
@@ -304,6 +305,15 @@ venvPath = "."
 venv = ".venv"
 
 
+[tool.pyrefly]
```
sobolevn, line 308: Since `pyrefly` has quite a few false-positives for now, let's only enable it on `django_modern_rest/` and maybe (?) on `typesafety/`

Other files can be ignored for now, please remove all changes to them :)

pyproject.toml
```diff
@@ -87,6 +87,7 @@ polyfactory = "^3.2"
 
 httpx = "^0.28"
 xmltodict = "^1.0"
+pyrefly = "^0.49.0"
```
```suggestion
pyrefly = "^0.49"
```

Please run `poetry update` again after this change. And please move it to be near `pyright`

sobolevn (approved):
Great, thanks!

PR #384, opened 2026-01-25
json_dumps internal helper

### LINKED ISSUE #383: Create `json_dumps` internal helper
We have to dump `json` in some internal places of our code: https://github.com/wemake-services/django-modern-rest/blob/9e6132e80f5387f9c7f8b7fe5649a2da1a4f5d31/django_modern_rest/openapi/renderers/base.py#L16-L31

But, right now we don't use `msgspec` there if it is available. But, we should :)
So, let's create a new `internal/json` file and make a helper there. Then use it in `base.py`

bagowix:
# I have made things!

json_dumps uses msgspec if available, otherwise json

## Checklist

## Related issues

- Closes #383

### DIFF
--- a/django_modern_rest/internal/json.py
+++ b/django_modern_rest/internal/json.py
@@ -23,12 +23,12 @@ def _wrap_bytes_dumper(dumper: Callable[[ConvertedSchema], bytes]) -> _Dumper:
     return wrapper
 
 
-try:  # noqa: WPS229
+try:
     import msgspec
-
-    _json_dumps: _Dumper = _wrap_bytes_dumper(msgspec.json.encode)
 except ImportError:  # pragma: no cover
-    _json_dumps = json.dumps
+    _json_dumps: _Dumper = json.dumps
+else:
+    _json_dumps = _wrap_bytes_dumper(msgspec.json.encode)
 
 
 def json_dumps(schema: ConvertedSchema) -> SerializedSchema:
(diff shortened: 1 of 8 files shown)


tests/test_unit/test_openapi/test_renderers.py
```diff
@@ -150,3 +157,47 @@ def test_renderer_with_simple_decorators(
 
     assert isinstance(renderer.decorators, list)
     assert len(renderer.decorators) == expected_len
+
+
+def reload_json_and_base_modules() -> ModuleType:
```
sobolevn: we don't need these tests :)

sobolevn (commented):
other than one comment - lgtm

tests/test_unit/test_openapi/test_renderers.py
bagowix: removed :)

django_modern_rest/internal/json.py
```diff
@@ -0,0 +1,45 @@
...
+
+    return wrapper
+
+
+try:  # noqa: WPS229
+    import msgspec
+
+    _json_dumps: _Dumper = _wrap_bytes_dumper(msgspec.json.encode)
+except ImportError:  # pragma: no cover
+    _json_dumps = json.dumps
```
sobolevn:
```suggestion
try:
    import msgspec
except ImportError:  # pragma: no cover
    _json_dumps: _Dumper = json.dumps
else:
    _json_dumps = _wrap_bytes_dumper(msgspec.json.encode)
```

sobolevn (approved):

Issue #55, opened 2025-10-16
Thread-safety should be a default


sobolevn:
We need a way to ensure that our framework is thread safe. We should test for it and claim the threading safety as a feature.

PR #386, opened 2026-01-26
chore: force free gil behavior

thepabloaguilar:
# I have made things!

This PR forces the Free GIL behavior for the versions that supports it, in the current CI setup is only `3.14t`. When the Python runtime identifies a dependency not marked explicit as _free gil supported_ it fallback to the default and use GIL, setting the `PYTHON_GIL=0` forces to use the Free GIL version even not having all dependencies with that mark.

I think would be great to have this forced by default so we can see future problems and fix them (or at least try to fix/open an issue), also this could be related to desired thread safe behavior for this framework.

## Checklist

## Related issues

Refs #55

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,9 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.11', '3.12', '3.13', '3.14', '3.14t']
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Let's try this out

Issue #394, opened 2026-01-27
Design authentication and permissions


sobolevn:
I like how drf has two different concepts for this: https://www.django-rest-framework.org/api-guide/authentication/

See:

```python
from rest_framework.authentication import SessionAuthentication, BasicAuthentication
from rest_framework.permissions import IsAuthenticated
from rest_framework.response import Response
from rest_framework.views import APIView

class ExampleView(APIView):
    authentication_classes = [SessionAuthentication, BasicAuthentication]
    permission_classes = [IsAuthenticated]

    def get(self, request, format=None):
        content = {
            'user': str(request.user),  # `django.contrib.auth.User` instance.
            'auth': str(request.auth),  # None
        }
        return Response(content)
```

`django-ninja` on the other hand only has `auth=` parameter for everything: https://django-ninja.dev/guides/authentication/

## Criterial

Our approach:
- [x] Must be configurable on all levels
- [x] Must have all the required openapi schema information
- [x] Must support sync and async authentication methods

I think that we should stay away from `@login_required()` and `AUTH_BACKENDS`, because API should not mess with the original framework. Keeping security configuration separate from the framework is a good thing.

Here's what the base example should look like:

```python
class UserController(
    Body[_UserDocument],
    Controller[PydanticSerializer],
):
    authentication = [JWTAuth]
    permissions = [IsAuthenticated]

    def post(self) -> _UserDocument[_UserOutputData]:
        return _UserDocument(
            uid=uuid.uuid4(),
        )
```

OpenAPI security definition: https://learn.openapis.org/specification/security.html

## Builtin types to ship

We need several builtin auth providers:
- [x] basic auth
- [x] django session
- [x] jwt

Others can be built by the community as plugins.

Refs https://github.com/wemake-services/django-modern-rest/issues/1#issuecomment-3384860676

PR #400, opened 2026-01-28
Add security and auth

sobolevn:
Initial implementation.

Refs #394


django_modern_rest/security/http.py
```diff
@@ -0,0 +1,119 @@
...
+        elif len(parts) == 2 and parts[0].lower() == 'basic':
+            encoded = parts[1]
+        else:
+            return None
+
+        try:
+            username, password = b64decode(encoded).decode().split(':', 1)
+        except Exception:
+            return None
+        return unquote(username), unquote(password)
```
milssky, line 70: If the user will use the password `Super%20Password`. It can change it like

```python
In [1]: from urllib.parse import unquote
In [2]: import base64
In [3]: creds = 'user:Super%20Password'
In [4]: header_value = base64.b64encode(creds.encode()).decode()
In [5]: decoded_str = base64.b64decode(header_value).decode()
In [6]: extracted_user, extracted_pass = decoded_str.split(':', 1)
In [7]: final_pass = unquote(extracted_pass)
In [8]: final_pass
Out[8]: 'Super Password'
```

This combination is quite possible when using password generators.

sobolevn: This is not a problem, because you are choosing a password for basic auth.

PR #411, opened 2026-01-30
feat: improved Makefile with better command handling and help message

Mr-Sunglasses:
# I have made things!

## Checklist


sobolevn (approved):
Thank you! You might also contribute the same thing to:
- https://github.com/wemake-services/wemake-python-styleguide/blob/master/Makefile
- https://github.com/wemake-services/wemake-python-package/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/Makefile

PR #416, opened 2026-01-31
add `pytest-freethreaded` plugin

thepabloaguilar:
# I have made things!

This PRs introduces a new pytest plugin called [pytest-freethreaded](https://github.com/tonybaloney/pytest-freethreaded) which runs each test multiple times using threads and potentially tell us what are the areas we need to look for non thread safety things.

I've added a constraint to only install this plugin for Python 3.14+ versions!

> [!NOTE]
> This PR is not intended to be merge, it's more a exploration. The fixes that would come from the results here will be fixed in separated PRs and then we rebase this branch to get the fixed stuff

## Checklist

## Related issues

Refs #55


thepabloaguilar:
I've looked at some errors, the errors are happening because the plugin does not run with `asyncio` support 😞

thepabloaguilar:
Ok, in fact it's just because of the pytest plugin initialization order to be arbitrary

sobolevn:
It now has a lot of merge conflicts :(

PR #437, opened 2026-02-04
Schema generator

milssky:
# I have made things!

I'm definitely missing something, but all the basic functionality is there. Queries, parameters, requests and responses are generated.

## Checklist

- [x] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)
- [x] I have created at least one test case for the changes I have made
- [ ] I have updated the documentation for the changes I have made

PR #456, opened 2026-02-07
`TypedDictExtractor`

milssky:
# I have made things!

Fixed type detection errors and a new `TypedDictExtractor`.

## Checklist

## Related issues

Refs #437

### DIFF
--- a/django_modern_rest/openapi/extractors/stdlib.py
+++ b/django_modern_rest/openapi/extractors/stdlib.py
@@ -14,51 +14,31 @@ class TypedDictExtractor(FieldExtractor[type[dict[str, Any]]]):
     def is_supported(cls, source: Any) -> bool:
         return is_typeddict(source)
 
+    # TODO: What do we do with __extra_items__ here?
     @override
     def extract_fields(
         self,
         source: type[dict[str, Any]],
     ) -> list[FieldDefinition]:
+        definitions: list[FieldDefinition] = []
         try:
             type_hints = get_type_hints(source)
         except (ValueError, TypeError, AttributeError):
-            return []
+            return definitions
 
-        required_keys, optional_keys = self._get_keys(source)
-        return self._create_definitions(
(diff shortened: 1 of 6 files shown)


django_modern_rest/openapi/extractors/stdlib.py
```diff
@@ -0,0 +1,64 @@
...
+        optional_keys: frozenset[str],
+    ) -> list[FieldDefinition]:
+        return [
+            FieldDefinition(
+                name=name,
+                annotation=annotation,
+                extra_data={'is_required': name in required_keys},
+            )
+            for name, annotation in type_hints.items()
+            if name in required_keys or name in optional_keys
```
sobolevn: Is there any way to create a key that is not in these two groups?

tests/test_unit/test_openapi/test_extractors/test_typeddict_field_extractor.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class _TestTypedDict(TypedDict):
+    name: str
+    age: int
+
+
+class _TestTypedDictRequired(TypedDict, total=False):
+    name: str
+    age: int
```
sobolevn, line 20: Please test:
- Inherited typed dicts
- `Required`
- `NotRequired`
- `ReadOnly`
- `closed=True`
- `extra_items=X`

If `pydantic` / `msgspec` support inline typed dict, then we also need to support them. (this can be a separate task)

django_modern_rest/openapi/extractors/stdlib.py
milssky: In general, you can simply define required keys based on the following considerations: all keys that are not in `__required_keys__` are optional. I returned the code to this state and added a comment about `__optional_keys__`.

sobolevn (approved):

PR #474, opened 2026-02-09
Add `csrf_token` to openapi view

milssky:
# I have made things!

## Checklist

### DIFF
--- a/django_modern_rest/openapi/views.py
+++ b/django_modern_rest/openapi/views.py
@@ -2,13 +2,19 @@ from collections.abc import Callable
 from typing import Any, ClassVar, final
 
 from django.http import HttpRequest, HttpResponse, HttpResponseBase
+from django.utils.decorators import method_decorator
 from django.views import View
+from django.views.decorators.csrf import ensure_csrf_cookie
 from typing_extensions import override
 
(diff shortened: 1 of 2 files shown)


django_modern_rest/openapi/views.py
```diff
@@ -2,13 +2,19 @@
...
 from django.views import View
+from django.views.decorators.csrf import ensure_csrf_cookie
 from typing_extensions import override
 
 from django_modern_rest.openapi.converter import ConvertedSchema
 from django_modern_rest.openapi.renderers import BaseRenderer
 
 
+@method_decorator(
+    ensure_csrf_cookie,  # pyrefly: ignore[bad-argument-type]
```
kondratevdev, line 15: I'm not sure if it's always right to set csrf in the view.

milssky: we can add it into renderer. But it's even worse.

sobolevn (commented):
Question: how current code would behave if we disable `'django.middleware.csrf.CsrfViewMiddleware'` middleware?

sobolevn (approved):
I am fine with the proposed solution :)

kondratevdev (approved):

PR #503, opened 2026-02-12
CSRF support

### LINKED ISSUE #464: Figure out what's the case with CSRF
Right now all `curl` and swagger requests to our example app trigger CSRF error.
I need to figure out how DRF handles this.

kondratevdev:
# I have made things!

## Checklist

## Related issues

Closes #464


django_modern_rest/security/django_session/auth.py
```diff
@@ -16,7 +29,29 @@
...
+        return reason
+
+
+def _enforce_csrf(request: HttpRequest) -> None:
+    """Perform CSRF validation using ``_CheckCSRF``."""
+    check = _EnsureCsrfToken(lambda _: None)  # type: ignore[arg-type]
+    check.process_request(request)
+    reason = check.process_view(request, None, (), {})  # type: ignore[arg-type]
+    if reason:
+        raise PermissionDeniedError(f'CSRF Failed: {reason}')
```
sobolevn: Can we raise `APIError` instead?

We use named exceptions when they are very specific. This one is not really that specific.

django_test_app/server/apps/models_example/services.py
```diff
@@ -4,8 +4,15 @@
 from server.apps.models_example.models import Role, Tag, User
 
 
+class UniqueEmailError(Exception):
+    """Email must be unique."""
+
+
 def user_create_service(user_schema: serializers.UserCreateSchema) -> User:
     """This is a function just for the demo purpose, it is usually a class."""
+    if User.objects.filter(email=user_schema.email).exists():
```
sobolevn: This should not be outside of a transaction. Moreover, it will trigger `IntegrityError`, it is safer.

django_test_app/server/apps/models_example/services.py
```diff
@@ -4,8 +4,15 @@
...
+        raise UniqueEmailError
+
     with transaction.atomic():
         role = Role.objects.create(name=user_schema.role.name)
         user = User.objects.create(
```
sobolevn:
```suggestion
        try:
            user = User.objects.create(
            # ...
        except IntegrityError:
            raise UniqueEmailError from None
```

sobolevn (approved):

PR #507, opened 2026-02-13
Support Allow header

kondratevdev:
# I have made things!

## Checklist


django_modern_rest/internal/middleware_wrapper.py
sobolevn: can we type it as `self: Controller[BaseSerializer]`?

sobolevn (approved):
Great, thanks!

PR #511, opened 2026-02-14
Operation builder class

kondratevdev:
# I have made things!

## Checklist


django_modern_rest/openapi/builders/__init__.py
kondratevdev, line 1: For the same naming as in the main part of the code, I suggest renaming generators to builders.
Thus, we will include `*_builder_cls` to the components.

sobolevn (approved):

PR #521, opened 2026-02-15
Support `schemathesis` `status_code_conformance` check

kondratevdev:
# I have made things!

## Checklist


sobolevn (commented):
Please, add a unit test for component metadata. See `test_valid_api_error_contoller_level` as an example.

We need to be sure that `NOT_FOUND` is present.

sobolevn (approved):

PR #525, opened 2026-02-16
Made the `KwargMapper` configurable

kondratevdev:
# I have made things!

## Checklist

### DIFF
--- a/django_modern_rest/openapi/mappers.py
+++ b/django_modern_rest/openapi/mappers.py
@@ -1,3 +1,4 @@
+import contextlib
 from collections import OrderedDict, defaultdict, deque
 from collections.abc import (
     Iterable,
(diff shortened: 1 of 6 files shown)


django_modern_rest/openapi/mappers.py
```diff
@@ -1,3 +1,4 @@
+import contextlib
```
milssky, line 1: Looks like we need a `mappers` package =)

milssky (commented):
Could it be that the kwargs in the dtos from different libraries may differ? Could it be possible to provide for this?

See TODO: 

https://github.com/wemake-services/django-modern-rest/pull/525/changes#diff-f3272c356e26e2dcbec09a7f5f9cd98666c590da8a73a3ae26686dcee54f6d13L57-L59

django_modern_rest/openapi/mappers.py
kondratevdev: I think is module currently enough for two classes

milssky (approved):

PR #529, opened 2026-02-17
Security scheme

milssky:
# I have made things!

Suddenly it turned out that the Header object in the specification does not have in and name =D
image
Deleted it. 

And I made a registry for security schemes for consistency.

## Checklist


django_modern_rest/openapi/objects/header.py
kondratevdev: Please, fix docs

sobolevn (approved):

PR #523, opened 2026-02-15
Enable `positive_data_acceptance` in `schemathesis` #496

F0RRZZ:
- Issue: https://github.com/wemake-services/django-modern-rest/issues/496

PR #540, opened 2026-02-19
Enable positive data acceptance

### LINKED ISSUE #496: Enable `positive_data_acceptance` in `schemathesis`
At the moment, this check is disabled in the [configuration](https://github.com/wemake-services/django-modern-rest/blob/38febfe3e158b4f1a5646832c240155f91c09a12/schemathesis.toml#L15).

We need to enable it (delete the line, because by default we have everything enabled) and fix the error that this check catches.

milssky:
# I have made things!

## Checklist

## Related issues

Refs #523 
Closes #496

### DIFF
--- a/dmr/components.py
+++ b/dmr/components.py
@@ -259,6 +259,20 @@ class Query(ComponentParser, Generic[_QueryT]):
 
     We don't inference this value in any way, it is up to users to set.
     Inspecting annotations is hard and produce a lot of errors.
+
+    Additionally, users can customize whether the string literal ``'null'``
+    should be cast to Python's ``None``.
+    To do so, use ``__dmr_cast_null__`` optional attribute.
+    It defaults to ``True``.
+    If set to ``False``, ``'null'`` will be treated as a regular string.
+
+    .. code:: python
+
+        >>> class SearchQuery(pydantic.BaseModel):
+        ...     __dmr_cast_null__: ClassVar[bool] = False
+        ...
+        ...     query: str  # will be 'null' if ?query=null is sent
+
     """
 
     parsed_query: _QueryT
@@ -278,9 +292,15 @@ class Query(ComponentParser, Generic[_QueryT]):
             '__dmr_force_list__',
             frozenset(),
         )
+        cast_null: bool = getattr(
+            field_model,
+            '__dmr_cast_null__',
+            True,
+        )
         return convert_multi_value_dict(
             blueprint.request.GET,
             force_list,
+            cast_null=cast_null,
         )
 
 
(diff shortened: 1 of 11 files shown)


dmr/components.py
```diff
@@ -278,7 +297,33 @@ def provide_context_data(
...
+            blueprint.request.GET,
+            force_list,
+        )
+
+        return {
+            param_key: cls._clean_param_value(param_value)
+            for param_key, param_value in request_params.items()
+        }
+
+    @classmethod
```
sobolevn: Can we move all of this into `convert_multi_value_dict`? I want to have clean component definitions.

tests/test_integration/test_openapi/test_schema.py
```diff
@@ -19,7 +19,11 @@ def api_schema(db: Any) -> 'OpenApiSchema':
     return st.openapi.from_wsgi(reverse('openapi:json'), application)
 
 
-schema = st.pytest.from_fixture('api_schema')
+# NOTE: We skip negotiation tests because our implementation of
```
sobolevn:
```suggestion
# TODO: We skip negotiation tests because our implementation of
```

Please, open a new issue about it. We can provide a full-featured parser later on.

dmr/components.py
milssky: I thought about it, but I decided that the utility function that triggers MultipleValueDict should be cleaner =D
That's why I decided that `Query` should know how to work with query parameters.

The decision is yours.

sobolevn: `convert_multi_value_dict` is the proper place :)

dmr/components.py
```diff
@@ -257,25 +257,6 @@ class Query(ComponentParser, Generic[_QueryT]):
     This will parse a query like ``?query=text&query=match&reversed=1``
     into the provided model.
 
-    Additionally, this parser automatically converts the string literal
```
sobolevn: You can keep this doc, it is still useful :)

dmr/components.py
```diff
@@ -257,25 +257,6 @@ class Query(ComponentParser, Generic[_QueryT]):
...
-
-        >>> class NullableQuery(pydantic.BaseModel):
-        ...     category: str | None
-        ...     tags: list[str | None]
-        ...
-        ...     __dmr_force_list__: ClassVar[frozenset[str]] = frozenset((
-        ...         'tags',
-        ...     ))
-
-    Will parse a request like ``?category=null&tags=python&tags=null``
```
sobolevn: Maybe let's add `__dmr_cast_null__` with `True` as default to control this? Because in some cases `'null'` might be required as a regular string.

milssky: An excellent idea, you have.

I'll do it =)

sobolevn (approved):

PR #578, opened 2026-02-27
Issue 462

### LINKED ISSUE #462: JWT token blacklist app: design conditional apps
We need to design how we can add conditional django apps that have models to the framework.
For example, right now we can add an app to blacklist some jwt tokens.

What's interesting about that?
1. We need to design this app in a way that it is not required by default, with no visible parts
2. We need to design it that it will be very easy to install and use. For example: it should provide a mixin class that can be used together with `JWTSyncAuth` / `JWTAsyncAuth` and a pre-existing common subclass

So, usage will look something like:

```python
from django_modern_rest.some.place.for.app.models import BlacklistedJWTToken

class BlacklistMixin:
     def check_auth(self, user: 'AbstractBaseUser', token: JWTToken) -> None:
        super().check_user()
        if BlacklistedJWTToken.objects.filter(jti=token.jti).exists():
            # Token was blacklisted:
            raise NotAuthenticatedError
```

We use `jti` to identify tokens, as other jwt's do.
It is now possible after https://github.com/wemake-services/django-modern-rest/pull/461

Sources of inspiration: 
- https://github.com/jazzband/djangorestframework-simplejwt/blob/master/rest_framework_simplejwt/token_blacklist/models.py
- https://django-rest-framework-simplejwt.readthedocs.io/en/latest/blacklist_app.html

Chyrets:
# I have made things!

## Checklist

## Related issues

Closes #462


django_test_app/server/settings.py
```diff
@@ -48,6 +48,7 @@
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'dmr.apps.DjangoModernRestConfig',
+    'dmr.security.blocklist',
```
kondratevdev, line 51: Isn't enough `dmr.apps.DjangoModernRestConfig` to initialize models?

dmr/security/blocklist/auth.py
```diff
@@ -0,0 +1,94 @@
...
+        """Add token to the blocklist."""
+        jti = token.jti
+        exp = token.exp
+        user = self.get_user(token)
+
+        return self.blocklist_model.objects.get_or_create(
+            user=user,
+            jti=jti,
+            expires_at=exp,
+        )
```
kondratevdev, line 61:
```suggestion
        return self.blocklist_model.objects.get_or_create(
            user=self.get_user(token),
            jti=token.jti,
            expires_at=token.exp,
        )
```

dmr/security/blocklist/auth.py
```diff
@@ -0,0 +1,94 @@
...
+        """Add token to the blocklist."""
+        jti = token.jti
+        exp = token.exp
+        user = await self.get_user(token)
+
+        return await self.blocklist_model.objects.aget_or_create(
+            user=user,
+            jti=jti,
+            expires_at=exp,
+        )
```
kondratevdev:
```suggestion
        user = await self.get_user(token)
        return await self.blocklist_model.objects.aget_or_create(
            user=user,
            jti=token.jti,
            expires_at=token.exp,
        )
```

tests/test_unit/test_security/test_blocklist/test_models.py
```diff
@@ -0,0 +1,55 @@
+import datetime as dt
+import secrets
+
+import pytest
+from django.contrib.auth.models import User
+from faker import Faker
+
+from dmr.security.blocklist.models import BlocklistedJWTToken
+
+jti = secrets.token_hex()
```
kondratevdev:
```suggestion
_JTI = secrets.token_hex()
```

tests/test_unit/test_security/test_blocklist/test_models.py
```diff
@@ -0,0 +1,55 @@
...
+expires_at = dt.datetime.now(dt.UTC) + dt.timedelta(days=1)
```
kondratevdev:
```suggestion
_EXPIRES_AT = dt.datetime.now(dt.UTC) + dt.timedelta(days=1)
```

tests/test_unit/test_security/test_blocklist/test_models.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@pytest.mark.django_db
+def test_token_expires_at_field(token: BlocklistedJWTToken) -> None:
+    """Test expires_at field."""
+    assert token.expires_at == expires_at
+
+
+@pytest.mark.django_db
+def test_token_str(token: BlocklistedJWTToken) -> None:
```
kondratevdev, line 43: It looks like we can make one test `test_model_fields` and combine all the previous tests.

tests/test_unit/test_security/test_blocklist/test_blocklist_auth.py
```diff
@@ -0,0 +1,207 @@
...
+        },
+    )
+
+    response = await dmr_async_rf.wrap(
+        _BlocklistAsyncController.as_view()(request),
+    )
+
+    assert isinstance(response, HttpResponse)
+    assert response.headers == {'Content-Type': 'application/json'}
+    assert response.status_code == HTTPStatus.OK
```
kondratevdev, line 246: Please, also check `response.content` using `snapshot()`

tests/test_unit/test_security/test_blocklist/test_blocklist_auth.py
```diff
@@ -0,0 +1,207 @@
...
+def test_add_to_blocklist(
+    build_user_token: _TokenBuilder,
+) -> None:
+    """Ensure that blocklist method works."""
+    token = build_user_token()
+    auth = MyJWTSyncAuth()
+    decoded_token = auth.decode_token(token)
+    auth.blocklist(decoded_token)
+
+    assert auth.blocklist_model.objects.filter(jti=decoded_token.jti).exists()
```
kondratevdev:
```suggestion
    token = build_user_token()
    auth = MyJWTSyncAuth()
    decoded_token = auth.decode_token(token)

    auth.blocklist(decoded_token)

    assert auth.blocklist_model.objects.filter(jti=decoded_token.jti).exists()
```

Separating this into Arrange, Act, and Assert blocks would make it easier to read.

django_test_app/server/settings.py
Chyrets: without this model BlocklistedJWTToken is considered abstract

sobolevn: yes, app must be always explicit.

tests/test_unit/test_security/test_blocklist/test_models.py
Chyrets: i can combine jti and expires_at tests but not user

django_test_app/server/settings.py
```diff
@@ -12,6 +12,12 @@
 
 from pathlib import Path
 
+import django_stubs_ext
+
+# Patch all the types we use as Generic in django.
+# https://github.com/typeddjango/django-stubs/tree/master/ext
+django_stubs_ext.monkeypatch()
```
Chyrets: without this line when testing django_test_app python version below 3.14 gives error 
TypeError: type 'CharField' is not subscriptable

sobolevn: Nope, this is not correct. Please, use `'CharField[...]'

We can't ship this to users requiring `django_stubs_ext.monkeypatch()`

tests/test_unit/test_security/test_blocklist/test_models.py
kondratevdev: You can just do one test (not parameterized) with several assert :)

dmr/security/blocklist/models.py
```diff
@@ -0,0 +1,36 @@
...
+class BlocklistedJWTToken(models.Model):
+    """Model for Blocklisted token."""
+
+    user: 'models.ForeignKey[Any, Any]' = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+    )
+    jti: 'models.CharField[Any, Any]' = models.CharField(
```
sobolevn:
```suggestion
    jti: 'models.CharField[str | int | Combinable, str]' = models.CharField
```

Please, copy correct annotations from https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/fields/__init__.pyi

`User` should be probably annotated as `AbstractBaseUser`, not `Any`

dmr/security/blocklist/models.py
```diff
@@ -0,0 +1,36 @@
...
+    updated_at: 'models.DateTimeField[Any, Any]' = models.DateTimeField(
+        auto_now=True,
+    )
+
+    class Meta:
+        abstract = BlocklistConfig.name not in settings.INSTALLED_APPS
+
+    @override
+    def __str__(self) -> str:
+        return f'Token for {self.user} {self.jti}'
```
sobolevn:
```suggestion
        return f'Blocked JWT token for {self.user} {self.jti}'
```

tests/test_unit/test_security/test_blocklist/test_blocklist_auth.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+_TokenBuilder: TypeAlias = Callable[..., str]
+
+
+@pytest.fixture
+def build_user_token(user: User, settings: LazySettings) -> _TokenBuilder:
+    """Token factory for tests."""
+
+    def factory(**kwargs: Any) -> str:
```
sobolevn: let's type kwargs, there are just several values.

tests/test_unit/test_security/test_blocklist/test_blocklist_auth.py
```diff
@@ -0,0 +1,237 @@
...
+from dmr.security.blocklist.auth import (
+    JWTTokenBlocklistAsyncMixin,
+    JWTTokenBlocklistSyncMixin,
+)
+from dmr.security.jwt.auth import JWTAsyncAuth, JWTSyncAuth
+from dmr.security.jwt.token import JWTToken
+from dmr.test import DMRAsyncRequestFactory, DMRRequestFactory
+
+_EXP = dt.datetime.now(dt.UTC) + dt.timedelta(days=1)
+_JTI = secrets.token_hex()
```
sobolevn:
```suggestion
_EXP: Final = dt.datetime.now(dt.UTC) + dt.timedelta(days=1)
_JTI: Final = secrets.token_hex()
```

tests/test_unit/test_security/test_blocklist/test_blocklist_auth.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+class _JWTTokenKwargs(TypedDict):
+    exp: NotRequired[dt.datetime]
+    iat: NotRequired[dt.datetime]
+    iss: NotRequired[str]
+    aud: NotRequired[str | Sequence[str]]
+    jti: NotRequired[str]
+    extras: NotRequired[dict[str, Any]]
+    leeway: NotRequired[int]
```
sobolevn:
```suggestion
class _JWTTokenKwargs(TypedDict, total=False):
    exp: dt.datetime
    iat: dt.datetime
    iss: str
    aud: str | Sequence[str]
    jti: str
    extras: dict[str, Any]
    leeway: int
```

dmr/security/blocklist/auth.py
```diff
@@ -0,0 +1,87 @@
...
+    def blocklist(
+        self: _JWTSyncAuth,
+        token: JWTToken,
+    ) -> tuple[BlocklistedJWTToken, bool]:
+        """Add token to the blocklist."""
+        return self.blocklist_model.objects.get_or_create(
+            user=self.get_user(token),
+            jti=token.jti,
+            expires_at=token.exp,
+        )
```
sobolevn, line 61:
```suggestion
    ) -> tuple[BlocklistedJWTToken, bool]:
        """Add token to the blocklist."""
        return self.blocklist_model.objects.get_or_create(
            jti=token.jti,
            defaults={
                'user': self.get_user(token),
                'expires_at': token.exp,
            },    
        )
```

`jti` is a unique index, it is enough for the query, but when creating a new model, we need extra data. See https://docs.djangoproject.com/en/6.0/ref/models/querysets/#get-or-create

dmr/security/blocklist/auth.py
```diff
@@ -0,0 +1,87 @@
...
+    async def blocklist(
+        self: _JWTAsyncAuth,
+        token: JWTToken,
+    ) -> tuple[BlocklistedJWTToken, bool]:
+        """Add token to the blocklist."""
+        user = await self.get_user(token)
+        return await self.blocklist_model.objects.aget_or_create(
+            user=user,
+            jti=token.jti,
+            expires_at=token.exp,
```
sobolevn: Please, do the same here

tests/test_unit/test_security/test_blocklist/test_blocklist_auth.py
```diff
@@ -0,0 +1,237 @@
...
+@pytest.mark.django_db
+def test_add_to_blocklist(
+    build_user_token: _TokenBuilder,
+) -> None:
+    """Ensure that blocklist method works."""
+    token = build_user_token(exp=_EXP, jti=_JTI)
+    auth = MyJWTSyncAuth()
+    decoded_token = auth.decode_token(token)
+
+    auth.blocklist(decoded_token)
```
sobolevn: Please, assert the return type. Here and for `await auth.blocklist` as well.

setup.cfg
```diff
@@ -50,9 +50,12 @@ per-file-ignores =
...
   # Ignore autogenerated migrations:
   django_test_app/server/apps/*/migrations/*.py: WPS102, WPS301, WPS432, WPS458
+  dmr/security/blocklist/migrations/*.py: WPS102, WPS301, WPS432, WPS458
   # Disable some complexity checks for tests:
   django_test_app/server/apps/*/views.py: WPS202
   # Enable string overuse tests, nested classes, and multiline strings:
   tests/*.py: WPS202, WPS204, WPS211, WPS218, WPS226, WPS430, WPS431, WPS462
   # Enable to import annotations from __future__ for Sphinx extensions
   docs/tools/sphinx_ext/*.py: WPS201, WPS202, WPS204, WPS226
+  # Disable some lints for django models
```
```suggestion
  # Disable some lints for django models:
```

sobolevn (commented):
We are almost there! Great work, @Chyrets 👏

tests/test_unit/test_security/test_blocklist/test_blocklist_auth.py
```diff
@@ -0,0 +1,251 @@
...
+    decoded_token = auth.decode_token(token)
+
+    blocklisted_token, created = await auth.blocklist(decoded_token)
+
+    assert blocklisted_token.jti == decoded_token.jti
+    assert blocklisted_token.expires_at == decoded_token.exp
+    assert str(
+        getattr(blocklisted_token.user, auth.user_id_field),
+    ) == auth.claim_from_token(decoded_token)
+    assert created is True
```
sobolevn, line 202: Please, add a test with double `blocklist()` call. Test that second `created is False`

sobolevn (approved):
Great work, thanks a lot!

Issue #569, opened 2026-02-26
Big refactoring of `openapi/`


sobolevn:
We have our first prototype of `openapi` spec generation implementation in `openapi/`, however, it is far from perfect.

I would like to refactor it to solve these problems:

- [ ] We need `build_schema` as a public API, not an internal helper
- [ ] I would prefer to have all current docs renderers as regular views, not as magic in `openapi_spec`

It will transform:

```python
urlpatterns = [
    path(
        'docs/',
        openapi_spec(router, renderers=[JsonRenderer(), SwaggerRenderer()]),
    ),
]
```

into:

```python
schema = build_schema(router)

urlpatterns = [
    path(
        'docs/swagger',
        SwaggerView.as_view(schema=schema),
    ),
    path(
        'docs/openapi.json',
        OpenAPIJsonView.as_view(schema=schema),
    ),
]
```

- [ ] We should move all schema generation to the real objects. For example, `components.Headers` should generate header parameters spec
- [ ] It should be easy to add local type definitions that will be rendered in a certain way
- [ ] We should support `Literal`
- [ ] Our support for generics is not correct
- [ ] We should support `Annotated[list[...], '']`, right now it crashes
- [ ] We should support all fields from our serializers, including custom ones, see #446 
- [ ] We should support all extra fields from https://docs.pydantic.dev/2.0/usage/types/extra_types/phone_numbers/
- [ ] We should define clear separation between what we support and what serializer supports. We should support all native types, while serializer must support all its own fields and types

This issue will be updated once new tasks will be created.

PR #592, opened 2026-03-04
Making objects mutable and inheritable

kondratevdev:
# I have made things!

## Checklist

## Related issues

Refs #569


dmr/openapi/objects/components.py
```diff
@@ -17,8 +17,7 @@
     )
 
 
-@final
-@dataclass(frozen=True, kw_only=True, slots=True)
+@dataclass(kw_only=True)
```
sobolevn: `slots` must stay

sobolevn (commented):
Yes, please test that you can modify `build_schema()` result inplace. Also, replace all `Sequence` items to be `list[]` and all `Mapping[]`s to be `dict`s

schemathesis.toml
```diff
@@ -2,11 +2,6 @@
 # Limit the number of generated examples per API operation to keep tests fast.
 max-examples = 100
 
-[checks]
-# Docs: https://schemathesis.readthedocs.io/en/stable/reference/checks
-# We will expand the checks as we implement the openapi schemageneration.
-enabled = true
```
kondratevdev, line 8: All check enabled by default. We have no reason to duplicate it

dmr/openapi/objects/enums.py
```diff
@@ -1,8 +1,6 @@
 import enum
-from typing import final
 
 
-@final
 class OpenAPIFormat(enum.StrEnum):
```
sobolevn, line 6: Enums with values are final anyway, so let's keep the decorator for clarity.

sobolevn (approved):

PR #612, opened 2026-03-10
Add single-file step-by-step example for beginers

SmthFail:
Added a documentation for single-file example for beginners. Add expanded comments and explanation


sobolevn:
Thanks for the idea! We are working on it in #665

PR #618, opened 2026-03-11
Add schemathesis docs

sobolevn:
Friendly ping to @Stranger6667

Our work on OpenAPI spec is done. We generate a spec so detailed that it passes `schemathesis` by default. There are several problems with `xml`-test parser that we use, but it will also be fixed soon.

If you want to add something / change - please feel free :)
We are also accepting [testimonials](https://github.com/wemake-services/django-modern-rest?tab=readme-ov-file#testimonials), just saying 🌚️️

### DIFF
--- a/docs/pages/testing.rst
+++ b/docs/pages/testing.rst
@@ -77,3 +77,99 @@ Let's reuse the models for data generation in tests!
 
 Which will make your tests simple, fast,
 and will help you find unexpected corner cases.
+
+
+Property-based API testing
+--------------------------
+
+There's a great tool called
+`schemathesis <https://github.com/schemathesis/schemathesis>`_
+that can be used to test your API to match your OpenAPI spec.
+
+Official docs: https://schemathesis.readthedocs.io
+
+``schemathesis`` is not bundled together with the ``django-modern-rest``.
+You have to install it with:
+
+.. tabs::
+
+    .. tab:: :iconify:`material-icon-theme:uv` uv
+
+        .. code-block:: bash
+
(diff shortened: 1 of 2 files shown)


docs/pages/testing.rst
```diff
@@ -77,3 +77,99 @@ Let's reuse the models for data generation in tests!
...
+  :language: python
+  :linenos:
+
+What will happen here?
+
+1. ``schemathesis`` with load OpenAPI schema definition
+   from the ``reverse('openapi')`` URL
+2. Then we will create a top level ``schema`` object from the ``api_schema``
+   pytest fixture. It is needed to create a property-based test case
+3. Lastly, we create a generated test case with the help of
```
Stranger6667, line 129: Reads a bit incomplete - am I missing something? I.e. I was expecting "with the help of X"

docs/pages/testing.rst
```diff
@@ -77,3 +77,99 @@ Let's reuse the models for data generation in tests!
...
+
+    from dmr.test import DMRClient
+
+    def test_with_conditional_logic(dmr_client: DMRClient) -> None:
+        response = dmr_client.post(
+           '/users',
+           data={'name': 'Alice'},
+       )
+
+       assert schema['/users']['POST'].is_valid_response(response.json())
```
Stranger6667, line 175: Btw, I am planning to also add request validation (somebody asked about this in Discord). The use case would be to validate manually-constructed requests / requests coming from other frameworks

Stranger6667 (commented):
❤️

PR #660, opened 2026-03-15
feat: enhance JWT authentication to support custom headers

maksimzayats:
# I have made things!

This PR fixes JWT OpenAPI generation so the documented auth contract matches runtime behavior.

Previously, JWT auth always emitted a generic HTTP auth scheme in OpenAPI, even when JWTAsyncAuth / JWTSyncAuth were configured with a custom header or custom scheme. That meant generated docs and clients did not reflect the real request format.

With this change:

- Default `Authorization: Bearer <token>` auth is still documented as an HTTP bearer scheme
- Custom JWT configs like `X-Api-Auth: JWT <token>` are now documented as a header-based apiKey scheme with the real header name
## Checklist

### DIFF
--- a/dmr/security/jwt/auth.py
+++ b/dmr/security/jwt/auth.py
@@ -111,14 +111,22 @@ class _BaseJWTAuth:  # noqa: WPS214, WPS230
     @property
     def security_schemes(self) -> dict[str, SecurityScheme | Reference]:
         """Provides a security schema definition."""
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Great stuff, please, don't forget to add changelog entries for new features / bugfixes next time :)

There's also a similar problem with HttpBasic auth.

PR #664, opened 2026-03-15
add ping feature

### LINKED ISSUE #606: Support recurring ping events in SSE
https://github.com/wemake-services/django-modern-rest/issues/188

Example: https://github.com/sysid/sse-starlette/blob/main/sse_starlette/sse.py#L197

RenameMe1:
# I have made things!

## Checklist

## Related issues

Closes #606


dmr/sse/stream.py
```diff
@@ -161,8 +166,21 @@ async def _events_pipeline(self) -> AsyncIterator[bytes]:
         )
         async with context:
             try:
-                async for event in self._streaming_content:
+                while True:
+                    await self._send_lock.acquire()
+                    start = dt.datetime.now()
```
sobolevn: `time.perf_counter()`

dmr/sse/stream.py
```diff
@@ -161,8 +166,21 @@ async def _events_pipeline(self) -> AsyncIterator[bytes]:
         )
         async with context:
             try:
-                async for event in self._streaming_content:
+                while True:
+                    await self._send_lock.acquire()
```
sobolevn:
```suggestion
                    async with self._send_lock:
```

dmr/sse/stream.py
```diff
@@ -1,3 +1,5 @@
+import datetime as dt
+from asyncio import Lock
```
sobolevn:
```suggestion
import asyncio
```

dmr/sse/stream.py
```diff
@@ -99,6 +102,8 @@ def __init__(  # noqa: WPS211
             self._pipeline = self.validation_pipeline
         else:
             self._pipeline = ()
+        self.ping_interval = ping_interval
+        self._send_lock = Lock()
```
sobolevn:
```suggestion
        self._send_lock = asyncio.Lock()
```

dmr/sse/stream.py
```diff
@@ -161,8 +166,21 @@ async def _events_pipeline(self) -> AsyncIterator[bytes]:
...
             try:
-                async for event in self._streaming_content:
+                while True:
+                    await self._send_lock.acquire()
+                    start = dt.datetime.now()
+
+                    event = await anext(self._streaming_content)
                     event = self._apply_event_pipeline(event)
+
+                    send_time = dt.datetime.now() - start
```
sobolevn:
```suggestion
                    send_time = time.perf_counter() - start
```

dmr/sse/stream.py
```diff
@@ -161,8 +166,21 @@ async def _events_pipeline(self) -> AsyncIterator[bytes]:
...
+
+                    if self.ping_interval < send_time.total_seconds():
+                        self._send_lock.release()
+
+                    if not self._send_lock.locked:
```
sobolevn: Not clear why this condition exists.

dmr/sse/stream.py
```diff
@@ -161,8 +166,21 @@ async def _events_pipeline(self) -> AsyncIterator[bytes]:
...
+                        yield self._ping()
```
sobolevn: `_ping()` should produce `b": ping\n\n"` events

sobolevn (commented):
Thanks a lot! Things to do:

1. Basic docs
2. Docs with ADT schemas
3. Tests

dmr/sse/stream.py
```diff
@@ -159,17 +163,56 @@ async def _events_pipeline(self) -> AsyncIterator[bytes]:
...
+            if event_task is None:
+                event_task = asyncio.create_task(self._next_event())
+
+            done, _ = await asyncio.wait(
+                {event_task},
+                timeout=self.ping_interval,
+            )
+
+            if not done:
+                yield b'ping\r\n\r\n'
```
sobolevn:
```suggestion
                # We don't create intermediate `SSEvent` instances
                # for performance reasons here.
                yield b'ping\r\n\r\n'
```

sobolevn (commented):
Great! One thing: `ping` should be a default. Disabling `ping` should be optional.

Can you please fix that?
You can set the same default ping interval as fastapi has.

dmr/sse/stream.py
```diff
@@ -159,17 +163,58 @@ async def _events_pipeline(self) -> AsyncIterator[bytes]:
...
+
+    async def _ping_stream(self) -> AsyncIterator[bytes]:
+        event_task: asyncio.Task[SSE] | None = None
+
+        while True:
+            if event_task is None:
+                event_task = asyncio.create_task(self._next_event())
+
+            done, _ = await asyncio.wait(
+                {event_task},
```
sobolevn:
```suggestion
                [event_task],
```

lists are faster to create

dmr/sse/stream.py
```diff
@@ -155,17 +158,58 @@ async def _events_pipeline(self) -> AsyncIterator[bytes]:
...
+    async def _default_stream(self) -> AsyncIterator[bytes]:
+
+        async for event in self._streaming_content:
+            event = self._apply_event_pipeline(event)
+            yield self.serializer.serialize(
+                event,
+                renderer=self.sse_renderer,
+            )
+
+    async def _ping_stream(self) -> AsyncIterator[bytes]:
```
milssky: Hi! 

It seems that there may be hanging tasks in such code if the client disconnects unexpectedly (the connection is broken for some reason), here is a minimal demonstration. it seems like this might be a problem in the production code.

Reproduce code:

```python
import asyncio

class SSEStreamer:
    def __init__(self):
        self.ping_interval = 0.5

    async def _next_event(self):
        try:
            await asyncio.sleep(3)
            return "Real Event"
        except asyncio.CancelledError:
            raise

    async def _ping_stream(self):
        event_task = None
        while True:
            if event_task is None:
                event_task = asyncio.create_task(self._next_event())

            done, pending = await asyncio.wait(
                [event_task],
                timeout=self.ping_interval,
            )

            if not done:
                yield b'ping\r\n\r\n'
                continue

            event = event_task.result()
            event_task = None
            yield f"event: {event}\n\n".encode()


async def simulate_client_disconnect():
    streamer = SSEStreamer()
    stream = streamer._ping_stream()

    async for chunk in stream:
        print(chunk)
        break

    await stream.aclose()

    leaked = [t for t in asyncio.all_tasks() if t is not asyncio.current_task()]
    print(leaked)
    assert leaked

await simulate_client_disconnect()

# b'ping\r\n\r\n'
# [<Task pending name='Task-195' coro=<SSEStreamer._next_event() running at <ipython-input-9-86e872d063b7>:9> wait_for=<Future pending cb=[Task.task_wakeup()]>>]

PR #668, opened 2026-03-15
docs: properly quote shell args in example output

sesav:
I looked over the project and its endpoints and noticed that a lot of the curl commands in the docs are missing quotes. 

If you've got spaces, ampersands, or these brackets { } < >, you've gotta put quotes around it for the shell. 

Just a quick fix for all the curl requests in the docs.

Before the fix:
```
$ curl http://127.0.0.1:8000/api/math/ -d {"left": 1, "right": 0} -H Content-Type: application/json
```

After the fix:
```
$ curl http://127.0.0.1:8000/api/math/ -d '{"left": 1, "right": 0}' -H 'Content-Type: application/json
```


sobolevn (approved):
this is really cool! thank you!

PR #671, opened 2026-03-15
fix: add support for additional JWT verification options

maksimzayats:
# I have made things!

Fixed `JWToken.decode` so `verify_iat`, `verify_jti`, and `verify_sub` are no longer ignored and are forwarded into PyJWT’s decode options in `token.py`. Added a regression test in `test_jwt_token.py` to verify those flags are passed through correctly.

## Checklist


tests/test_unit/test_security/test_jwt/test_jwt_token.py
```diff
@@ -258,3 +258,59 @@ def test_verify_nbf_can_be_disabled_independently() -> None:
...
+        algorithm='HS256',
+        verify_iat=False,
+        verify_jti=False,
+        verify_sub=False,
+    )
+
+    assert token.sub == 'foo'
+    assert captured_options['verify_iat'] is False
+    assert captured_options['verify_jti'] is False
+    assert captured_options['verify_sub'] is False
```
sobolevn:
```suggestion
```

sobolevn (approved):
Thanks!

PR #674, opened 2026-03-15
fix: update session auth OpenAPI schema for session-backed CSRF

maksimzayats:
# I have made things!

Stop advertising CSRF as a cookie-based OpenAPI security scheme when Django is configured with CSRF_USE_SESSIONS=True.

This updates DjangoSessionSyncAuth and DjangoSessionAsyncAuth to make the generated security schema environment-aware, while keeping the existing behavior for the default cookie-backed CSRF mode. Tests were added to cover both configurations.

## Checklist


dmr/security/django_session/auth.py
```diff
@@ -106,6 +110,9 @@ def provide_response_specs(
             ),
         ]
 
+    def _uses_csrf_cookie(self) -> bool:
+        return not settings.CSRF_USE_SESSIONS
```
sobolevn: this is simple enough not to use a custom method

dmr/security/django_session/auth.py
```diff
@@ -57,26 +57,30 @@ def __init__(
...
             self.security_scheme_name: SecurityScheme(
                 type='apiKey',
                 name=settings.SESSION_COOKIE_NAME,
                 security_scheme_in='cookie',
                 description='Reusing standard Django auth flow for API',
             ),
-            # TODO: this is not right if `CSRF_USE_SESSIONS` is used:
-            self.csrf_scheme_name: SecurityScheme(
+        }
+        if self._uses_csrf_cookie():
```
sobolevn, line 68: Should we somehow specify that the session is expected? `apiKey` maybe?

tests/test_unit/test_security/test_django_session_auth.py
```diff
@@ -167,3 +169,25 @@ def test_schema(
         'django_session': [],
         'csrf': [],
     })
+
+
+@pytest.mark.parametrize('typ', [DjangoSessionSyncAuth, DjangoSessionAsyncAuth])
+def test_schema_with_csrf_sessions(
+    settings: LazySettings,
+    typ: type[DjangoSessionSyncAuth] | type[DjangoSessionAsyncAuth],
```
sobolevn, line 179:
```suggestion
    settings: LazySettings,
    *,
    typ: type[DjangoSessionSyncAuth] | type[DjangoSessionAsyncAuth],
```

nitpick for our parameter style

dmr/security/django_session/auth.py
maksimzayats: Hi @sobolevn ! I'm not sure tbh...

sobolevn: When `CSRF_USE_SESSIONS` is set, django loads `request.session` from this cookie: https://github.com/django/django/blob/3abf89887993140d28676f26420ee0d46a617f51/django/contrib/sessions/middleware.py#L19

Then it loads the CSRF from the session here: https://github.com/django/django/blob/3abf89887993140d28676f26420ee0d46a617f51/django/middleware/csrf.py#L229-L237

So, when `CSRF_USE_SESSIONS` is set, we must require the session cookie.

PR #675, opened 2026-03-15
fix: improve JWT claim validation and error handling

maksimzayats:
# I have made things!

Ensure JWToken.decode() returns NotAuthenticatedError instead of leaking KeyError or TypeError when required claims are missing or malformed and claim verification is disabled.

Changes:
- validate required sub, exp, and iat claims explicitly during decode
- convert malformed datetime claims into auth failures
- add regression tests for missing exp, iat, sub, and invalid datetime claim values

## Checklist


sobolevn (approved):
Btw, I took this piece of code from `litestar`, you can contribute a similar fix there.

PR #680, opened 2026-03-16
test: add unit tests for format_error

sesav:
Add several tests targeting `format_error` in `dmr/errors.py`.

- Handles string input, optionally including `loc` and `error_type`, and manages all exception types: `ValidationError`, `RequestSerializationError`, `ResponseSchemaError`, `NotAcceptableError`, `NotAuthenticatedError`, `InternalServerError`.
   and the fallback mechanism for `NotImplementedError`.

- Verify `InternalServerError` works correctly with `DEBUG` as `True` and `False`.

### DIFF
--- a/tests/test_unit/test_errors/test_format_error.py
+++ b/tests/test_unit/test_errors/test_format_error.py
@@ -1,0 +1,146 @@
+from http import HTTPStatus
+
+import pytest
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thank you!

Issue #678, opened 2026-03-16
Formalize our CDN / static files support for OpenAPI renderers


sobolevn:
I think that we should be able to configure: do we want to use local static files, served by django for OpenAPI renderers and interactive APIs, or should we use CDN for this.

Currently:
- Swagger uses local static files
- Scalar / Redoc use CDNs

Plan:
1. Introduce `openapi_static_cdn: {'scalar': 'https://cdn_url.com'}` option, where we can specify when these libs should use CDN
2. Only when some lib has this explicit setting, we should use the CDN. Otherwise, they should use local static files which should be copied to `dmr`
3. We need to document the EXACT versions we have bundleded in our static files
4. We need to copy their licenses as well, not just static files. Because we take their code / build result. They can go to our `LICENSE.md` at the very bottom, specify why it is there

PR #702, opened 2026-03-20
Formalize our CDN / static files support for OpenAPI renderers

F0RRZZ:
- Issue: https://github.com/wemake-services/django-modern-rest/issues/678


dmr/openapi/views/scalar.py
```diff
@@ -25,12 +26,17 @@ class ScalarView(OpenAPIView):
...
+        scalar_cdn = cdn_config.get('scalar')
+
         return render(
             request,
             self.template_name,
             context={
                 'title': self.schema.info.title,
                 'schema': self.dumps(self.schema.convert()),
+                'use_cdn': bool(scalar_cdn),
+                'scalar_cdn': scalar_cdn,
```
sobolevn: we don't need `use_*` variable, `*_cdn` is enough :)

licenses/swagger-ui.txt
sobolevn, line 1: Let's put it to `licenses/*` and split this file into several ones.

sobolevn (commented):
Thanks a lot! Great stuff.

Please:

- [ ] also pin the versions that you have downloaded
- [ ] and add a `autouse=True` fixture with `param=use_cdn=[True, False]`
- [ ] add a doc note about this into `openapi.rst`

Makefile
```diff
@@ -32,7 +32,7 @@ type-check: ## Run all type checkers we support
 
 .PHONY: spell-check
 spell-check: ## Run spell checking
-	$(POETRY) run codespell dmr tests docs typesafety README.md CONTRIBUTING.md CHANGELOG.md
+	$(POETRY) run codespell dmr tests docs typesafety README.md CONTRIBUTING.md CHANGELOG.md --skip="dmr/static/*,*.js,*.css,*.min.*"
```
kondratevdev: use `[tool.codespell]` `pyproject.toml` section instead

dmr/openapi/views/scalar.py
kondratevdev: Also, after that, `scalar_cdn` will be used only once


```suggestion
                'scalar_cdn': cdn_config.get('swagger'),
```

dmr/openapi/views/swagger.py
```diff
@@ -25,12 +26,17 @@ class SwaggerView(OpenAPIView):
...
+        swagger_cdn = cdn_config.get('swagger')
+
         return render(
             request,
             self.template_name,
             context={
                 'title': self.schema.info.title,
                 'schema': self.dumps(self.schema.convert()),
+                'use_cdn': bool(swagger_cdn),
+                'swagger_cdn': swagger_cdn,
```
kondratevdev: Same as previous

sobolevn (commented):
and also, don't forget about `configuration.rst` docs :)

sobolevn (approved):
Great stuff! Thanks!

PR #707, opened 2026-03-21
feat: wrap user-facing strings in gettext_lazy for i18n support

### LINKED ISSUE #426: Support django's `i18n` for translation
All our user-facing texts must use django's i18n.
We will only provide en-us translation for now. Later more can be added if needed.

cmdtorch:
# I have made things!

## Checklist

## Related issues

- Closes #426

## Notes

Wraps all user-facing strings in Django's `gettext_lazy` for i18n support,
following the same scope as DRF. Only `en_US` translation is provided for now.

For `InternalServerError.default_message` and `NotAuthenticatedError.default_message`
I used `ClassVar[Promise]` as the type annotation since `gettext_lazy` returns
a lazy proxy, not a plain `str`. Happy to change this if there's a preferred approach.

Also, I may have missed some user-facing strings, if you spot any,
I'll wrap them too.

Overall, open to any feedback or suggestions :)


dmr/internal/django.py
```diff
@@ -41,9 +41,16 @@
 from django.http.multipartparser import MultiPartParser, MultiPartParserError
 from django.http.request import HttpRequest, QueryDict
 from django.utils.datastructures import CaseInsensitiveMapping, MultiValueDict
+from django.utils.translation import gettext_lazy as _
 
 from dmr.exceptions import RequestSerializationError
 
+_utf8_required_msg = _(
```
sobolevn:
```suggestion
_UTF8_REQUIRED_MSG: Final = _(
```

sobolevn (commented):
Please, rename all constants and use `Final` :)
And also update `CHANGELOG.md`

tests/test_unit/test_i18n/test_strings.py
kondratevdev, line 1: Please, use `@pytest.mark.parametrize`

tests/test_unit/test_i18n/test_strings.py
```diff
@@ -1,49 +1,58 @@
...
+        (NotAuthenticatedError, 'Not authenticated'),
+    ],
+)
+def test_default_message_fallback_to_english(
+    *,
+    exception_cls: type,
+    expected: str,
+) -> None:
+    """Ensure default_message falls back to English."""
     with override('xx'):
```
kondratevdev: What `'xx'` mean?
If the value doesn't make sense, then I suggest using `faker.text()`.

cmdtorch: 'xx' is a non-existent locale (like en, fr and etc.) to guarantee no translation exists and trigger the English fallback. faker.text() generates a random string. 'xx' is deterministic and clearly communicates that we need a non-existent locale.

cmdtorch: Maybe I should add a comment?
```python
with override('xx'):  # 'xx' is a non-existent language code
```

sobolevn: 1. `wrong_lang_code = 'xx'`
2. `with override(wrong_lang_code):` 

:)

sobolevn (approved):
Thank you!

Issue #701, opened 2026-03-20
limit_to_content_types on global ResponseSpec raises ResponseSchemaError (422) when error handler returns different content type than negotiated renderer.


sh1rokovs:
**Summary**

When `limit_to_content_types` is set on a `ResponseSpec` registered in `DMR_SETTINGS[Settings.responses]` (global error responses), and the global error handler returns an `HttpResponse` whose `Content-Type` differs from the request's negotiated renderer content type, dmr raises a `ResponseSchemaError` and returns a 422 instead of the intended error response.

When `limit_to_content_types` is set application/problem+json
<img width="1675" height="731" alt="Image" src="https://github.com/user-attachments/assets/e275ea47-1146-47fc-93be-531b2c6f6155" />
<img width="1689" height="738" alt="Image" src="https://github.com/user-attachments/assets/b8132804-6ddb-4400-861f-44a4ca4f7003" />

When `limit_to_content_types` is set None
<img width="1680" height="686" alt="Image" src="https://github.com/user-attachments/assets/f7db9759-7551-4c56-b07b-25a2d23a03fb" />
<img width="1657" height="721" alt="Image" src="https://github.com/user-attachments/assets/f9722933-b0b3-4cad-954f-de4bf96576ba" />

**Environment**

- dmr version: 0.2.0
- Python: 3.12
- Django: 5.2

**Steps to reproduce**

```python
# settings.py
  DMR_SETTINGS = {
      Settings.responses: [
          ResponseSpec(
              ProblemDetail,
              status_code=HTTPStatus.UNAUTHORIZED,
              description="Unauthorized",
              limit_to_content_types=frozenset({"application/problem+json"}),
          ),
      ],
      Settings.global_error_handler: "myapp.error_handler.dmr_global_error_handler",
  }
```

```python
# error_handler.py
  def dmr_global_error_handler(endpoint, controller, exc):
      if isinstance(exc, NotAuthenticatedError):
          return JsonResponse(
              {"type": "about:blank", "title": "Unauthorized",
  "status": 401},
              status=401,
              content_type="application/problem+json",
          )
      raise
```

1. Send a request without authentication to any protected endpoint.
2. Client sends `Accept: application/json` (or `Accept: */*` header (ex.: Postman)).

**Expected behavior**

The global error handler returns a 401 response with `Content-Type: application/problem+json`. Since the `ResponseSpec` for 401 allows `application/problem+json`, validation should pass and the client should receive the 401

**Actual behavior**

[...]

PR #711, opened 2026-03-21
fix: Fixed response validation when global error handler returns `HttpResponse` with a different content type than the negotiated renderer

sh1rokovs:
# I have made things!

## Checklist

## Related issues

- Refs: [701](https://github.com/wemake-services/django-modern-rest/issues/701)


sobolevn (commented):
we need a test for this :)

Issue #718, opened 2026-03-22
Translate `django-modern-rest` to your native language!


sobolevn:
Docs: https://docs.djangoproject.com/en/6.0/topics/i18n/

How to do it?

1. Run `poetry install --all-groups --all-extras`
2. Run `poetry run django-admin makemessages -l YOUR_LOCALE`, for example: `poetry run django-admin makemessages -l ru_RU`, full list of locales: https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes
3. Translate the new file created in `dmr/locale/YOUR_LOCALE/LC_MESSAGES/django.po`
4. Run `poetry run django-admin compilemessages`
5. Done! You are awesome, submit the PR :)

We welcome all contributions!

vgnatyuk-ai:
Hello, I can try  translate into russian

Satevg:
I think I noticed some inconsistencies that we have in master

  - hi_HI → hi_IN 
  - ne_NE → ne_NP
  - tt → tt_RU

🤷

gas12241:
Hello, my name is George and I'm looking to help (part of CodePath's AI301 course where we take on Open Source issues - this will be my first!)

Could I add a translation for Mexican Spanish if possible? I checked the codebase and didn't see it in there (I looked for either SP or ES as language codes).

Axiya3749:
Hi! I'd like to work on this. I can translate django-modern-rest into both Simplified Chinese (zh-CN) and Traditional Chinese (zh-TW). I see Russian is already in progress, so I'll focus on these two locales.
Quick question: would you prefer these as a single PR covering both locales, or two separate PRs (one per locale)? Happy to follow whichever convention you'd like. Thanks!

sobolevn:
Two PRs would be better :)

Axiya3749:
Quick note for anyone else picking this up: the poetry install --all-groups --all-extras command in the setup instructions looks outdated, the project's CONTRIBUTING.md now uses uv sync --all-groups --all-extras instead. Just flagging in case it's worth updating!

sobolevn:
updated :)

iraleGurudatta:
Hi! I'd like to work on adding a Marathi (mr) translation for the user-facing error messages, if that's still open. Will follow the steps in the internationalization docs and open a PR referencing this issue. 🙂

sobolevn:
Please, go ahead :)

soil0119:
Can I try translate into korea?

PR #720, opened 2026-03-22
feat: add Kazakh (kk_KZ) translation

deliro:
## Summary
- Add Kazakh (`kk_KZ`) translation for all 12 user-facing strings
- Includes both source `.po` and compiled `.mo` files

Refs #718

## Test plan
- [ ] Verify `.po` file contains all translatable strings
- [ ] Verify `.mo` file compiles without errors
- [ ] Verify translations render correctly with `LANGUAGE_CODE = 'kk-kz'`


sobolevn (approved):
Great stuff! I don't know the language to validate, but I will trust the llm here. I am pretty sure it can be fixed later on.

PR #726, opened 2026-03-22
feat: add Brazilian Portuguese translation

thepabloaguilar:
# I have made things!

I've added the Brazilian Portuguese translation: __pt_BR__

## Checklist

## Related issues

Refs #718


pyproject.toml
```diff
@@ -347,4 +347,4 @@ project-includes = ["dmr", "typesafety"]
 
 [tool.codespell]
 # codespell configuration: https://pypi.org/project/codespell
-skip = '__pycache__,_build,.mypy_cache,swagger-*.js,dmr/static/*,*.js,*.css,*.min.*,*.msgpack'
+skip = '__pycache__,_build,.mypy_cache,swagger-*.js,dmr/static/*,*.js,*.css,*.min.*,*.msgpack,dmr/locale/*.po'
```
thepabloaguilar: I had to add this because `codespell` was checking Portuguese words in the `*.po` file, another possible solution would be add skip to each specific `locale`!

`codespell` uses `fnmatch` for this so we can use a regex like `dmr/locale/(?!en)_.*`

sobolevn (approved):
Great stuff! Thanks a lot!

Vamos Brasil! 🇧🇷

Issue #709, opened 2026-03-21
Idea: Tracecov support in test clients


Stranger6667:
I see that tracecov is being integrated into the test suite in #650 which is super cool to see! 

Do you think there is a place for a tracecov integration via [test helpers](https://github.com/wemake-services/django-modern-rest/blob/master/dmr/test.py) too? I.e a built-in test client that automatically reports api coverage after tests - for situations where users write their tests manually it could provide useful indication on how much of the schema they cover and what are the gaps (gaps info is available in raw report data).

There are already some [Django-specific helpers available](https://docs.tracecov.sh/integration/python/#django-django-rest-framework).

Also, if there is anything missing in tracecov, please let me know :)

FYI: since `0.17` tracecov supports markdown reports, which I added to [schemathesis's GH action](https://github.com/schemathesis/action?tab=readme-ov-file#coverage-report). Not sure if it is immediately useful here, but just in case.

kondratevdev:
Yes, it would be great if tracecov had its own pytest plugin. I agree that it should support the same flags as the CLI when integrated with Schemathesis.

`map.check_thresholds(min_operations=0.8, ...)` also sounds like a good idea. It would be useful if the `map` exposed a low-level API for manual integration with pytest, in case custom behavior is needed.

It would also be valuable to have a unified API (for example, via a fixture) that allows registering interactions from Schemathesis, the Django test client, and RequestFactory within pytest.

Stranger6667:
Great! I'll work on it soon - will let you know about the progress.

> It would also be valuable to have a unified API (for example, via a fixture) that allows registering interactions from Schemathesis, the Django test client, and RequestFactory within pytest.

How would you see such an API? I.e. right now `CoverageMap` has namespaced `record` methods for each situation - would you like to have a method that will automatically detect the right type and record it accordingly?

Stranger6667:
I.e. something that will expose a map registered here?

```python
@pytest.fixture(scope='session')
def register_tracecov_map(pytestconfig: pytest.Config) -> RegisterTracecovMap:
    """Append tracecov summary to terminal when a map was registered."""
    def factory(map: tracecov.CoverageMap) -> None:
        pytestconfig.stash[_TRACECOV_MAP_KEY] = map

    return factory

@pytest.fixture(scope='session')
def tracecov_map(pytestconfig):
    # with some wrapper that prevent KeyError with some user-friendly message
    return pytestconfig.stash[_TRACECOV_MAP_KEY]

[...]

PR #735, opened 2026-03-24
Tracecov integration

kondratevdev:
# I have made things!

## AI Policy

- [ ] I have disclosed AI usage in this PR,
      if any AI / LLM / coding agent was participating in the making of this PR
- [ ] I have removed any "Co-Authored-By" lines that attribute coding agents
- [ ] I have carefully reviewed the final result, produced by the AI, manually

## Checklist

## Related issues

Refs #709


docs/pages/testing.rst
```diff
@@ -189,3 +189,35 @@ Example:
        )
 
        assert schema['/users']['POST'].is_valid_response(response.json())
+
+
+API coverage with TraceCov
+--------------------------
+
+`TraceCov <https://docs.tracecov.sh/>`_ can be used as an optional API
```
sobolevn, line 197: Add a note that it should be installed independently.

docs/pages/testing.rst
```diff
@@ -189,3 +189,35 @@ Example:
...
+when ``tracecov_map`` is missing or inactive, this integration is skipped and
+fixtures return regular DMR clients without tracking.
+
+This means ``schemathesis`` and ``tracecov`` work well together: ``schemathesis``
+checks schema conformance, while ``tracecov`` reports endpoint coverage depth for
+the same test run.
+
+Example ``conftest.py``:
+
+.. code-block:: python
```
sobolevn: Can we include our real fixture here?

docs/pages/testing.rst
```diff
@@ -189,3 +189,35 @@ Example:
...
+coverage layer for ``django-modern-rest`` test suites. It complements regular
+integration tests and ``schemathesis`` runs by showing which OpenAPI operations
+and parameters were actually exercised.
+
+When ``tracecov_map`` is configured, ``dmr_client`` and ``dmr_async_client``
+automatically register requests in TraceCov. When TraceCov is not installed, or
+when ``tracecov_map`` is missing or inactive, this integration is skipped and
+fixtures return regular DMR clients without tracking.
+
+This means ``schemathesis`` and ``tracecov`` work well together: ``schemathesis``
```
sobolevn: Please include an image or even the report itself of how `tracecov` looks. Without it - it is not quite clear.

dmr_pytest.py
```diff
@@ -19,20 +19,49 @@
     )
 
 
+def _maybe_track_client(
```
sobolevn: Can we move this to `dmr.test`? This way `dmr_pytest` will be minimal.

sobolevn (commented):
I would also add a bit about "why" people should use this. How is this different from a regular coverage? Why is it important to find bugs? How is it related with our semantic schema? Why only `dmr` can do this and other frameworks 🌚️️️️ can't?

dmr_pytest.py
kondratevdev: What about `pytest.FixtureRequest`? In this fucntion we explictly use it

sobolevn: it is ok to just type it as `Any`. It is our internal function after all.
Or type it with a protocol

`dmr_pytest` now is included in the docs as-is. Any non-fixture code will confuse readers.

sobolevn (approved):
Great stuff!

PR #743, opened 2026-03-25
fix: Fixed header type to compatibility with wsgiref

RenameMe1:
## Checklist


sobolevn (approved):
Thanks!

PR #750, opened 2026-03-26
feat: Improve doc examples with 'assert-error-text'

### LINKED ISSUE #742: Improve doc examples with `"fail-with-body": false`
There are many cases where we show different failures, example https://github.com/wemake-services/django-modern-rest/blob/24c2545d852521bfcbeb29093c0d973782260e53/docs/examples/routing/handler500.py#L36

But, it is not really safe right now. Why?
Because any error can happen, we won't know that some other error happened.

So, let's add `"assert-error-text": "text to be found"`, it will assert that `config['assert-error-text']` is present in the response.

We will add this flag to all `run` commands with `"fail-with-body": false` to be sure that the text we expect really is present.

Mr-Sunglasses:
# I have made things!
Improve doc examples with 'assert-error-text'

## AI Policy
I've used claude opus 4.6 to summarize different error's in the examples and ask it to fix a lint error

- [X] I have disclosed AI usage in this PR,
      if any AI / LLM / coding agent was participating in the making of this PR
- [X] I have removed any "Co-Authored-By" lines that attribute coding agents
- [X] I have carefully reviewed the final result, produced by the AI, manually

## Checklist

## Related issues

Closes: #742


sobolevn (commented):
niiiiice, please fix the CI :)

sobolevn (approved):
Let's go! 🎉

PR #791, opened 2026-04-02
Add `set_cookies` function

### LINKED ISSUE #773: Provide `dmr.cookies.set_cookies` helper function
Currently we have two places where we duplicate this code: https://github.com/wemake-services/django-modern-rest/blob/a6bc356da698b5959cef80e93e15e234eb8599c1/dmr/streaming/controller.py#L172-L178

and 

https://github.com/wemake-services/django-modern-rest/blob/a6bc356da698b5959cef80e93e15e234eb8599c1/dmr/response.py#L256-L259

So, it can probably be useful to others as well.
Let's move this to `cookies.py` as a public helper function.

Also, document it in `public-api.rst`

This is a very easy task for new contributors.

neocim:
Closes #773

I added the public helper function `set_cookies` to `dmr.cookies` and documented it in `public-api.rst`.

### DIFF
--- a/docs/pages/deep-dive/public-api.rst
+++ b/docs/pages/deep-dive/public-api.rst
@@ -54,6 +54,9 @@ Response, headers and cookies
 .. autoclass:: dmr.cookies.NewCookie
   :members:
 
+.. autofunction:: dmr.cookies.set_cookies
+  :members:
+
 
 Validation
 ----------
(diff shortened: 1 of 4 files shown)


sobolevn (approved):
Great stuff, thanks!

docs/pages/deep-dive/public-api.rst
```diff
@@ -54,6 +54,9 @@ Response, headers and cookies
 .. autoclass:: dmr.cookies.NewCookie
   :members:
 
+.. autofunction:: dmr.cookies.set_cookies
+  :members:
```
sobolevn:
```suggestion
```

PR #797, opened 2026-04-03
Migrate from poetry to uv

Satevg:
# I have made things!

  Migrate from `Poetry` to `uv`

  Replaces `Poetry` with `uv` for dependency management and packaging.

  - `pyproject.toml` migrated to PEP 621 + hatchling build backend
  - `poetry.lock` replaced with `uv.lock`
  - CI updated to `astral-sh/setup-uv@v8.0.0`
  - `ReadTheDocs`, `Makefile`, and `CONTRIBUTING` updated accordingly

## AI Policy

## Checklist


.github/workflows/test.yml
```diff
@@ -35,22 +35,18 @@ jobs:
...
-    - name: Install poetry
-      run: |
-        curl -sSL "https://install.python-poetry.org" | python
-
-        # Adding `poetry` to `$PATH`:
-        echo "$HOME/.poetry/bin" >> "$GITHUB_PATH"
+    - name: Install uv
+      uses: astral-sh/setup-uv@v8.0.0
+      with:
+        version: "0.11.3"
```
sobolevn: We don't require a specific version here

pyproject.toml
```diff
@@ -1,24 +1,22 @@
 [build-system]
-requires = ["poetry-core>=2.2"]
-build-backend = "poetry.core.masonry.api"
+requires = ["hatchling"]
```
sobolevn, line 2: Why not `uv_build`?

sobolevn (commented):
Please, rebase your branch :)

pyproject.toml
Satevg: It seems that `uv_build` works with package directories only. This is not going to work, since we need `dmr_pytest.py` as well.

https://github.com/astral-sh/uv/issues/16766

in a local `.whl` built with `uv_build` I saw just:
```
    dmr/                                                                                      
    django_modern_rest-0.4.0.dist-info/  
```

sobolevn (approved):
Thanks a lot!

Satevg:
> One more idea: we use `make` to build things, but we can try to modernize the setup with `just`. If you are up to it, here's the template we can use [typeddjango/django-stubs#3240](https://github.com/typeddjango/django-stubs/pull/3240)

tbh, this is the first time I hear about `just`, interesting. I'll have a look :rocket:

pyproject.toml
DhavalGojiya: > since we need dmr_pytest.py as well.

In that case I think you need to consider restructure the location of this file.
You can move this file `dmr_pytest.py` inside `dmr/` directory if this file is part of your package. uv build backend will completely work fine in that case.

> in a local .whl built with uv_build I saw just:
    dmr/                                                                                      
    django_modern_rest-0.4.0.dist-info/  

This is expected behavior, The `.whl` is directly downloaded by the end user via `pip`. This zip file then extracted in projects `.venv/**/site-packages/` directory. 
So you need to exclude everything that is not useful for end user in the wheel distribution. 
You can include this all test files in source distributions but not in wheel distribution.

PR #834, opened 2026-04-06
build: remove hatch from build dependencies

subnix:
`hatch` itself is not required for building. Having only `hatchling` should be enough.

## AI Policy

## Checklist


sobolevn (approved):
Thank you! Great find!

PR #847, opened 2026-04-08
Support CSP for OpenAPI UI

kondratevdev:
# I have made things!

## AI Policy

## Checklist


pyproject.toml
```diff
@@ -81,6 +81,7 @@ dmr = "dmr_pytest"
 # It is installed in `cibuildwheel` and must be minimal.
 unit-test = [
   "django>=5.2,<6",
+  "django-csp>=4.0",
```
```suggestion
  "django-csp>=4.0,<5",
```

This should go into `dev`, here we only add things that are needed for `test_unit` tests.

docs/pages/openapi/openapi.rst
```diff
@@ -72,6 +72,12 @@ What happens in the example above?
 
   You can also modify the exact versions that we use for each tool this way.
 
+  The UI shipped by ``django-modern-rest`` are prepared for
```
sobolevn: Can you please be more specific here? Are there views that do better / worth? What the known problems?

I would also suggest to provide CSP configuration example.

django_test_app/server/settings.py
```diff
@@ -142,3 +149,18 @@
 # https://docs.djangoproject.com/en/5.2/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
+
+# Content Security Policy:
+# https://django-csp.readthedocs.io/en/latest/configuration.html
+
+CONTENT_SECURITY_POLICY = {
+    'EXCLUDE_URL_PREFIXES': ['/docs'],
```
sobolevn: Do we have to ignore all `/docs/` views? Or only some of them like scalar, redoc?

If so, let's be specific.

We totally don't need to ignore `/docs/openapi.json` and `/docs/openapi.yaml`

pyproject.toml
kondratevdev: Pipeline failed without it: https://github.com/wemake-services/django-modern-rest/pull/847/checks?sha=3a5ed8ae6808f9ae1f75c97d8356ab5bf5e6f83e

sobolevn: Oh, I see. Because it is listed in the `INSTALLED_APPS` 👍

docs/pages/integrations.rst
```diff
@@ -163,6 +163,45 @@ is required.
...
+Known caveats:
+
+- Some upstream OpenAPI bundles inject styles at runtime, so a very strict
+  policy can still break the page.
+- When CSP is a hard requirement, start with local bundled assets and test
+  the exact renderer and version you plan to deploy.
+
+Example ``django-csp`` setup:
+
+.. literalinclude:: /../django_test_app/server/settings.py
```
sobolevn, line 194: Let's make a an exact link to some of our existing commits on GitHub with exact lines.

Or maybe even better to `wemake-django-tempalte`.
Because in this setup - we will screw up the formatting.

kondratevdev: There are none at the moment. I'll fix it on the wemake-django-template as I make changes there. I made a note to myself.

sobolevn (approved):

PR #849, opened 2026-04-08
feat: validate async endpoints

### LINKED ISSUE #843: Validate that http endpoints can't be async generators
```python
import dataclasses
from collections.abc import AsyncIterator

from dmr.plugins.msgspec import MsgspecSerializer
from dmr.streaming.sse import SSEController, SSEvent


@dataclasses.dataclass(frozen=True, slots=True)
class _User:
    email: str


class UserEventsController(SSEController[MsgspecSerializer]):
    async def get(self) -> AsyncIterator[SSEvent[_User]]:
        # all SSEvent fields can be customized:
        yield SSEvent(
            _User(email='first@example.com'),
            event='user',
        )
        yield SSEvent(
            _User(email='second@example.com'),
            event='user',
        )
```

This code must raise an error during controller creation. We need to run a validation for that.
We can do that in a regular https://github.com/wemake-services/django-modern-rest/blob/master/dmr/validation/controller.py

Because this validation rule should be also viable without explicit `StreamingController` subclassing.

milssky:
# I have made things!

## AI Policy

## Checklist

## Related issues

Closes #843


dmr/validation/controller.py
```diff
@@ -19,12 +20,30 @@ def __call__(
...
+    def _validate_async_generator_endpoints(
+        self,
+        controller: type['Controller[BaseSerializer]'],
+    ) -> None:
+        for method_name in controller.allowed_http_methods:
+            method = getattr(controller, method_name, None)
+            if isinstance(
+                method,
+                types.FunctionType,
+            ) and inspect.isasyncgenfunction(method):
```
sobolevn: While we are at it: let's all check for sync generators. Because they also won't work.

tests/test_unit/test_controllers/test_controller_validation.py
```diff
@@ -174,3 +176,25 @@ def handle_error(
...
             raise NotImplementedError
+
+
+def test_endpoint_rejects_async_gen() -> None:
+    """Ensure endpoints cannot be async generators."""
+
+    class _NoOpOptimizer(PydanticEndpointOptimizer):
+        @override
+        @classmethod
+        def optimize_endpoint(cls, metadata: Any) -> None:  # noqa: WPS324
```
sobolevn, line 189: Sorry, why do you need this type?

milssky: If i use controller like this 

```python
    with pytest.raises(
        EndpointMetadataError,
        match='is an async generator',
    ):

        class _BadController(Controller[PydanticSerializer]):
            async def get(self) -> AsyncIterator[int]:
                yield 1  # pragma: no cover
```

I have pydantic error =(

```
.venv/lib/python3.13/site-packages/pydantic/_internal/_generate_schema.py:659: in _unknown_type_schema
    raise PydanticSchemaGenerationError(
E   pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for collections.abc.AsyncIterator[int]. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.
E
E   If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.
E
E   For further information visit https://errors.pydantic.dev/2.12/u/schema-for-unknown-type
```

sobolevn: Got it!

Please, then also test that `SSEController` an `JsonLinesController` do not need this.

sobolevn (approved):

Issue #851, opened 2026-04-08
Rewrite `controller.request.accepts` in compiled C version


sobolevn:
@albertedwardson I found a case where we can benefit from C compilation :)
Currently there are several places where we use default Django's `controller.request.accepts` or `self.request.accepts`.

However, it REALLY slow. We can provide our own `accepts` function in `compiled.py`.
It can be at least several times faster.

Do you want to contribute? :)

albertedwardson:
yeah, sure!

PR #854, opened 2026-04-08
provide faster alternative `dmr.compiled.accepted_header` in favour of django.http `HttpRequest.accepts`

albertedwardson:
Refs: #851


dmr/_compiled/negotiation.py
```diff
@@ -65,6 +65,23 @@ def accepted_type(
...
+
+
+def header_value_matches_media_type(header_value: str, media_type: str) -> bool:
+    return accepted_type(header_value, (media_type,)) is not None
+
+
+def request_accepts(request: _HasHeadersAsMapping, media_type: str) -> bool:
+    """Does the client accept a response in the given media type?"""
+    return header_value_matches_media_type(
+        request.headers.get('Accept', '*/*'),
```
sobolevn: you can pass just `headers`, not `request`. This way we can remove the protocol and my wild guess is that it will be faster.

dmr/_compiled/negotiation.py
```diff
@@ -65,6 +65,23 @@ def accepted_type(
     return None
 
 
+class _HasHeadersAsMapping(Protocol):
+    @property
+    def headers(self) -> Mapping[str, str]: ...
+
+
+def header_value_matches_media_type(header_value: str, media_type: str) -> bool:
```
sobolevn: Let's remove this function. We don't need it and it creates an extra call in the hot path.

dmr/_compiled/negotiation.py
```diff
@@ -65,6 +65,23 @@ def accepted_type(
...
+    return accepted_type(header_value, (media_type,)) is not None
+
+
+def request_accepts(request: _HasHeadersAsMapping, media_type: str) -> bool:
```
sobolevn: Maybe `accepted_header`? Since we are going to remove the request.

sobolevn (commented):
Nice, great stuff! Please, add a changelog entry.

dmr/_compiled/negotiation.py
albertedwardson: not fully sure about the resulted name and the docstring though... naming things is hard...

sobolevn:
Wow, great work! I obviously missed the `','` trick :)

dmr/_compiled/negotiation.py
```diff
@@ -74,7 +120,9 @@ class _MediaTypeHeader:
     def __init__(self, type_str: str) -> None:
         # preserve the original parameters, because the order might be
         # changed in the dict
-        self.params_str = ''.join(type_str.partition(';')[1:])
+        self.params_str = (
+            f';{type_str.partition(";")[2]}' if ';' in type_str else ''
```
sobolevn: you can use `noqa` here, since we are going for speed, not readability.

```suggestion
            f';{type_str.partition(";")[2]}' if ';' in type_str else ''  # noqa: WPS237
```

dmr/negotiation.py
```diff
@@ -372,7 +322,6 @@ def get_conditional_types(
 
 def accepts(request: HttpRequest, content_type: str) -> bool:
```
sobolevn, line 366: This is a very different function. It checks that a selected renderer has a required content type.

Since you pass `*/*` it always returns `True` now, when it should not, that's why tests fail :)

Let's add an argument `strict: bool = True`, when it is true pass `''`, when false, pass `*/*`

dmr/_compiled/negotiation.py
```diff
@@ -48,13 +49,21 @@ def accepted_type(
         Otherwise the provided type is returned as-is.
 
     """
-    accepted_types = sorted(
-        [_MediaTypeHeader(typ) for typ in accept_value.split(',')],
-        key=_by_priority,
-        reverse=True,
-    )
+    if not accept_value or length_hint(provided_types) == 0:
```
albertedwardson: mypyc calls python version of length_hint unfortunately, will revert

albertedwardson:
marking this as ready since I won't have much free time to work on this anytime soon, sorry for unfinished work

dmr/_compiled/negotiation.py
```diff
@@ -22,13 +22,13 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-
+# flake8: noqa: WPS232, WPS402
```
sobolevn: We prefer to list this in `setup.py`, it is way easier to find.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ of requirements for an API to count as public.
 
 ### Features
 
+- Added `accepted_header` as a faster alternative to `django`'s `HttpRequest.accepts`, #854
```
sobolevn: Please, move this `WIP` section

sobolevn (commented):
Looks great! Thank you! 

But, you would need to add more tests to reach 100% coverage

sobolevn (approved):
Great stuff! Thanks a lot!

PR #859, opened 2026-04-08
feat: support orjson as alternative JSON backend for JsonParser/JsonRenderer

### LINKED ISSUE #857: Support `orjson` as a possible backend for `JsonParser` / `JsonRenderer`
Currently we always use `json` as a module in `JsonParser` and `JsonRenderer`: https://github.com/wemake-services/django-modern-rest/blob/69822ff0053e533331cfda884c44eba05e46e062/dmr/parsers.py#L132-L134

But, we can make this module customizable. We can pass `orjson` module there, since it also has a similar API.

Something like:

```python
class JsonParser(Parser):
    __slots__ = ('_json_module', ...)
    def __init__(self, json_module: _JsonModule = json, ...) -> None:
       self._json_module = json_module
```

And later `self._json_module.loads(...)`

This would need multiple test cases: install `orjson` as `unit-test` dependency and add unit tests for this new renderer / parser.
And we can document that with an example in the `integrations.rst` docs.

muratclk:
## Summary

Closes #857

Adds a `json_module` parameter to `JsonParser` and `JsonRenderer` so users can pass alternative JSON backends like `orjson` for faster JSON processing.

### Changes

- **`dmr/parsers.py`**: Added `__init__` with `json_module: ModuleType = json` parameter and `__slots__`. Uses `self._json_module.loads()` instead of `json.loads()`.
- **`dmr/renderers.py`**: Added `json_module` keyword param to `JsonRenderer.__init__`. For non-stdlib modules (like orjson), uses `dumps(obj, default=hook)` which handles orjson's bytes-returning API. `validation_parser` passes through the module.
- **`pyproject.toml`**: Added `orjson>=3.10,<4` to `unit-test` dependency group.
- **`tests/test_unit/test_parsers/test_orjson_parser.py`**: 5 tests covering valid JSON parsing, empty body handling, invalid JSON error, complex pydantic model roundtrip, and renderer bytes output.
- **`docs/pages/integrations.rst`**: Added "Alternative JSON backends" section with usage example.
- **`CHANGELOG.md`**: Added feature entry.

### Design decisions

- `json_module` defaults to stdlib `json` — fully backward compatible, no breaking changes.
- Renderer detects non-stdlib modules via `self._json_module is not json` to use the appropriate calling convention (`default=` instead of `cls=`, handling bytes return).
- Used `types.ModuleType` for the type hint, matching the pattern of accepting any module with `.loads()`/`.dumps()` interface.


dmr/parsers.py
```diff
@@ -98,11 +99,15 @@ class JsonParser(Parser):
 
     """
 
-    __slots__ = ()
+    __slots__ = ('_json_module',)
 
     content_type = 'application/json'
     """Works with ``json`` only."""
 
+    def __init__(self, json_module: ModuleType = json) -> None:
```
sobolevn: The annotation is not correct. We do not allow any module, we should expect a protocol with `loads` for `Parser` and `dumps` for `Renderer`. You can copy types from here: https://github.com/python/typeshed/blob/main/stdlib/json/__init__.pyi 

But, only copy params that we actually use in the logic.

These protocols can be placed into `dmr/internal

Also: 

```suggestion
    def __init__(self, content_type: str = 'application/json', *, json_module: ModuleType = json) -> None:
```

dmr/parsers.py
```diff
@@ -98,11 +99,15 @@ class JsonParser(Parser):
 
     """
 
-    __slots__ = ()
+    __slots__ = ('_json_module',)
 
     content_type = 'application/json'
```
sobolevn: Also, move this to the `__init__` method, it should be the first parameter with the default value.

dmr/renderers.py
```diff
@@ -147,10 +149,12 @@ def __init__(
         content_type: str = 'application/json',
         *,
         encoder_cls: type[DjangoJSONEncoder] = _DMREncoder,
+        json_module: ModuleType = json,
```
sobolevn: Same here about the type :)

dmr/renderers.py
```diff
@@ -171,6 +175,16 @@ def render(
         # msgspec returns `bytes`, we prefer to use `bytes` by default
         # and not to create extra strings when not needed in "fast" mode.
         # We don't really care about raw json implementation. It is a fallback.
+        if self._json_module is not json:
```
sobolevn: Ok, I see. Then the whole idea is not going to work.
Let's instead create a function attribute, which will be named `_json_dumps`. Here's even easier API:

```python
def _json_dumps(
    to_serialize: Any,
    serializer_hook: Callable[[Any], Any],
) -> bytes:
    return json.dumps(
        to_serialize,
        cls=self._encoder_cls,
        serializer_hook=serializer_hook,
        # We need this flag to produce the same results as `msgspec`:
        separators=(',', ':'),
    ).encode('utf8')
```

and then use it like so:

```python
def __init__(self, *, json_dumps: TYPE = _json_dumps) -> None:
    self._json_dumps = json_dumps
```

and

```python
return self._json_dumps(to_serialize, serializer_hook)
```

and just create the same function for `orjson` in tests.
We don't want to provide this function or depend on `orjson` in dmr directly :)

Since it is a very simple case, we can just document it.

docs/pages/integrations.rst
```diff
@@ -10,6 +10,33 @@ Integrations
...
+-------------------------
+
+By default, ``JsonParser`` and ``JsonRenderer`` use Python's built-in
+``json`` module. You can swap in a faster backend like
+`orjson <https://github.com/ijl/orjson>`_ by passing the ``json_module``
+parameter:
+
+.. code-block:: python
+
+    import orjson
```
sobolevn: Please, use `.. literalinclude`, we need to type check all examples.

tests/test_unit/test_parsers/test_orjson_parser.py
```diff
@@ -0,0 +1,156 @@
...
+from dmr import Body, Controller
+from dmr.parsers import JsonParser
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.renderers import JsonRenderer
+from dmr.settings import Settings
+from dmr.test import DMRRequestFactory
+
+
+@pytest.fixture(autouse=True)
+def _clear_parser_and_renderer(
```
sobolevn:
```suggestion
def _orjson_parser_and_renderer(
```

sobolevn (commented):
Great stuff! Thanks a lot! Some small design ideas :)

sobolevn:
Hi, @muratclk! Thanks a lot for your help. Sorry, but I had to finish this myself, because I really want to release a new version today :)

And this feature must 100% be in it.
You will be fully credited for this PR. Amazing work!

muratclk:
> Hi, @muratclk! Thanks a lot for your help. Sorry, but I had to finish this myself, because I really want to release a new version today :)
> 
> And this feature must 100% be in it. You will be fully credited for this PR. Amazing work!

Hi sobolevn I just saw your review. Thanks for finishing the unfinished work. I’m glad if I was able to contribute to the project.

sobolevn:
@muratclk please send more PRs :) 
We need a lot of people to help!

PR #875, opened 2026-04-10
Update docs: polish landing page

Satevg:
# I have made things!

I'm not a designer or a FE person, but it seemed that the docs main page could be improved, since it's the first page, that potential user visits..

Tried to keep changes minimal. Up to you guys to decide if that's fine, I just gave a try

## AI Policy

## Checklist

### DIFF
--- a/docs/_static/css/landing.css
+++ b/docs/_static/css/landing.css
@@ -1,0 +1,116 @@
+/* Subtle green radial glow from the top, spanning the full page */
+body {
+  background-image: radial-gradient(
+    ellipse 80% 50% at 50% 0%,
+    rgba(45, 164, 78, 0.09) 0%,
+    transparent 60%
+  );
+  background-attachment: fixed;
+}
+
+html.dark body {
+  background-image: radial-gradient(
+    ellipse 80% 50% at 50% 0%,
+    rgba(63, 185, 80, 0.09) 0%,
+    transparent 60%
+  );
+}
+
+/* Tighten hero spacing */
+.yue {
+  padding-bottom: 1rem !important;
+}
+
+.yue section:first-of-type h1 {
+  margin-top: 1.5rem !important;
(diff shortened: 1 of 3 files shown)


sobolevn (approved):
Yes, it looks much better now!

Image

Satevg:
@sobolevn I'd like to dive into the code and do something, since I do have some spare time (not sure how much though). In issue tracker, I see that all "good first use" issues are assigned already. Do you have any ideas what else could be picked up?

sobolevn:
@Satevg yes, right now we need a cursor-based paginator :)
It should not be too hard.

Here's an example implementation: https://github.com/photocrowd/django-cursor-pagination/blob/master/cursor_pagination.py

API design:
- Ideally it should be a subclass for Django's native `Paginator` object
- It should work very similar to it
- Ideally it should be possible to use it with our own `Paginated` class, but we are completely free to add a new interface for it :)

Would you like to help with this feature? :)

Another option is to move to `just` from `make`, which is also on my bucket list.

Satevg:
> @Satevg yes, right now we need a cursor-based paginator :) It should not be too hard.
> 
> Here's an example implementation: https://github.com/photocrowd/django-cursor-pagination/blob/master/cursor_pagination.py
> 
> API design:
> 
> * Ideally it should be a subclass for Django's native `Paginator` object
> * It should work very similar to it
> * Ideally it should be possible to use it with our own `Paginated` class, but we are completely free to add a new interface for it :)
> 
> Would you like to help with this feature? :)
> 
> Another option is to move to `just` from `make`, which is also on my bucket list.

I'd start with `just`, since it's a quick win before the weekend :smile: 
I'll have a look at new paginator next week hopefully.
Also noticed that "router tags" recent feature request. Might be a good candindate for the 1st PR as well.

PR #891, opened 2026-04-13
apply dmr json dump to test mixin and factories

Nalsher:
# I have made things!
Apply _encode_json to DMRMixin and DMRRequestFactories
Cheers!
-->

## AI Policy

- [x] I have read and agree to the [AI Policy](https://github.com/wemake-services/django-modern-rest/blob/master/.github/AI_POLICY.md), removed any "Co-Authored-By" lines attributing coding agents, and manually reviewed the final result

## Checklist

PR #935, opened 2026-04-20
use msgspec in dmr_rf and dmr_client

### LINKED ISSUE #889: Optimize `dmr_rf` and `dmr_client` test features for faster tests
We provide our own fixtures that are subclasses of Django's builtin one.

Django has this code inside: https://github.com/django/django/blob/7dc826b9758d634623a6f5ca05d0ca2048a0ce48/django/test/client.py#L450-L458

```python
    def _encode_json(self, data, content_type):
        """
        Return encoded JSON if data is a dict, list, or tuple and content_type
        is application/json.
        """
        should_encode = JSON_CONTENT_TYPE_RE.match(content_type) and isinstance(
            data, (dict, list, tuple)
        )
        return json.dumps(data, cls=self.json_encoder) if should_encode else data
```

We need to optimize this method, because `json.dumps` is slow, but `msgspec` is fast.
It won't be noticable on 1 tests, but will be on 2k tests (which is the case for our own test base).

See how we conditionally dump our schema here: https://github.com/wemake-services/django-modern-rest/blob/master/dmr/openapi/dump.py

Maybe we should reuse the same for tests?

Satevg:
# I have made things!

use msgspec in `dmr_rf` and `dmr_client`

Closes #889 

## AI Policy

## Checklist


sobolevn (commented):
Please, take a look at https://github.com/wemake-services/django-modern-rest/pull/891 You can finish it if you want, ideally by cherry-picking commits from there to your branch. I will keep both of you as authors in this case :)

Satevg:
> Please, take a look at #891 You can finish it if you want, ideally by cherry-picking commits from there to your branch. I will keep both of you as authors in this case :)

Omg, I'm stupid, wouldn't do this if I'd saw open MR :smile: :man_facepalming:

sobolevn (approved):
Looks great! Thanks a lot!

PR #947, opened 2026-04-21
fix(security/token): catch PyJWTError instead of DecodeError in JWToken.encode()

### LINKED ISSUE #943: JWToken.encode() catches jwt.DecodeError instead of encoding-related exceptions
### What's wrong?

JWToken.encode() in `dmr/security/jwt/token.py` catches `jwt.DecodeError` when an error occurs during token encoding. jwt.DecodeError is raised during decoding, not encoding. If encoding fails due to an  invalid key format (jwt.exceptions.InvalidKeyError) or an unsupported algorithm (jwt.exceptions.InvalidAlgorithmError), those exceptions are not subclasses of jwt.DecodeError and will propagate uncaught - resulting in an unhandled 500 with a raw Python traceback instead of a clean InternalServerError.                                                                                                      
                                                                                                                                                                                                            
```python                                                                                                                                                                                      
  try:                                                                                                                                                                                                      
      return jwt.encode(payload=..., key=secret, algorithm=algorithm)
  except (jwt.DecodeError, NotImplementedError):                            <!---------- HERE                                                                                                                                 
      raise InternalServerError('Failed to encode token') from None
```



### How it should be?

Catch `jwt.exceptions.PyJWTError`, which is the common base class for all PyJWT exceptions including InvalidKeyError and InvalidAlgorithmError

### Used versions

0.7.0

### OS information

MacOS

Peopl3s:
# I have made things!

## AI Policy

## Checklist

- [X] I have created at least one test case for the changes I have made

## Related issues

Closes #943


sobolevn (approved):

sobolevn:
I fixed the test in https://github.com/wemake-services/django-modern-rest/pull/949

PR #950, opened 2026-04-21
fix: enforce max_length=100 on TagSchema and RoleSchema name fields (#944)

### LINKED ISSUE #944: TagSchema and RoleSchema do not enforce max_length, causing unhandled DB errors
### What's wrong?

The Tag and Role Django models define name = CharField(max_length=100), but the corresponding Pydantic schemas accept any-length strings.

```python
class TagSchema(pydantic.BaseModel):
      name: str  <!-----------------------------HERE                                                                                                                                                                 
                                                        
  class RoleSchema(pydantic.BaseModel):                                                                                                                                                                     
      name: str  <!-----------------------------HERE 
```
When a client sends a name longer than 100 characters, Pydantic validation passes (HTTP 422 is never raised), and the value reaches the database layer where PostgreSQL raises DataError: value too long for type character varying(100). This results in an unhandled 500 instead of a proper 422 validation error. On SQLite the value is silently stored without truncation, masking the bug in tests.

### How it should be?

Add max_length constraints to match the model definition:
 
```python                                                                                                                                                                                                          from pydantic import Field
                                                                                                                                                                                                            class TagSchema(pydantic.BaseModel):                  
    name: str = Field(max_length=100)
                                                                                                                                                                                                            class RoleSchema(pydantic.BaseModel):
    name: str = Field(max_length=100)
```

### Used versions

0.7.0

### OS information

MacOS

mvanhorn:
## Summary

Resolves #944.

The Django \`Tag\` and \`Role\` models in \`django_test_app/server/apps/model_fk/models.py\` define \`name = CharField(max_length=100)\`, but \`TagSchema\` / \`RoleSchema\` in the matching \`serializers.py\` accepted any-length strings. Over-long names passed Pydantic validation and failed at the database layer: PostgreSQL raised \`DataError: value too long for type character varying(100)\` (unhandled 500), SQLite silently truncated, masking the bug in tests.

## Changes

- \`django_test_app/server/apps/model_fk/serializers.py\`: \`name: str = pydantic.Field(max_length=100)\` on both schemas. Matches the \`PageQuery\` pattern already used in the same file (\`pydantic.Field(default=10, ge=1, le=100)\`).
- \`tests/test_integration/test_model_fk/test_model_fk.py\`: parametrized regression test \`test_user_create_name_too_long_rejected\` that posts a 101-character name in either \`role.name\` or \`tags[0].name\` and asserts a rejection with the \`at most 100 characters\` value-error detail.
- \`CHANGELOG.md\`: new bullet under \`## WIP > ### Bugfixes\`.

## Note on the status code

The issue body suggested this should raise \`HTTP 422\`, but DMR's default validation error response is \`HTTP 400\` (same as \`test_user_create_unique_email_error\` uses \`HTTP 409\` for its own conventions). The test asserts \`400\` and pins the Pydantic error detail (\`type: value_error\`, message contains \`at most 100 characters\`) so the regression is tight regardless of the status code convention.

## Test plan

- \`uv run pytest tests/test_integration/test_model_fk/\` -> 8 passed (4 new parametrized cases + 4 existing)
- \`uv run pytest tests/\` -> 2220 passed, 46 skipped, 35 subtests passed (no regressions)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,10 +36,6 @@ of requirements for an API to count as public.
   `HttpResponse` with a different content type than the negotiated
   renderer, #711
 - Fixed `collectstatic` failure when using `ManifestStaticFilesStorage`, #927
-- Fixed `TagSchema` and `RoleSchema` in the `model_fk` test app accepting
-  names longer than the `max_length=100` database constraint, which raised a
-  500 on PostgreSQL and silently truncated on SQLite instead of returning
-  422, #944
 
 ### Misc
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ of requirements for an API to count as public.
   `HttpResponse` with a different content type than the negotiated
   renderer, #711
 - Fixed `collectstatic` failure when using `ManifestStaticFilesStorage`, #927
+- Fixed `TagSchema` and `RoleSchema` in the `model_fk` test app accepting
+  names longer than the `max_length=100` database constraint, which raised a
+  500 on PostgreSQL and silently truncated on SQLite instead of returning
+  422, #944
 
 ### Misc
 
--- a/tests/test_integration/test_model_fk/test_model_fk.py
+++ b/tests/test_integration/test_model_fk/test_model_fk.py
@@ -70,8 +70,17 @@ def test_user_create_name_too_long_rejected(
     validation and failed at the database layer with an unhandled 500
     (PostgreSQL) or were silently truncated (SQLite).
     """
-    role_name = 'r' * 101 if over_long_field == 'role' else faker.name()
-    tag_name = 't' * 101 if over_long_field == 'tag' else faker.name()
+    min_chars = 101
+    role_name = (
+        faker.pystr(min_chars=min_chars, max_chars=min_chars + 10)
+        if over_long_field == 'role'
+        else faker.name()
+    )
+    tag_name = (
+        faker.pystr(min_chars=min_chars, max_chars=min_chars + 10)
+        if over_long_field == 'tag'
+        else faker.name()
+    )
     request_data = {
         'email': faker.email(),
         'role': {'name': role_name},
--- a/tests/test_integration/test_model_fk/test_model_fk.py
+++ b/tests/test_integration/test_model_fk/test_model_fk.py
@@ -57,6 +57,39 @@ def test_user_create_models_example(
 
 
 @pytest.mark.django_db
+@pytest.mark.parametrize('over_long_field', ['role', 'tag'])
+def test_user_create_name_too_long_rejected(
+    dmr_client: DMRClient,
+    faker: Faker,
+    over_long_field: str,
+) -> None:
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -36,6 +36,10 @@ of requirements for an API to count as public.
   `HttpResponse` with a different content type than the negotiated
   renderer, #711
 - Fixed `collectstatic` failure when using `ManifestStaticFilesStorage`, #927
+- Fixed `TagSchema` and `RoleSchema` in the `model_fk` test app accepting
+  names longer than the `max_length=100` database constraint, which raised a
+  500 on PostgreSQL and silently truncated on SQLite instead of returning
+  422, #944
```
```suggestion
```

this is a test-only change, it does not need a changelog entry.

tests/test_integration/test_model_fk/test_model_fk.py
```diff
@@ -56,6 +56,39 @@ def test_user_create_models_example(
...
+    over_long_field: str,
+) -> None:
+    """Names longer than ``Tag``/``Role`` ``max_length=100`` are rejected.
+
+    Regression for #944: previously the Pydantic schemas did not mirror the
+    Django model ``max_length=100`` constraint, so over-long names passed
+    validation and failed at the database layer with an unhandled 500
+    (PostgreSQL) or were silently truncated (SQLite).
+    """
+    role_name = 'r' * 101 if over_long_field == 'role' else faker.name()
```
sobolevn:
```suggestion
    min_chars = 100
    role_name = faker.pystr(min_chars=min_chars, max_chars=min_chars + 10) if over_long_field == 'role' else faker.name()
```

sobolevn: the same for the second field :)

mvanhorn:
Thanks @sobolevn - both applied in 3cc3070.

- CHANGELOG.md: removed the #944 entry. Agreed, test-only.
- tests/.../test_model_fk.py: switched both fields to `faker.pystr(min_chars=..., max_chars=...)`.

One nit I flagged myself: I used `min_chars = 101` (not `100`) so every generated string is strictly above the `max_length=100` constraint. With `min_chars=100`, `faker.pystr` can return a length-100 string, which Pydantic's `max_length=100` accepts (inclusive), and the test would intermittently flip from "must reject" to "must accept." Happy to change it to `100` if you'd rather keep the exact suggested line - just flag it.

tests/test_integration/test_model_fk/test_model_fk.py
```diff
@@ -56,6 +56,48 @@ def test_user_create_models_example(
...
+    request_data = {
+        'email': faker.email(),
+        'role': {'name': role_name},
+        'tags': [{'name': tag_name}],
+    }
+
+    response = dmr_client.post(
+        reverse('api:model_fk:user'),
+        data=request_data,
+    )
```
sobolevn:
```suggestion
    response = dmr_client.post(
        reverse('api:model_fk:user'),
        data={
            'email': faker.email(),
            'role': {'name': role_name},
            'tags': [{'name': tag_name}],
        },
    )
```

sobolevn (approved):

PR #958, opened 2026-04-22
Speedup slow test_schemathesis in local environment

Satevg:
# I have made things!

When developing locally, `just test` is a common command, though it's very slow on a schematesis test. 

What's changed:

1. Eliminating the `DEBUG=True` pass              

`@schema.parametrize()` expands test_schemathesis into N separate tests  one per API operation in the schema. The parent conftest's `params=[True, False]` doubles that to 2N runs. The `DEBUG=True` half all skip immediately, but "immediately" isn't free: each still goes through fixture setup - transactional_db and admin_user (creates a DB row) - before the test body even runs.

  2. Reduce max_examples locally
  
  Default 100 generated request payloads per API operation is probably overhead for local testing (though kept 100 unchanged for the CI)
  
  ----------

`time uv run pytest tests/test_integration/test_openapi/test_schema.py`
On my local machine **on `master`:**
```
real    2m0,807s
user    1m56,825s
sys     0m3,918s
```
On **this branch:**

```
real    0m51,862s
user    0m50,058s
sys     0m1,793s

```

## AI Policy

## Checklist

### DIFF
--- a/tests/test_integration/conftest.py
+++ b/tests/test_integration/conftest.py
@@ -26,6 +26,7 @@ def tracecov_map() -> 'tracecov.CoverageMap | None':
 def _set_dmr_settings(settings: LazySettings) -> None:
     settings.DMR_SETTINGS = {
         Settings.openapi_examples_seed: 1,
+        # It might be already defined in some other place:
         **getattr(settings, 'DMR_SETTINGS', {}),
     }
 
--- a/tests/test_integration/conftest.py
+++ b/tests/test_integration/conftest.py
@@ -22,16 +22,17 @@ def tracecov_map() -> 'tracecov.CoverageMap | None':
     return tracecov.CoverageMap.from_dict(schema.convert())
 
 
+@pytest.fixture(autouse=True)
+def _set_dmr_settings(settings: LazySettings) -> None:
+    settings.DMR_SETTINGS = {
+        Settings.openapi_examples_seed: 1,
+        **getattr(settings, 'DMR_SETTINGS', {}),
+    }
+
+
 @pytest.fixture(autouse=True, params=[True, False])
 def _modify_integration_settings(
     settings: LazySettings,
     request: pytest.FixtureRequest,
 ) -> None:
-    # Django common settings:
     settings.DEBUG = request.param  # We run tests in both modes.
-    # Our own settings:
-    settings.DMR_SETTINGS = {
-        Settings.openapi_examples_seed: 1,
-        # It might be already defined in some other place:
-        **getattr(settings, 'DMR_SETTINGS', {}),
-    }
--- a/tests/test_integration/test_openapi/test_schema.py
+++ b/tests/test_integration/test_openapi/test_schema.py
@@ -81,12 +81,11 @@ st.openapi.format(
 )
 def test_schemathesis(
     case: st.Case,
-    settings: LazySettings,
     tracecov_map: 'tracecov.CoverageMap | None',
 ) -> None:
     """Ensure that API implementation matches the OpenAPI schema."""
-    if settings.DEBUG or tracecov_map is None:
-        pytest.skip(reason='DEBUG=True or missing `tracecov`')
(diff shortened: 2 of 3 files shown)


kondratevdev:
I'm not sure if it's worth putting the configuration in the code for local checks. You can always change this number in `schemathesis.toml`.
It is also important to understand that schemathesis generates UP to 100 queries, if the endpoint has fewer test options, then there will be fewer test cases.

sobolevn (commented):
Let's remove `max-examples` from `schemathesis.toml` then. I don't want to have two sources of truth

tests/test_integration/test_openapi/test_schema.py
```diff
@@ -61,14 +76,16 @@ def api_schema(transactional_db: None, admin_user: User) -> OpenApiSchema:
 
 
 @schema.parametrize()
+@h_settings(
+    max_examples=_MAX_EXAMPLES,
```
kondratevdev:
```suggestion
    @h_settings(max_examples=_MAX_EXAMPLES)
```

tests/test_integration/conftest.py
```diff
@@ -22,16 +22,17 @@ def tracecov_map() -> 'tracecov.CoverageMap | None':
     return tracecov.CoverageMap.from_dict(schema.convert())
 
 
+@pytest.fixture(autouse=True)
+def _set_dmr_settings(settings: LazySettings) -> None:
+    settings.DMR_SETTINGS = {
+        Settings.openapi_examples_seed: 1,
+        **getattr(settings, 'DMR_SETTINGS', {}),
```
kondratevdev, line 30:
```suggestion
       # It might be already defined in some other place:
        **getattr(settings, 'DMR_SETTINGS', {}),
```

kondratevdev: Please give it back. This is an important hint :)

Satevg:
> It is also important to understand that schemathesis generates UP to 100 queries, if the endpoint has fewer test options, then there will be fewer test cases.


Fair point, but without limiting `max_examples` to 25, the test gives me +~ 30-40 seconds locally. So it does a lot: > 25 <100

tests/test_integration/test_openapi/test_schema.py
```diff
@@ -7,12 +8,18 @@
...
+from hypothesis import settings as h_settings
 from hypothesis import strategies
 from schemathesis.specs.openapi.schemas import OpenApiSchema
 
 from django_test_app.server.wsgi import application
 from dmr.validation import ResponseValidator
 
+_LOCAL_MAX_EXAMPLES = 25
+_MAX_EXAMPLES = (
+    h_settings().max_examples if os.environ.get('CI') else _LOCAL_MAX_EXAMPLES
```
sobolevn:
```suggestion
    100 if os.environ.get('CI') else _LOCAL_MAX_EXAMPLES
```

tests/test_integration/test_openapi/test_schema.py
```diff
@@ -7,12 +8,18 @@
...
 from django.contrib.auth.models import User
 from django.urls import reverse
+from hypothesis import settings as h_settings
 from hypothesis import strategies
 from schemathesis.specs.openapi.schemas import OpenApiSchema
 
 from django_test_app.server.wsgi import application
 from dmr.validation import ResponseValidator
 
+_LOCAL_MAX_EXAMPLES = 25
```
sobolevn:
```suggestion
_LOCAL_MAX_EXAMPLES: Final = 25
```

tests/test_integration/test_openapi/test_schema.py
```diff
@@ -7,12 +8,18 @@
...
+_MAX_EXAMPLES = (
```
sobolevn:
```suggestion
_MAX_EXAMPLES: Final = (
```

sobolevn (approved):

PR #959, opened 2026-04-22
fix(throttling): replace per-process locks with atomic cache increment

### LINKED ISSUE #942: Throttle might have TOCTOU race condition on some cache backends
### What's wrong?

The throttle implementation in `dmr/endpoint.py` and `dmr/throttling/base.py` uses threading.Lock / asyncio.Lock to protect the cache read-modify-write cycle:   
                                             
```python                                                                                                                                                                                                          
  with self._sync_lock:
      throttle(self, controller)                                                                                                                                                                            
```          
                                                                                                                                               
These locks are per-process Python objects. In a multi-worker deployment (e.g. Gunicorn with 4 workers, uvicorn with multiple processes), each worker has its own independent lock instance. Two workers  can simultaneously read the same counter from a shared cache backend (Redis/Memcached), both pass the limit check, and bot write an incremented value - effectively allowing up to `max_requests * num_workers` requests through before throttling kicks in.                                                                                                                                                  
                                                        
  Example with max_requests=5, num_workers=4:
  1. All 4 workers read count=5 from Redis
  2. All 4 pass the count < max_requests check (5 < 5 is false... but 4 < 5 is true for count=4)                                                                                                            
  3. All 4 write count=5                                                                        
  4. Up to 4 extra requests bypass the limit

There is a similar thing in litestar https://github.com/litestar-org/litestar/issues/4700

### How it should be?

At minimum, document the single-process limitation clearly so operators know they need a Redis-backed atomic throttle backend in production. Or use atomic cache operations for the increment step with Redis

### Used versions

0.7.0

### OS information

MacOS

Peopl3s:
# I have made things!

## AI Policy

## Checklist

## Related issues
Closes #942

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,13 @@ of requirements for an API to count as public.
 ### Misc
 
 - Optimized `dmr_client` and `dmr_rf` test fixtures to use `msgspec` for JSON encoding when available, #889
+- Fixed a race condition in `SyncThrottle` / `AsyncThrottle` under multi-worker deployments
+  (Gunicorn, uvicorn `--workers N`): per-process `threading.Lock` / `asyncio.Lock` guards
+  were removed and replaced with an atomic `cache.add` + `cache.incr` increment on the
+  `DjangoCache` backend. For Redis and Memcached this makes the counter safe across all
+  workers; `LocMemCache` (typically used in tests) keeps the single-process guarantee it
+  always had. Custom backends that do not implement `incr_with_expiry` (return `None`)
+  transparently fall back to the previous read-modify-write path.
 
 
 ## Version 0.7.0 (2026-04-14)
--- a/dmr/endpoint.py
+++ b/dmr/endpoint.py
@@ -1,6 +1,4 @@
-import asyncio
 import inspect
-import threading
 from collections.abc import Awaitable, Callable, Mapping, Sequence, Set
 from functools import wraps
 from http import HTTPStatus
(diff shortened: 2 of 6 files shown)


sobolevn (commented):
Thanks a lot for your help! I would like to re-iterate the design in several ways.

Important: we are still in Alpha. We can break anything we want. So, now our task is to create the most advanced, flexible, and awesome design.

So, here are my main concerns:

1. Now when I think about it, I don't quite like that `Backend` has 4 methods. What if some backends won't support async / sync API? We need to split backends into sync and async ones. 
2. This means that we would have to decide which default backend to use: `SyncDjangoCache` or `AsyncDjangoCache` inside the contructor
3. I would be **very** happy to see algorithms as pure objects, this way we can compile them with `mypyc` to gain a lot of speed
4. Let's re-think the backend API. `get` and `set` can be refactored into something atomic
5. I like that endpoint has no locks now, let's keep it this way

Current implementation seems a bit too clunky :(
Let's brainshtorm together!

sobolevn:
@Peopl3s what do you think about this idea: we can keep `DjangoSession` backend unsafe, in most cases `100` or `101` requests do not really matter. We can just keep everything as-is.
And just document this properly.

But, we can create `RedisAtomic` backend that will ensure that all per-process workers respect the limit. For very specific, for example, paid API limits. 

`RedisAtomic` can use explicit `Redis` API based on pipelines, so we can be sure that it is atomic.

Happy to hear your feedback :)

Peopl3s:
@sobolevn The RedisAtomic idea sounds great. As far as I understand, although I might be confused, Redis pipelines don't provide strict atomicity. So, you'd have to invent something like a Lua-script

sobolevn:
> In addition, pipelines can also ensure the buffered commands are executed atomically as a group. 

I am pretty sure they are atomic: https://redis.readthedocs.io/en/stable/advanced_features.html#pipelines

It can be disabled by `.pipeline(transaction=False)`, but by deafult they are on.

PR #964, opened 2026-04-23
fix: fall back to non-streaming default when Accept only matches stream

### LINKED ISSUE #962: SSE endpoints return 500 on 4xx/5xx errors when Accept is `text/event-stream` without `application/json`
### What's wrong?

When a streaming endpoint (e.g. `SSEController`) receives a request with `Accept: text/event-stream` **without** `application/json` as a fallback — the canonical case for browser `EventSource` — any response that isn't a successful stream (4xx validation errors, `APIError` raised before the first `yield`, etc.) crashes with `DataRenderingError: SSERenderer can only render SSE protocol instances, got <class 'dict'>` and the client gets a 500 instead of the documented 4xx/5xx JSON body.

### Reproducer

```python
# views.py
from collections.abc import AsyncIterator
from http import HTTPStatus

import msgspec
from dmr import ResponseSpec, modify
from dmr.plugins.msgspec import MsgspecSerializer
from dmr.response import APIError
from dmr.streaming.sse import SSEController, SSEvent


class _Payload(msgspec.Struct):
    n: int


class _Err(msgspec.Struct):
    detail: list


class Events(SSEController[MsgspecSerializer]):
    error_model = _Err

    @modify(extra_responses=[ResponseSpec(_Err, status_code=HTTPStatus.NOT_FOUND)])
    async def get(self) -> AsyncIterator[SSEvent[_Payload]]:
        # Any error raised before the first yield reproduces the bug:
        raise APIError(
            {"detail": [{"msg": "nope", "type": "user_msg"}]},
            status_code=HTTPStatus.NOT_FOUND,
        )
```

```bash
# Works — curl defaults to Accept: */*, non-streaming negotiation picks JSON:
$ curl -i http://localhost:8000/events/
HTTP/1.1 404 Not Found
Content-Type: application/json
{"detail":[{"msg":"nope","type":"user_msg"}]}

# Broken — mirrors what EventSource sends:
$ curl -i -H 'Accept: text/event-stream' http://localhost:8000/events/
HTTP/1.1 500 Internal Server Error
```

### Stacktrace

```
File ".../dmr/endpoint.py", line 365, in decorator
    func_result = await self.handle_async_error(controller, exc)
File ".../dmr/endpoint.py", line 270, in handle_async_error
    return self._global_error_handler(controller, exc)
File ".../dmr/errors.py", line 308, in global_error_handler
    return controller.to_error(controller.format_error(exc), ...)
File ".../dmr/controller.py", line 297, in to_error
    return self.to_response(raw_data, ..., renderer=(
        renderer or request_renderer(self.request, use_nonstreaming_renderer=True)
    ))
File ".../dmr/response.py", line 253, in build_response
    serializer.serialize(raw_data, renderer=renderer)
File ".../dmr/streaming/sse/renderer.py", line 74, in render
    raise DataRenderingError(
dmr.exceptions.DataRenderingError: SSERenderer can only render SSE protocol instances, got <class 'dict'>
```

### Root cause

[`dmr/negotiation.py:184-196`][neg] — `ResponseNegotiator.__call__` does two
negotiations for streaming endpoints:

```python
renderer = _negotiate_renderer(request, self._renderers, default=self._default)
request.__dmr_renderer__ = renderer                         # step 1: matches SSE  ✓
if self._streaming:
    request.__dmr_nonstreaming_renderer__ = _negotiate_renderer(
        request, self._non_streaming_renderers,
        default=self._non_streaming_default,
    )                                                       # step 2: raises  ✗
```

Step 1 succeeds (SSE renderer matches `text/event-stream`). Step 2 iterates only over non-streaming renderers (`application/json`); `_negotiate_renderer` in [`dmr/internal/negotiation.py:94-106`][inner] uses its `default=` **only** when the `Accept` header is absent and raises `NotAcceptableError` otherwise. `__dmr_nonstreaming_renderer__` is never set.

The exception propagates into `Endpoint._async_endpoint`'s generic `except Exception` branch → `handle_async_error` → `_global_error_handler` → `controller.to_error(format_error(exc))`. `to_error` looks up `__dmr_nonstreaming_renderer__` (missing → `None`), falls back to `__dmr_renderer__` (the SSE renderer), and passes a dict to `SSERenderer.render` which crashes.

The same path also breaks response validation in `DEBUG`: `dmr/validation/response.py:61-64` uses `request_renderer(use_nonstreaming_renderer=True)` and ends up calling `validation_parser` on `StreamingRenderer`, which raises `NotImplementedError('StreamingRenderer must not return a parser')`.

### Why CI doesn't catch this

- `tests/test_unit/test_streaming/test_streaming_negotiation.py` always sends a combined Accept (`text/event-stream, application/json`) — step 2 succeeds.
- SSE error tests (`test_sse_validation.py`, `test_sse_error_handling.py`) use `dmr_async_rf.get('/…/')` without an explicit `Accept`, so `_negotiate_renderer` hits the `if accept is None: return default` branch and never raises.

No test currently exercises the `Accept: text/event-stream`-only + error-path combination, which is exactly the default behavior of browser `EventSource`.

### Docs acknowledge the client-side contract

`docs/examples/streaming/streaming_negotiation.py` and `docs/pages/streaming/common.rst` show combined Accept. The practical problem:
**browser `EventSource` cannot set custom headers**, so the recommended client-side contract is only achievable with `@microsoft/fetch-event-source` or similar polyfills — heavy just to get a 4xx body through.

### How it should be?

When the streaming negotiation has already succeeded, a non-matching non-streaming renderer shouldn't cascade into a 500. Error responses and response-validator paths should fall back to the configured non-streaming default:

```diff
--- a/dmr/negotiation.py
+++ b/dmr/negotiation.py
@@ -10,6 +10,7 @@ from typing_extensions import overload
 from dmr.compiled import accepted_type
 from dmr.exceptions import NotAcceptableError
 from dmr.internal.negotiation import negotiate_renderer as _negotiate_renderer
@@ -184,11 +185,18 @@ class ResponseNegotiator:
         renderer = _negotiate_renderer(
             request,
             self._renderers,
             default=self._default,
         )
         request.__dmr_renderer__ = renderer  # type: ignore[attr-defined]
         if self._streaming:
-            request.__dmr_nonstreaming_renderer__ = _negotiate_renderer(  # type: ignore[attr-defined]
-                request,
-                self._non_streaming_renderers,
-                default=self._non_streaming_default,
-            )
+            try:
+                non_streaming = _negotiate_renderer(
+                    request,
+                    self._non_streaming_renderers,
+                    default=self._non_streaming_default,
+                )
+            except NotAcceptableError:
+                # Main (streaming) negotiation already succeeded. A
+                # non-streaming renderer is only needed for 4xx/5xx
+                # error bodies and response validation — use the
+                # configured default so those paths keep working for
+                # clients that only accept the streaming media type
+                # (e.g. browser EventSource).
+                non_streaming = self._non_streaming_default
+            request.__dmr_nonstreaming_renderer__ = non_streaming  # type: ignore[attr-defined]
         return renderer
```
### Regression test

Add to `tests/test_unit/test_streaming/test_streaming_negotiation.py` (or a dedicated `ResponseNegotiator` unit-test file) a case that mirrors browser `EventSource`:

```python
def test_stream_only_accept_falls_back_to_non_streaming_default(
    # ... same fixture setup as existing tests
):
    """`Accept: text/event-stream` without `application/json` — the
    real-world browser `EventSource` case. Must not raise
    `NotAcceptableError`: `__dmr_nonstreaming_renderer__` must be
    populated with the default so `to_error` and the response
    validator can serialize dict error bodies.
    """
    request = dmr_async_rf.get(
        '/sse/',
        headers={'Accept': str(ContentType.event_stream)},
    )
    renderer = negotiator(request)

    assert renderer is sse_renderer
    assert request.__dmr_renderer__ is sse_renderer
    assert request.__dmr_nonstreaming_renderer__ is json_renderer
```

Also assert that an Accept which matches **no** renderer on a streaming endpoint (`application/xml`) still raises `NotAcceptableError` — fallback only applies when the main negotiation succeeded.

### Why existing tests stay green

- `test_streaming_negotiation.py` sends `Accept: text/event-stream, application/json` — matches on the first attempt, never enters `except`.
- Error tests (`test_sse_validation.py`, `test_sse_error_handling.py`) use `dmr_async_rf.get('/…/')` without `Accept` — hits `if accept is None: return default`.
- Non-streaming endpoint tests (`test_negotiation.py`) have `_streaming=False`
  — the `if self._streaming:` branch isn't touched.

### Alternatives considered

1. **`strict=False` parameter for `_negotiate_renderer`**: more explicit, but changes a mypyc-compiled helper used in multiple call sites. Scoped fix above is equivalent for the affected spot and safer.
2. **Client-side**: document combined Accept + `@microsoft/fetch-event-source`. Already possible, but browser `EventSource` is in the HTML spec and shouldn't require a polyfill for error bodies.
3. **Server-side workaround in user code**: subclass `ResponseNegotiator`, swap `endpoint_cls`. We currently ship this in our app; it's the kind of boilerplate that belongs in the framework.

### Non-goals

- Non-streaming endpoints still raise `NotAcceptableError` on Accept mismatch (correct behavior — unchanged).
- Streaming happy path unchanged.
- No public API changes to `ResponseNegotiator.__init__` / `__call__`.

Happy to follow up with a PR: fix + regression test + CHANGELOG entry.

### Used versions

```
pip --version
pip 24.x  # via uv-managed venv on Python 3.13.11

pip freeze | rg -i '^(django-modern-rest|django|msgspec|pydantic)'
django==5.2.13
django-modern-rest==0.7.0
msgspec==0.21.1
```

Handler: ASGI (`uvicorn`). Reproduced on 0.7.0 `master` — `dmr/negotiation.py` and `dmr/internal/negotiation.py` are identical to the versions shipped in 0.7.0 at the lines referenced above.

### OS information

Linux/x86_64 container running the same `uvicorn` + Django 5.2 stack.

MikeVL:
When a streaming endpoint received `Accept: text/event-stream` without `application/json` — the default browser `EventSource` case — the `ResponseNegotiator` raised `NotAcceptableError` during the non-streaming pass. Any 4xx/5xx error body (or response-validation path) then tried to render a dict through `SSERenderer` and crashed with a 500.

The main (streaming) negotiation has already succeeded at that point, so cascading to a 500 is wrong. Catch `NotAcceptableError` in that second pass and fall back to the configured non-streaming default.

Closes #962

# I have made things!

## AI Policy

## Checklist

## Related issues

- Closes #962

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,11 @@ of requirements for an API to count as public.
 - Fixed a bug that `ObtainTokensAsyncController` was not setting
   the `request.auser` attribute, #953
 - Fixed a bug that `JWTSyncAuth` was not setting `request.auser`, #953
+- Fixed `ResponseNegotiator` raising `NotAcceptableError` on streaming
+  endpoints when `Accept: text/event-stream` was sent without
+  `application/json` (the default browser `EventSource` case), which
+  made 4xx/5xx error bodies and response validation crash with a 500
+  instead of rendering the configured non-streaming default, #962
 
 ### Misc
 
--- a/dmr/negotiation.py
+++ b/dmr/negotiation.py
@@ -5,7 +5,11 @@ from typing import TYPE_CHECKING, Any, Final, Literal, final, overload
 from django.http.request import HttpRequest
 from django.utils.translation import gettext_lazy as _
 
-from dmr.exceptions import EndpointMetadataError, RequestSerializationError
+from dmr.exceptions import (
+    EndpointMetadataError,
+    NotAcceptableError,
+    RequestSerializationError,
+)
 from dmr.internal.negotiation import ConditionalType as _ConditionalType
 from dmr.internal.negotiation import media_by_precedence
 from dmr.internal.negotiation import negotiate_renderer as _negotiate_renderer
@@ -188,11 +192,21 @@ class ResponseNegotiator:
         )
         request.__dmr_renderer__ = renderer  # type: ignore[attr-defined]
         if self._streaming:
-            request.__dmr_nonstreaming_renderer__ = _negotiate_renderer(  # type: ignore[attr-defined]
-                request,
-                self._non_streaming_renderers,
-                default=self._non_streaming_default,
-            )
+            try:
+                non_streaming = _negotiate_renderer(
+                    request,
+                    self._non_streaming_renderers,
+                    default=self._non_streaming_default,
+                )
+            except NotAcceptableError:
+                # Main (streaming) negotiation already succeeded.
+                # A non-streaming renderer is only needed for 4xx/5xx
+                # error bodies and response validation — fall back to
+                # the configured default so those paths keep working
+                # for clients that only accept the streaming media
+                # type (e.g. browser ``EventSource``).
+                non_streaming = self._non_streaming_default
+            request.__dmr_nonstreaming_renderer__ = non_streaming  # type: ignore[attr-defined]
         return renderer
 
 
--- a/tests/test_unit/test_streaming/test_streaming_negotiation.py
+++ b/tests/test_unit/test_streaming/test_streaming_negotiation.py
@@ -3,8 +3,12 @@ from http import HTTPStatus
 from typing import Any
 
 import pytest
+from django.http import HttpResponse
 from typing_extensions import override
 
+from dmr import APIError, ResponseSpec, modify
+from dmr.errors import ErrorModel, format_error
+from dmr.exceptions import NotAcceptableError
 from dmr.negotiation import ContentType
 from dmr.plugins.pydantic import PydanticSerializer
 from dmr.renderers import JsonRenderer
@@ -13,7 +17,7 @@ from dmr.streaming import StreamingController, StreamingResponse
 from dmr.streaming.jsonl.renderer import JsonLinesRenderer
 from dmr.streaming.jsonl.validation import JsonLinesStreamingValidator
 from dmr.streaming.renderer import StreamingRenderer
-from dmr.streaming.sse import SSEvent
+from dmr.streaming.sse import SSEController, SSEvent
 from dmr.streaming.sse.renderer import SSERenderer
 from dmr.streaming.sse.validation import SSEStreamingValidator
 from dmr.test import DMRAsyncRequestFactory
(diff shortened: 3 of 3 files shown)


sobolevn (approved):
Great, thank you!

PR #965, opened 2026-04-23
Explicitly pin unit tests to django==5.2

Satevg:
Allow `Django 6.x` in unit test group

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,9 @@ dmr = "dmr_pytest"
 # This is a special group, it has special rules.
 # It is installed in `cibuildwheel` and must be minimal.
 unit-test = [
-  "django>=5.2,<6.1",
+  # Pin to Django 5.2 (LTS): it supports Python 3.11+, while Django 6.0
+  # requires Python 3.12+. CI only tests Django 5.2
+  "django~=5.2.0",
   "django-csp>=4.0,<5.0",
   "pytest>=9.0,<10",
   "pytest-cov>=7.0,<8",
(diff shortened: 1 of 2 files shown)


sobolevn (approved):

sobolevn:
Oh, this is not that simple. If we change this, we will run unit tests on different Django version depending on the Python version.

Satevg:
> Oh, this is not that simple. If we change this, we will run unit tests on different Django version depending on the Python version.

Yeah, that's true. 3.11 will use django=5.2.x Other 6.x

What's the problem with this? I don't quite get, since dmr officially support Django 6. Or you mean that we need a Django version matrix in CI?

sobolevn:
So, instead I propose to pin it to `~=5.2.0` and document why we use it for tests. Including different supported python versions for 5.2 and 6.0

sobolevn:
> What's the problem with this? I don't quite get, since dmr officially support Django 6. Or you mean that we need a Django version matrix in CI?

This can cause different CI flukes in my experience. It is better to have one version for all tests if possible, plus additional job for different django versions.

Satevg:
> So, instead I propose to pin it to `~=5.2.0` and document why we use it for tests. Including different supported python versions for 5.2 and 6.0

Ok, pinned version for unit tests. Now the comment will help to avoid confusion hopefully. As I was confused by main <6.1 dep and <6 in unit tests :smile:

pyproject.toml
```diff
@@ -86,7 +86,9 @@ dmr = "dmr_pytest"
 # This is a special group, it has special rules.
 # It is installed in `cibuildwheel` and must be minimal.
 unit-test = [
-  "django>=5.2,<6",
+  # Pin to Django 5.2 (LTS): it supports Python 3.11+, while Django 6.0
+  # requires Python 3.12+. CI only tests Django 5.2
```
```suggestion
  # requires Python 3.12+. CI runs different django versions as separate jobs
```

sobolevn (approved):
Thanks!

Issue #911, opened 2026-04-16
Improve `django-sessions` auth support


sobolevn:
There are multiple corner-cases that we miss right now:

- [x] Test different `settings.SESSION_COOKIE_NAME` for cookie views
- [x] Test `CSRF_USE_SESSIONS` and merge https://github.com/wemake-services/django-modern-rest/pull/674 
- [x] Ensure that `CSRF` cookie / session is correctly sent with our pre-existing `views.py`
- [x] Document that `CSRF` is sent
- [x] Add `schemathesis` auth for `django_session` openapi security scheme

Any help with any of these issues is appreciated.

sobolevn:
@Satevg are you interested in this task? :)
It is rather big one, with multiple subtasks.

Satevg:
Given that #966 merged and #967 is about to be merged soon hopefully (btw #674 can be closed), do I understand correctly, that the only item that remaining is:

> Add schemathesis auth for django_session openapi security scheme

@sobolevn ?

sobolevn:
Yes! Thanks a lot!

PR #967, opened 2026-04-24
feat: add CSRF cookie to django session login response schema

Satevg:
# I have made things!

  - Added CSRF cookie to the OpenAPI response schema for `DjangoSessionSyncController` and         
  `DjangoSessionAsyncController` login endpoints; omitted when `CSRF_USE_SESSIONS=True`
  - Added integration tests verifying the CSRF cookie is set on login and absent when            
  `CSRF_USE_SESSIONS=True`                                                                         
  - Added unit test asserting the schema correctly declares both cookies
  - Docs update                              
                                                    

## AI Policy

## Checklist

## Related issues

Refs #911

### DIFF
--- a/dmr/security/django_session/views.py
+++ b/dmr/security/django_session/views.py
@@ -66,7 +66,6 @@ class DjangoSessionSyncController(
                 else {
                     settings.CSRF_COOKIE_NAME: CookieSpec(
                         skip_validation=True,
-                        required=False,
                         description='CSRF protection.',
                     ),
                 }
@@ -138,7 +137,6 @@ class DjangoSessionAsyncController(
                 else {
                     settings.CSRF_COOKIE_NAME: CookieSpec(
                         skip_validation=True,
-                        required=False,
                         description='CSRF protection.',
                     ),
                 }
--- a/dmr/security/django_session/views.py
+++ b/dmr/security/django_session/views.py
@@ -60,6 +60,17 @@ class DjangoSessionSyncController(
         status_code=HTTPStatus.OK,
         cookies={
             settings.SESSION_COOKIE_NAME: CookieSpec(skip_validation=True),
+            **(
+                {}
+                if settings.CSRF_USE_SESSIONS
+                else {
+                    settings.CSRF_COOKIE_NAME: CookieSpec(
+                        skip_validation=True,
+                        required=False,
+                        description='CSRF protection.',
+                    ),
+                }
+            ),
         },
     )
     def post(self, parsed_body: Body[_RequestModelT]) -> _ResponseT:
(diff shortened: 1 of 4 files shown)


tests/test_integration/test_django_session_auth/test_django_sessions_auth_views.py
```diff
@@ -145,3 +145,58 @@ def test_wrong_auth_params(
...
+    url: str,
+) -> None:
+    """Ensures the login response sets the CSRF cookie and session cookie."""
+    response = dmr_client.post(
+        url,
+        data={'username': user.username, 'password': password},
+    )
+
+    assert response.status_code == HTTPStatus.OK, response.content
+    assert response.cookies[settings.SESSION_COOKIE_NAME]
```
sobolevn:
```suggestion
    assert response.cookies.keys() == {settings.SESSION_COOKIE_NAME, settings.CSRF_COOKIE_NAME}
```

tests/test_unit/test_security/test_django_session_auth.py
```diff
@@ -234,3 +242,67 @@ def test_schema_with_custom_cookie_names(
             security_scheme_in='cookie',
         ),
     })
+
+
+@final
+class _SyncLoginController(  # pragma: no cover
```
sobolevn: Why no cover? It is better to use `raise NotImplementedError` in specific methods that are not covered.

tests/test_unit/test_security/test_django_session_auth.py
```diff
@@ -234,3 +242,67 @@ def test_schema_with_custom_cookie_names(
...
+    'controller_cls',
+    [_SyncLoginController, _AsyncLoginController],
+)
+def test_login_schema_includes_csrf_cookie(
+    settings: LazySettings,
+    *,
+    controller_cls: type[_SyncLoginController] | type[_AsyncLoginController],
+) -> None:
+    """Ensures the 200 response schema declares session and CSRF cookies."""
+    settings.CSRF_USE_SESSIONS = False
```
sobolevn: This setting does not do much, because classes were already defined with the default setting.

sobolevn (commented):
Thanks!

dmr/security/django_session/views.py
```diff
@@ -121,6 +132,17 @@ class DjangoSessionAsyncController(
         status_code=HTTPStatus.OK,
         cookies={
             settings.SESSION_COOKIE_NAME: CookieSpec(skip_validation=True),
+            **(
+                {}
+                if settings.CSRF_USE_SESSIONS
+                else {
+                    settings.CSRF_COOKIE_NAME: CookieSpec(
+                        skip_validation=True,
```
sobolevn, line 139: question: can we please specify why this is not required? a comment maybe? I was under impression that it is done in `login` / `alogin` in `rotate_token` function, so it is kinda required. I might be wrong here :)

Satevg: oops, `required=False` left from the state where I didn't do the conditional check `if settings.CSRF_USE_SESSIONS` locally. Removed. Always set here for sure https://github.com/django/django/blob/d24160ab4855f6971aa1f437720adca6330048a2/django/middleware/csrf.py#L258

sobolevn (approved):

PR #969, opened 2026-04-25
Simplify RTD setup

sobolevn:
Refs https://docs.readthedocs.com/platform/stable/config-file/v2.html#uv


sobolevn:
`uv` is really bad, try to avoid it if possible :(

PR #976, opened 2026-04-26
Add `_parse_json` to `_DMRMixin`

kondratevdev:
# I have made things!

## AI Policy

## Checklist

### DIFF
--- a/dmr/internal/json.py
+++ b/dmr/internal/json.py
@@ -70,7 +70,7 @@ class NativeJson:
         ).encode('utf8')
 
     @classmethod
-    def loads(cls, to_deserialize: 'Raw', /) -> Any:
+    def loads(cls, to_deserialize: 'str | Raw', /) -> Any:
         """Internal method to load json as a simple object."""
         return json.loads(to_deserialize)
 
@@ -98,8 +98,10 @@ try:
     import msgspec
 except ImportError:  # pragma: no cover
     _json_dumps: Callable[[Any], str] = _compact_json_dumps
+    _json_loads: Callable[['str | Raw'], Any] = NativeJson.loads
 else:
     _json_dumps = _wrap_bytes_dumper(msgspec.json.encode)
+    _json_loads = msgspec.json.decode
 
 
 def json_dump(schema: Any) -> str:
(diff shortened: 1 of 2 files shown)


dmr/test.py
```diff
@@ -18,7 +19,16 @@ def _encode_json(self, data: Any, content_type: str) -> Any:
...
-    # TODO: add `def _parse_json` for `client`
+    def _parse_json(self, response: HttpResponse, **extra: Any) -> Any:
+        if not hasattr(response, '_json'):
+            if response.get('Content-Type') != self.default_content_type:
+                raise ValueError(  # pragma: no cover
+                    f'Content-Type header is "{response.get("Content-Type")}", '
+                    f'not "{self.default_content_type}"',
+                )
+
+            response._json = json_loads(response.text, **extra)  # type: ignore[attr-defined]  # noqa: SLF001
```
kondratevdev, line 41: It is important for us to pass `response.text`, not `response.content`, because this method expects text that has already been decoded with the correct `charset`. That is why I widened the accepted input type from `Raw` to `Raw | str`.

dmr/test.py
```diff
@@ -18,7 +19,16 @@ def _encode_json(self, data: Any, content_type: str) -> Any:
         )
         return json_dump(data) if should_encode else data
 
-    # TODO: add `def _parse_json` for `client`
+    def _parse_json(self, response: HttpResponse, **extra: Any) -> Any:
+        if not hasattr(response, '_json'):
```
sobolevn, line 34: 😱 

The worst piece of code in our code base.
But, we would have to keep it as-is.
Please, add a comment that it is taken from Django.

dmr/internal/json.py
```diff
@@ -114,3 +116,17 @@ def json_dump(schema: Any) -> str:
 
     """
     return _json_dumps(schema)
+
+
+def json_loads(raw: 'str | Raw') -> Any:
```
sobolevn: Do we really need this function? Can we just make `_json_loads` public?

kondratevdev: Same as `json_dump` function

sobolevn (approved):
Thanks!

PR #985, opened 2026-04-27
ci: add Valkey test job to CI matrix

### LINKED ISSUE #980: Test `Valkey` support
Currently we only test `redis` as the database for `SyncRedis` and `AsyncRedis` throttling backends. We also need to test Valkey, because we claim to support it.

nanookclaw:
Adds Valkey alongside Redis in CI to verify throttling backend compatibility.

**Changes:**
- : valkey/valkey:8-alpine service on port 6379
- : single matrix include (Python 3.13, valkey)

Scope per [maintainer guidance](https://github.com/wemake-services/django-modern-rest/issues/980#issuecomment-2793537085): single job, not full suite. Valkey is wire-compatible with Redis, so the same wait-for-it and test harness applies.

Closes #980

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,8 @@ jobs:
           - python-version: '3.14t'
             force_free_gil: true
             docker-compose-services: 'redis'
+          - python-version: '3.13'
+            docker-compose-services: 'valkey'
 
     steps:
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Looks great! Thanks!

PR #998, opened 2026-05-03
Add allow_unsafe_throttle_cache setting for cache backend safety checks #998

### LINKED ISSUE #978: Warn unsafe cache deployments
When `settings.DEBUG` is `False`, we should warn when throttling is used with `DjangoCache` and `LocMemory` or `DummyCache` backends.

Maybe not just warn, but raise an error? And only warn in development?

It is not safe.

n05tr0m0:
## AI Policy

## Checklist

## Related issues
Closes #978

`LocMemCache` and `DummyCache` don't share state between processes, so using them for throttling silently breaks rate limiting in production. This PR makes that configuration fail loudly.

By default, an `ImproperlyConfigured` exception is raised when an unsafe backend is detected. Setting `throttle_allow_unsafe_cache = True` in `DMR_SETTINGS` opts in to unsafe behavior – in that case an `UnsafeCacheBackendWarning` is emitted so at least you know it's on.

The check runs once in `__post_init__` when the cache backend is constructed, so there's no overhead on every request.


dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,58 @@
+import warnings
+
+from django.core.cache import caches
+from django.core.exceptions import ImproperlyConfigured
+
+_UNSAFE_CACHE_BACKENDS = frozenset((
```
sobolevn:
```suggestion
_UNSAFE_CACHE_BACKENDS: Final = frozenset((
```

dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,58 @@
...
+    'django.core.cache.backends.locmem.LocMemCache',
+    'django.core.cache.backends.dummy.DummyCache',
+))
+
+_WARNING_MSG = (
```
sobolevn:
```suggestion
_WARNING_MSG: Final = (
```

dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,58 @@
...
+    "Throttling is using '{backend}' cache backend which is not safe for "
+    'production: counters are NOT shared between processes/instances. '
+    'Use Redis or Memcached instead.'
+)
+
+_ERROR_MSG = (
```
sobolevn:
```suggestion
_ERROR_MSG: Final = (
```

dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,58 @@
...
+    _WARNING_MSG
+    + ' To suppress this error and run at your own risk, set '
+    + "'allow_unsafe_throttle_cache': True in DMR_SETTINGS."
+)
+
+# Tracks which cache names have already been checked to avoid redundant
+# checks on every throttling operation:
+_CHECKED_CACHES: set[str] = set()
```
sobolevn:
```suggestion
_CHECKED_CACHES: Final[set[str]] = set()
```

dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def is_cache_checked(cache_name: str) -> bool:
+    """Return whether the given cache name has already been checked."""
+    return cache_name in _CHECKED_CACHES
+
+
+def clear_safety_checks() -> None:
```
sobolevn: I would prefer not to have this mutable module state. Just check all items always, we do it on import time, it is safe to execute slow code there.

dmr/throttling/backends/django_cache.py
```diff
@@ -77,6 +87,7 @@ def incr(
         cache_key: str,
         algorithm: 'BaseThrottleAlgorithm',
     ) -> CachedRateLimit:
+        self._ensure_safety()
```
sobolevn: Nope, this should be called once on `__init__`. We can't check this on each request, it is way too slow.

docs/pages/throttling.rst
```diff
@@ -164,6 +164,22 @@ Full list of backends that we ship in ``django-modern-rest``:
 
   Some like ``django.core.cache.backends.dummy.DummyCache`` do nothing at all.
 
+  By default, **django-modern-rest** raises
+  :exc:`~django.core.exceptions.ImproperlyConfigured`
+  when detecting an unsafe cache backend for throttling.
+  To suppress this check and run at your own risk,
+  set ``allow_unsafe_throttle_cache`` to ``True`` in
```
sobolevn:
```suggestion
  set :data:`dmr.settings.Settings.allow_unsafe_throttle_cache` to ``True`` in
```

Don't forget to document this value in `configuration.rst` near `dmr.settings.Settings.throttling`

tests/test_unit/test_throttling/test_unsafe_cache_warning.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _reset_settings_cache() -> Generator[None, None, None]:
+    """Reset DMR settings cache and safety checks before/after each test."""
+    clear_settings_cache()
+    clear_safety_checks()
+    yield
+    clear_settings_cache()
+    clear_safety_checks()
```
sobolevn:
```suggestion
@pytest.fixture(autouse=True)
def _reset_settings_cache(
     dmr_clean_settings: None,
) -> None:
    """Reset DMR settings cache."""
```

dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,58 @@
...
+        _CHECKED_CACHES.add(cache_name)
+        return
+
+    from dmr.settings import Settings, resolve_setting  # noqa: PLC0415
+
+    allow_unsafe = resolve_setting(Settings.allow_unsafe_throttle_cache)
+    _CHECKED_CACHES.add(cache_name)
+    if allow_unsafe:
+        warnings.warn(  # noqa: B028
+            _WARNING_MSG.format(backend=backend),
```
sobolevn: Please, define a warning class for this, let's call it `UnsafeCacheBackendWarning`

sobolevn (commented):
Great work! Thanks a lot!

dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,46 @@
+import warnings
+from typing import Final
+
+from django.core.cache import caches
+from django.core.exceptions import ImproperlyConfigured
+
+
+class UnsafeCacheBackendWarning(UserWarning):
```
sobolevn:
```suggestion
@final
class UnsafeCacheBackendWarning(UserWarning):
```

Let's move this logic to `django_cache.py`, since it is doing the same thing :)

dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,46 @@
...
+)
+
+_ERROR_MSG: Final = (
+    _WARNING_MSG
+    + ' To suppress this error and run at your own risk, set '
+    + "'allow_unsafe_throttle_cache': True in DMR_SETTINGS."
+)
+
+
+def check_throttle_cache_safety(cache_name: str) -> None:
```
sobolevn:
```suggestion
def _check_throttle_cache_safety(cache_name: str) -> None:
```

this would become protected function in this case.

dmr/throttling/backends/_cache_safety.py
```diff
@@ -0,0 +1,46 @@
...
+    cache = caches[cache_name]
+    backend = f'{type(cache).__module__}.{type(cache).__qualname__}'
+
+    if backend not in _UNSAFE_CACHE_BACKENDS:
+        return
+
+    from dmr.settings import Settings, resolve_setting  # noqa: PLC0415
+
+    allow_unsafe = resolve_setting(Settings.allow_unsafe_throttle_cache)
```
sobolevn:
```suggestion
    allow_unsafe = resolve_setting(Settings.throttle_allow_unsafe_cache)
```

It is better to have common settings prefixes for common things.

tests/test_unit/test_throttling/test_unsafe_cache_warning.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@override_settings(CACHES=LOCMEM_CACHES, DMR_SETTINGS={})
+def test_raises_by_default_with_locmem() -> None:
+    """ImproperlyConfigured is raised for LocMemCache by default."""
+    with pytest.raises(
+        ImproperlyConfigured,
+        match='allow_unsafe_throttle_cache',
+    ):
+        check_throttle_cache_safety('default')
```
sobolevn: Let's test `DjangoCache` backend object creation here, since it is the public API, not this internal helper :)

sobolevn (commented):
One final round of review :)

pyproject.toml
```diff
@@ -336,7 +337,10 @@ isort.known-first-party = ["server", "examples"]
 # Strict `@xfail` by default:
 xfail_strict = true
 # Fail on warnings:
-filterwarnings = ["error"]
+filterwarnings = [
+    "error",
+    "always::dmr.throttling.backends._cache_safety.UnsafeCacheBackendWarning",
```
sobolevn: We can't do that: instead we should we should use `pytest.assertWarns` or `warnings.filter` for the specific warning places :)

This will not work, because we don't controll where warning can happen, we can introduce regressions in the future.

pyproject.toml
```diff
@@ -281,6 +281,7 @@ select = [
 ]
 ignore = [
   "A005",   # allow to shadow stdlib and builtin module names
+  "COM812", # trailing commas — conflicts with ruff formatter
```
sobolevn: This is fine, you can remove it. `ruff` just does not know what it is doing :(

tests/test_unit/test_throttling/test_unsafe_cache_warning.py
```diff
@@ -0,0 +1,96 @@
...
+)
+def test_no_warning_for_safe_backend() -> None:
+    """No warning or error for safe backends like Redis."""
+    SyncDjangoCache()
+
+
+@override_settings(CACHES=REDIS_CACHES, DMR_SETTINGS={})
+def test_no_error_for_safe_backend_default() -> None:
+    """No error for safe backends with default settings."""
+    SyncDjangoCache()
```
sobolevn: You can use a parametrized fixture to get a needed class (sync vs async) to test all backend versions in all tests here.

sobolevn (changes requested):

sobolevn (approved):
Thanks! I will take it from here ;)

PR #1000, opened 2026-05-04
replace Empty with sentinel EMPTY

### LINKED ISSUE #995: Convert our `Empty` class usage to `typing_extensions.Sentinel`
Current way: https://github.com/wemake-services/django-modern-rest/blob/5ca7975fb54112432769fe0b267773409ba78201/dmr/types.py#L48-L55

New (modern) way: https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.Sentinel

1. We need to change how `EmptyObj` is defined to be `EMPTY: Final = Sentinel('EMPTY')`
2. We need to change all `isinstance(..., Empty)` to check for the sentinel

This is a very easy task for new contributors :)

Nalsher:
# I have made things!
# Closes #995


dmr/types.py
sobolevn: We don't need this class anymore, you can remove it.

Nalsher: Done

sobolevn:
Sorry, but https://github.com/wemake-services/django-modern-rest/pull/1024 was first.

Issue #1020, opened 2026-05-15
Can't validate file responses without `as_attachment` in FileResponse


PashaWNN:
### What's wrong?

```
class PingController(Controller[MsgspecSerializer]):
    @validate(
        FileResponseSpec(),
        renderers=[FileRenderer('text/plain')],
    )
    def get(self) -> FileResponse:
            return FileResponse(io.BytesIO(b'Hello'), content_type='text/plain')
```

This code fails with server error due to  `FileRenderer.render() must not be called, instead return a FileResponse directly`
If I add `as_attachment` to FileResponse args, it works ok

### How it should be?

It should return "Hello" without requiring `as_attachment` in the FileResponse args

### Used versions

0.9.0

### OS information

Docker Python 3.14-slim

sobolevn:
We would also need a doc example :)

PR #1022, opened 2026-05-15
fix `Content-Disposition`, refs #1020

milssky:
# I have made things!

## AI Policy

## Checklist

## Related issues

Refs #1020

### DIFF
--- a/dmr/files.py
+++ b/dmr/files.py
@@ -118,7 +118,7 @@ class FileResponseSpec(ResponseSpec):
         kw_only=True,
         default_factory=lambda: {
             'Content-Length': HeaderSpec(),
-            'Content-Disposition': HeaderSpec(),
+            'Content-Disposition': HeaderSpec(required=False),
         },
     )
     file_body: type[FileBody] = dataclasses.field(
--- a/dmr/files.py
+++ b/dmr/files.py
@@ -26,19 +26,19 @@ if TYPE_CHECKING:
     from dmr.serializer import BaseSerializer
 
 
-def _response_headers(
+def file_response_headers(
     headers: Mapping[str, HeaderSpec] | None,
     *,
     as_attachment: bool,
-) -> Mapping[str, HeaderSpec] | None:
+) -> Mapping[str, HeaderSpec]:
     """Build headers expected from ``FileResponse``."""
-    if not as_attachment:
-        return headers
-
-    return {
+    response_headers = {
+        'Content-Length': HeaderSpec(),
         **(headers or {}),
-        'Content-Disposition': HeaderSpec(),
     }
+    if as_attachment:
+        response_headers['Content-Disposition'] = HeaderSpec()
+    return response_headers
 
 
 @dataclasses.dataclass(slots=True, frozen=True)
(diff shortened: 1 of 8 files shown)


dmr/files.py
```diff
@@ -118,7 +118,7 @@ class FileResponseSpec(ResponseSpec):
         kw_only=True,
         default_factory=lambda: {
             'Content-Length': HeaderSpec(),
-            'Content-Disposition': HeaderSpec(),
+            'Content-Disposition': HeaderSpec(required=False),
```
sobolevn: I don't quite like this :( 

It must be present for `as_attachment=True`, but it must not be present when `as_attachment=False`.

Let's maybe put this logic into `__post_init__` and add `as_attachment: bool = False` to `FileResponseSpec`?

This way we can control this easily.

docs/examples/using_controller/inline_file_response.py
```diff
@@ -0,0 +1,23 @@
...
+        renderers=[FileRenderer('text/plain')],
+    )
+    def get(self) -> FileResponse:
+        return FileResponse(
+            io.BytesIO(b'Hello'),
+            content_type='text/plain',
+        )
+
+
+# run: {"controller": "InlineFileController", "method": "get", "url": "/api/file/", "curl_args": ["-D", "-"]}  # noqa: ERA001, E501
```
sobolevn, line 23: Don't forget to show `# openapi:`

dmr/files.py
```diff
@@ -26,6 +26,21 @@
     from dmr.serializer import BaseSerializer
 
 
+def _response_headers(
```
sobolevn:
```suggestion
def file_response_headers(
```

It might be useful for end users :)

dmr/files.py
```diff
@@ -26,6 +26,21 @@
...
+    headers: Mapping[str, HeaderSpec] | None,
+    *,
+    as_attachment: bool,
+) -> Mapping[str, HeaderSpec] | None:
+    """Build headers expected from ``FileResponse``."""
+    if not as_attachment:
```
sobolevn: Let's define all headers here now, not in `default_factory`

dmr/files.py
```diff
@@ -118,14 +135,27 @@ class FileResponseSpec(ResponseSpec):
         kw_only=True,
         default_factory=lambda: {
```
sobolevn:
```suggestion
        default=None,
```

docs/pages/using-controller/files.rst
```diff
@@ -14,14 +14,26 @@ We support file and other binary responses.
...
+Set ``as_attachment=True`` when Django's ``FileResponse`` is returned as
+an attachment:
 
 .. literalinclude:: /examples/using_controller/file_response.py
   :caption: views.py
   :language: python
   :linenos:
   :emphasize-lines: 16, 17
 
+If you don't pass ``filename`` or ``as_attachment=True`` to Django's
```
sobolevn: This should be the first example, because it is the default one. `as_attachment` is an optional parameter, so it should be the second example.

Please, also explain what is the difference, provide a link to the further reading about `Content-Disposition` HTTP header.

tests/test_unit/test_openapi/__snapshots__/test_files_snapshots.ambr
```diff
@@ -355,12 +355,6 @@
                     "type": "string"
                   },
                   "required": true
-                },
```
sobolevn, line 358: Please, test both now :)

sobolevn (commented):
Things left to do:
1. Document `file_response_headers` with `.. autofunction`
2. Add `CHANGELOG.md` entry, this is a breaking change

sobolevn (approved):

PR #1044, opened 2026-05-27
Fix leaky bucket partial-drain rejection

### LINKED ISSUE #1043: Flaky test: `test_redis_sync_leaky_bucket`
Output:

```
 ______________________ test_redis_sync_leaky_bucket[True] ______________________
tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py:110: in test_redis_sync_leaky_bucket
    assert response.status_code == HTTPStatus.TOO_MANY_REQUESTS
E   assert 200 == <HTTPStatus.TOO_MANY_REQUESTS: 429>
E    +  where 200 = <HttpResponse status_code=200, "application/json">.status_code
E    +  and   <HTTPStatus.TOO_MANY_REQUESTS: 429> = HTTPStatus.TOO_MANY_REQUESTS
================================ tests coverage ================================
```

```
=========================== short test summary info ============================
FAILED tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py::test_redis_sync_leaky_bucket[True] - assert 200 == <HTTPStatus.TOO_MANY_REQUESTS: 429>
 +  where 200 = <HttpResponse status_code=200, "application/json">.status_code
 +  and   <HTTPStatus.TOO_MANY_REQUESTS: 429> = HTTPStatus.TOO_MANY_REQUESTS
============ 1 failed, 2479 passed, 10 skipped in 243.50s (0:04:03) ============

```

It fails from time to time: https://github.com/wemake-services/django-modern-rest/actions/runs/26490639540/job/78007463856?pr=1042

iccccccccccccc:
﻿Fixes #1043.

This was a small boundary bug in the leaky bucket check: after the bucket was full, a short drain could make the stored level lower than capacity, so the next request could sneak through even though adding a full request would overflow the bucket.

This changes the check to require enough free capacity for the whole request before accepting it, and keeps the Redis Lua script aligned with the Python implementation.

Tests run:

- uv run --all-extras --group unit-test pytest tests/test_unit/test_throttling/test_algorithms/test_leaky_bucket.py -q -o addopts=""
- uv run --all-extras --group unit-test --group dev-and-docs pytest tests/test_unit/test_throttling/test_algorithms/test_leaky_bucket.py tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py -q -o addopts=""
- uv run --all-extras --group dev ruff check dmr/throttling/algorithms.py dmr/throttling/lua.py tests/test_unit/test_throttling/test_algorithms/test_leaky_bucket.py tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py
- uv run --all-extras --group dev ruff format --check dmr/throttling/algorithms.py dmr/throttling/lua.py tests/test_unit/test_throttling/test_algorithms/test_leaky_bucket.py tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py

### DIFF
--- a/dmr/throttling/algorithms.py
+++ b/dmr/throttling/algorithms.py
@@ -202,7 +202,7 @@ class LeakyBucket(BaseThrottleAlgorithm):
         )
         cache_object = CachedRateLimit(history=[level], time=now)
 
-        if cache_object['history'][0] >= (
+        if cache_object['history'][0] + throttle.duration_in_seconds > (
             throttle.max_requests * throttle.duration_in_seconds
         ):
             raise TooManyRequestsError(
--- a/tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py
+++ b/tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py
@@ -103,6 +103,11 @@ def test_redis_sync_leaky_bucket(
         assert isinstance(response, HttpResponse)
         assert response.status_code == HTTPStatus.OK
 
+    keys = list(redis_client.scan_iter())
+    assert len(keys) == 1
+    # Simulate one elapsed Redis-server second without sleeping.
+    redis_client.hincrby(keys[0], 'last_time', -1)
+
     # Third is rejected:
     request = dmr_rf.get('/whatever/')
     response = _SyncController.as_view()(request)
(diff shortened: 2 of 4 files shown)


tests/test_unit/test_throttling/test_algorithms/test_leaky_bucket.py
```diff
@@ -95,6 +95,34 @@ def test_leaky_bucket_smooth_drain(
     assert response.status_code == HTTPStatus.TOO_MANY_REQUESTS
 
 
+def test_leaky_bucket_waits_for_full_slot(
```
sobolevn, line 98: This test only covers `python` implementation, not `lua` one. Can you please duplicate it in `tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py`?

tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py
```diff
@@ -103,6 +103,10 @@ def get(self) -> str:
         assert isinstance(response, HttpResponse)
         assert response.status_code == HTTPStatus.OK
 
+    keys = list(redis_client.scan_iter())
```
sobolevn, line 106: Why do we need this? :)

dmr/throttling/algorithms.py
```diff
@@ -202,7 +202,7 @@ def access(
         )
         cache_object = CachedRateLimit(history=[level], time=now)
 
-        if cache_object['history'][0] >= (
+        if cache_object['history'][0] + throttle.duration_in_seconds > (
```
sobolevn, line 205: Thanks! Good catch, indeed this was undertested :(
Should it be > or >=? Can we test exactly this corner case?
A comment in the test would be an amazing to have, that we explicitly check for `>` vs `>=`.

sobolevn (commented):
Thanks!

iccccccccccccc:
﻿Addressed, thanks for the pointers.

I added the Redis/Lua coverage for the same partial-drain case, left comments explaining the manual Redis timestamp shift, and made the `>` vs `>=` boundary explicit in both tests: exactly enough drained capacity should allow the next request.

Reran:
- `.\.venv\Scripts\python.exe -m pytest tests/test_unit/test_throttling/test_algorithms/test_leaky_bucket.py -q -o addopts=""`
- `.\.venv\Scripts\python.exe -m pytest tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py -q -o addopts=""`
- `git diff --check`

iccccccccccccc:
CI caught a style issue in the Redis test name (`WPS118`). I shortened that test name and force-pushed the same commit.

Reran locally:
- `.\.venv\Scripts\python.exe -m flake8 tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py`
- `.\.venv\Scripts\python.exe -m pytest tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py -q -o addopts=""` with a local Redis container
- `.\.venv\Scripts\python.exe -m pytest tests/test_unit/test_throttling/test_algorithms/test_leaky_bucket.py -q -o addopts=""`
- `git diff --check`

sobolevn (approved):

tests/test_integration/test_throttling/test_backends/test_redis_backend/test_redis_backend.py
iccccccccccccc: It is there to simulate one elapsed Redis-server second without sleeping. The Lua backend uses Redis server time, so moving `last_time` back makes the partial-drain state deterministic and keeps the test fast. I added the comment around the timestamp shift in the follow-up commit.

tests/test_unit/test_throttling/test_algorithms/test_leaky_bucket.py
iccccccccccccc: Yep, added the Redis/Lua version in the follow-up commit. It covers the same partial-drain case through the Redis backend, so the pure Python path and the Lua path both have regression coverage now.

dmr/throttling/algorithms.py
iccccccccccccc: I kept the runtime check strict here. The added boundary test covers the equality case explicitly: once exactly enough capacity has drained back, the request should be allowed, so we only reject when the bucket is still over capacity after accounting for the drain.

PR #1051, opened 2026-05-31
Add opaque token authentication with header, query, and cookie backends

### LINKED ISSUE #1032: Add token auth scheme
# Feature

## Thesis

DMR ships JWT auth but has no long-lived user-bound opaque token that authenticates without the access/refresh cycle. I'd like to propose adding one at `dmr.security.token`, alongside the existing JWT scheme.

Rough sketch of issuance:

```python
from dmr.security.token.models import Token

token, raw = Token.objects.create_token(user=request.user, name="CI pipeline")
# `raw` is shown once at creation
```

```http
GET /api/resource/ HTTP/1.1
Authorization: Token <raw_token>
```

Properties:

- Opaque (random string, no payload)
- Optional expiry that (maybe?) never-expires by default, configurable per token
- Named tokens (`"CI pipeline"`, `"Mobile app v2"`) so they're auditable and individually revocable
- Per-token revocation that doesn't touch the user's other tokens (an improvement over DRF's one-token-per-user default)
- Stored hashed, so the raw value is only ever returned at creation time

## Reasoning

JWTs are great for stateless, short-lived sessions, but they're an awkward fit for long-lived machine-to-machine credentials. For example, a CI job or embedded device doesn't want to implement refresh logic, and there's no clean per-credential revocation.

Worth calling out that I recognize this would introduce a DB lookup mechanism to the project, which brings it's own ... stuff.

If there's interest, I'm happy to implement it as I'm building exactly this in my own project, so I'd be porting work I'm doing anyway.

Dresdn:
# I have made things!

## AI Policy

## Checklist

## Related issues

Closes #1032


dmr/validation/settings.py
```diff
@@ -80,12 +82,19 @@ def _validate_structure(self) -> _SettingsModel:
...
 
-    def _validate_types(  # noqa: C901, WPS231, WPS238
+    def _validate_types(
         self,
         settings: _SettingsModel,
     ) -> None:
         # Some types are not compatible with pydantic / msgspec validation.
         # So, we validate them by hands.
+        self._validate_sequence_types(settings)
+        self._validate_scalar_types(settings)
```
Dresdn, line 92: @sobolevn refactored this since adding a check for the new `token_default_expiry` settings pushed complexity over the edge to:

```
dmr/validation/settings.py
  0:1      WPS232 Found module cognitive complexity that is too high: 8.3 > 8
```

dmr/security/token/auth/cookie.py
```diff
@@ -0,0 +1,99 @@
+from django.http import HttpRequest
+from typing_extensions import override
+
+from dmr.openapi.objects import Reference, SecurityScheme
+from dmr.security.token.auth.header import TokenAsyncAuth, TokenSyncAuth
+
+_DEFAULT_PARAM = 'token'
```
sobolevn:
```suggestion
_DEFAULT_PARAM: Final = 'token'
```

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,319 @@
...
+        - *cookie_name* — which cookie carries the raw token (default
+          ``token``).
+        - *security_scheme_name* — name used in OpenAPI security scheme map.
+
+        **Common configurations:**
+
+        .. code-block:: python
+
+            # Default — custom header, no prefix
+            TokenSyncAuth()  # X-API-Token: <token>
```
sobolevn: Please, use doctests

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,319 @@
...
+
+            # DRF-compatible
+            TokenSyncAuth(header_name='Authorization', prefix='Token')
```
sobolevn:
```suggestion
            >>> TokenSyncAuth(header_name='Authorization', prefix='Token')
```

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,319 @@
...
+if TYPE_CHECKING:
+    from django.contrib.auth.base_user import AbstractBaseUser
+
+    from dmr.controller import Controller
+    from dmr.endpoint import Endpoint
+    from dmr.security.token.models import Token
+    from dmr.serializer import BaseSerializer
+
+
+class _BaseTokenAuth:
```
sobolevn: I would move this to `token/auth/base.py`

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,319 @@
...
+        token = self.get_token(raw_token)
+        self.check_user(token.user)
+        if self.update_last_used:
+            self.mark_token_used(token)
+        self.set_request_attrs(request, token.user, token=token)
+        return token.user
+
+    def get_token(self, raw_token: str) -> 'Token':
+        """Look up and validate the token from the DB."""
+        model = _token_model()
```
sobolevn: Let's make `_token_model` a method, not a function. This way users will be able to customize it.

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,319 @@
...
+        token_hash = model.objects.hash_token(raw_token)
+        try:
+            token = model.objects.select_related('user').get(
+                token_hash=token_hash,
+            )
+        except ObjectDoesNotExist:
+            raise NotAuthenticatedError from None
+        if not token.is_active:
```
sobolevn: This should go into `check_token` method

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,319 @@
...
+    def authenticate(
+        self,
+        request: HttpRequest,
+        raw_token: str,
+    ) -> 'AbstractBaseUser':
+        """Run all auth pipeline."""
+        token = self.get_token(raw_token)
+        self.check_user(token.user)
+        if self.update_last_used:
+            self.mark_token_used(token)
```
sobolevn: This is not atomic :(

dmr/security/token/auth/cookie.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class CookieTokenSyncAuth(TokenSyncAuth):
+    """
+    Sync opaque token auth reading from a cookie.
+
+    .. warning::
+        Cookie-based authentication is vulnerable to CSRF attacks in
```
sobolevn, line 34: Maybe we should just force CSRF by default? Like we do in Session-based auth?

dmr/security/token/auth/query.py
```diff
@@ -0,0 +1,97 @@
+from django.http import HttpRequest
+from typing_extensions import override
+
+from dmr.openapi.objects import Reference, SecurityScheme
+from dmr.security.token.auth.header import TokenAsyncAuth, TokenSyncAuth
+
+_DEFAULT_PARAM = 'token'
```
sobolevn:
```suggestion
_DEFAULT_PARAM: Final = 'token'
```

dmr/security/token/models.py
```diff
@@ -0,0 +1,185 @@
...
+        expires_at: dt.datetime | Sentinel | None = _UNSET,
+    ) -> 'tuple[Token, str]':
+        """
+        Create a new token, returning ``(Token instance, raw token string)``.
+
+        If *expires_at* is not provided the token expires after
+        ``DMR_SETTINGS[Settings.token_default_expiry]``.
+        Pass ``None`` explicitly to create a non-expiring token.
+        """
+        if expires_at is _UNSET:
```
sobolevn: Let's refactor this into `get_expires_at` method

dmr/security/token/models.py
```diff
@@ -0,0 +1,185 @@
...
+        null=True,
+        blank=True,
+        verbose_name=_('Expires at'),
+    )
+    revoked_at: _DateTimeFieldNullable = models.DateTimeField(
+        null=True,
+        blank=True,
+        verbose_name=_('Revoked at'),
+    )
+    last_used_at: _DateTimeFieldNullable = models.DateTimeField(
```
sobolevn, line 49: Please add `assert_type` tests for this model: it should reveal correct types in all type-checkers for all model fields.

dmr/security/token/models.py
```diff
@@ -0,0 +1,185 @@
...
+from django.contrib.auth.base_user import AbstractBaseUser
+from django.db import models
+from django.db.models.expressions import Combinable
+from django.utils.crypto import salted_hmac
+from django.utils.translation import gettext_lazy as _
+from typing_extensions import Sentinel, override
+
+from dmr.settings import Settings, resolve_setting
+
+_UNSET: Final = Sentinel('_UNSET')
```
sobolevn: We already have this value in `dmr.types`, please use it

dmr/security/token/admin.py
```diff
@@ -0,0 +1,31 @@
...
+from django.contrib import admin
+
+from dmr.security.token.models import Token
+
+
+@admin.register(Token)
+class TokenAdmin(admin.ModelAdmin):  # type: ignore[type-arg]
+    """Admin configuration for opaque auth tokens."""
+
+    list_display = (
```
sobolevn, line 34: don't forget to add `user` to `list_select_related`

dmr/security/token/admin.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+from django.contrib import admin
+
+from dmr.security.token.models import Token
+
+
+@admin.register(Token)
+class TokenAdmin(admin.ModelAdmin):  # type: ignore[type-arg]
```
sobolevn:
```suggestion
class TokenAdmin(admin.ModelAdmin[Token]):
```

If you fear that this class is not generic in runtime, you can use 

```python
if TYPE_CHECKING:
    ModelAdmin = admin.ModelAdmin
else:
    class ModelAdmin(admin.ModelAdmin, Generic[_ModelT]): ...
```

tests/test_unit/test_security/test_token/test_token_models.py
```diff
@@ -0,0 +1,196 @@
+import datetime as dt
+
+import pytest
+from django.conf import LazySettings
+from django.contrib.auth.models import User
+
+from dmr.security.token.constants import TOKEN_DEFAULT_EXPIRY_DAYS
+from dmr.security.token.models import Token
+
+_CUSTOM_EXPIRY_DAYS = 90
```
sobolevn:
```suggestion
_CUSTOM_EXPIRY_DAYS: Final = 90
```

docs/pages/auth/common.rst
```diff
@@ -149,6 +149,12 @@ Select auth backend that fits your needs:
 
       Support for JWT tokens based auth.
 
+    .. grid-item-card:: Opaque Tokens
+      :link: token
+      :link-type: doc
+
+      Database-backed opaque token auth with revocation support.
```
sobolevn, line 181: Please, describe when to choose JWT vs Token. You can create a new table with important points just below.

docs/examples/auth/token/using_token_query.py
```diff
@@ -0,0 +1,29 @@
...
+    auth = (QueryTokenSyncAuth(),)
+
+    def get(self) -> str:
+        # Let's test that `User` has the correct type:
+        assert self.request.user.username
+        return 'authed'
+
+
+# Custom query parameter name (default is `?token=`):
+class CustomQueryParamController(Controller[PydanticSerializer]):
```
sobolevn: Please, split this example into its own file. One class - one example.

docs/examples/auth/token/using_token_header.py
```diff
@@ -0,0 +1,29 @@
...
+    auth = (TokenSyncAuth(),)
+
+    def get(self) -> str:
+        # Let's test that `User` has the correct type:
+        assert self.request.user.username
+        return 'authed'
+
+
+# Custom header name (default is `X-API-Token`):
+class CustomHeaderController(Controller[PydanticSerializer]):
```
sobolevn: and here

docs/examples/auth/token/token_lifecycle.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+        return TokenOut(
+            id=token.pk,
+            name=token.name,
+            expires_at=token.expires_at,
+            raw_token=raw_token,
+        )
+
+
+class RevokeTokenController(Controller[PydanticSerializer]):
```
sobolevn: and here

dmr/settings.py
```diff
@@ -158,6 +162,8 @@ class SettingsDict(TypedDict, total=False):
     Settings.global_error_handler: 'dmr.errors.global_error_handler',
     # Settings for middleware:
     Settings.django_treat_as_post: frozenset(('PUT', 'PATCH')),
+    # Token auth settings:
+    Settings.token_default_expiry: dt.timedelta(days=TOKEN_DEFAULT_EXPIRY_DAYS),
```
sobolevn: Do we really need this global setting? I would prefer to make this configurable in one place only: in auth definition.

sobolevn (commented):
Great stuff! This is the first round of review :)

dmr/security/token/auth/header.py
Dresdn: I'll update everything to doctests

dmr/security/token/auth/header.py
Dresdn: I went back and forth on this, but ultimately settled on this for a few reasons:
1. Proximity - `TokenSyncAuth()` is the really the base class to extend from should want to use a different method
2. Putting it in `token/auth/base.py` would mean it'd need to become public as `token/auth/header.py` would need to import it, otherwise we'd have `# noqa:` for private imports. With it then being public, would give the impression it's part of the API when really, it's not.
3. The same single-file context exists in `jwt/auth.py` which is why I ultimately settled on this pattern (since both `cookie.py` and `query.py` never import it

Thoughts?

dmr/security/token/auth/header.py
Dresdn: Good point

dmr/security/token/auth/header.py
Dresdn: Good call here too. I'll add a `self.check_token(token)` in `authenticate()` to allow for extending additional checks (like if someone wants a token to be valid on odd days or whatever)

dmr/security/token/auth/header.py
Dresdn: Oh this is a good catch. Let me go through the process as there might be some conditions needing to be addressed here

dmr/security/token/auth/query.py
Dresdn: So I should make constants `Final`?? :)

dmr/security/token/auth/cookie.py
Dresdn: Yes, I totally missed looking at how session based auth was implemented. Doing a quick look, thoughts on something like a `security/csrf.py` or `_csrf.py` module?

dmr/settings.py
Dresdn: I went back and forth on this, but ultimately settled on this because `TokenManager.create_token()` is a model-layer operation which can be called from management commands, the admin page, scripts, cli, etc., all of which have no relationship to the auth definition. Without a default, all those calls could accidentally create non-expiring tokens, which is a bad "default" imho. Thinking how I'd implement this in my own project, I'd want a global setting to ensure that by default, tokens won't be valid over X days.

dmr/security/token/auth/header.py
sobolevn: It is ok to import protected stuff for our own code.

sobolevn (commented):
One more thing: I would prefer to have a plain model with 0 methods / properties. 

I really like this design: https://github.com/wemake-services/django-modern-rest/blob/master/dmr/security/jwt/blocklist/auth.py

We would just have an auth class with some methods to `create_token`, `revoke_token`, etc.

We can possibly have a manager to simplify the token creation (?), but I would prefer to just have a method for this.

This way models can stay simple and thin. All the business logic will be handled by business entities, not Django.

dmr/security/token/models.py
```diff
@@ -46,12 +44,26 @@ def hash_token(self, raw_token: str) -> str:
...
+    ) -> dt.datetime | None:
+        """Resolve the expiry datetime, falling back to the global default."""
+        if expires_at is EMPTY:
+            default_expiry: dt.timedelta | None = resolve_setting(
+                Settings.token_default_expiry,
+            )
+            if default_expiry is None:
+                return None
+            return dt.datetime.now(dt.UTC) + default_expiry
+        return cast(dt.datetime | None, expires_at)
```
sobolevn: Please, remove the `cast`, just use a type ignore comment. This is a bug in type-checkers for now, because `sentinel` is not yet fully supported.

dmr/security/token/auth/base.py
```diff
@@ -0,0 +1,135 @@
...
+            # DRF-compatible
+            >>> auth = TokenSyncAuth(
+            ...     header_name='Authorization',
+            ...     prefix='Token'
+            ... )
+
+            # Bearer style
+            >>> auth = TokenSyncAuth(
+            ...     header_name='Authorization',
+            ...     prefix='Bearer'
```
sobolevn:
```suggestion
            ...     prefix='Bearer',
```

style nit

dmr/security/token/auth/base.py
```diff
@@ -0,0 +1,135 @@
...
+
+            >>> from dmr.security.token.auth.header import TokenSyncAuth
+
+            # Default — custom header, no prefix
+            >>> auth = TokenSyncAuth()  # X-API-Token: <token>
+
+            # DRF-compatible
+            >>> auth = TokenSyncAuth(
+            ...     header_name='Authorization',
+            ...     prefix='Token'
```
sobolevn:
```suggestion
            ...     prefix='Token',
```

style nit

dmr/security/token/auth/base.py
```diff
@@ -0,0 +1,135 @@
...
+        'prefix',
+        'query_param',
+        'security_scheme_name',
+        'update_last_used',
+    )
+
+    def __init__(
+        self,
+        *,
+        header_name: str = 'X-API-Token',
```
sobolevn: Sorry, I don't quite like this design. This auth knows about:
- query param name
- cookie name
- header name

But, in no real case we would have to define an auth with all three of these values. Let's refactor exact auth classes to handle values that they really need and know about.

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,222 @@
...
+            raise NotAuthenticatedError
+
+    def mark_token_used(self, token: 'Token') -> None:
+        """Persist token usage timestamp after successful authentication."""
+        now = dt.datetime.now(dt.UTC)
+        self.token_model().objects.filter(pk=token.pk).update(
+            last_used_at=now,
+            updated_at=now,
+        )
+        token.last_used_at = now
```
sobolevn: You did not update `updated_at` instance attribute here. Why?

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,222 @@
...
+        await (
+            self
+            .token_model()
+            .objects.filter(pk=token.pk)
+            .aupdate(
+                last_used_at=now,
+                updated_at=now,
+            )
+        )
+        token.last_used_at = now
```
sobolevn: same here

dmr/security/token/auth/cookie.py
```diff
@@ -0,0 +1,101 @@
...
+from django.http import HttpRequest
+from typing_extensions import override
+
+from dmr.openapi.objects import Reference, SecurityScheme
+from dmr.security.token.auth.header import TokenAsyncAuth, TokenSyncAuth
+
+_DEFAULT_PARAM: Final = 'token'
+
+
+class CookieTokenSyncAuth(TokenSyncAuth):
```
sobolevn: I would prefer that `TokenSyncAuth` and `CookieTokenSyncAuth` would be independent types with no direct inheritance. 

If you need to reuse some method, you can create one more layer `_HeaderLikeSyncAuth` and `_HeaderLikeAsyncAuth` to inherit from it from both `CookieTokenSyncAuth` and `HeaderTokenSyncAuth`

dmr/security/token/auth/cookie.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    Sync opaque token auth reading from a cookie.
+
+    .. warning::
```
sobolevn, line 33: Please, make this secure by default. Take a look at our `DjangoSession` auth, just require to check the CSRF as the first part of the auth process. See `_ensure_csrf` method.

If people need to disable it, they can subclass the auth and disable it. 

We can't compromise the security for usability.

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,222 @@
...
+    request.user = user
+
+    async def auser() -> 'AbstractBaseUser':  # noqa: WPS430
+        return user
+
+    request.auser = auser
+    request.__dmr_token__ = token  # type: ignore[attr-defined]
+
+
+class TokenSyncAuth(_BaseTokenAuth, SyncAuth):
```
sobolevn:
```suggestion
class HeaderTokenSyncAuth(_BaseTokenAuth, SyncAuth):
```

dmr/security/token/auth/base.py
```diff
@@ -0,0 +1,135 @@
...
+        prefix: str = '',
+        query_param: str = 'token',
+        cookie_name: str = 'token',
+        security_scheme_name: str = 'token',
+        update_last_used: bool = True,
```
sobolevn: Please, document a warning that `update_last_used` uses an extra database call on every auth and will slow things down by a significan amount.

dmr/security/token/models.py
```diff
@@ -0,0 +1,185 @@
...
+    def is_revoked(self) -> bool:
+        """Return True if the token has been revoked."""
+        return self.revoked_at is not None
+
+    @property
+    def is_active(self) -> bool:
+        """Return True if the token is neither expired nor revoked."""
+        return not self.is_expired and not self.is_revoked
+
+    def revoke(self, *, at: dt.datetime | None = None) -> None:
```
sobolevn: Please, remove all methods from models. Models must not know anything about any business logic.

Revoking tokens is a part of the auth business logic.

dmr/security/token/models.py
```diff
@@ -0,0 +1,185 @@
...
+    'models.ForeignKey[_BaseFkData | AbstractBaseUser, AbstractBaseUser]'
+)
+_CharField: TypeAlias = 'models.CharField[_BaseFkData, str]'
+_DateTimeField: TypeAlias = (
+    'models.DateTimeField[str | dt.datetime | Combinable, dt.datetime]'
+)
+_DateTimeFieldNullable: TypeAlias = 'models.DateTimeField[str | dt.datetime | Combinable | None, dt.datetime | None]'  # noqa: E501
+
+
+class TokenManager(models.Manager['Token']):
```
sobolevn: Please, remove the manager. We don't allow to use managers in our apps (with very rare exceptions, like `User`).

All this logic is just a business logic. All of this can just be functions.

dmr/settings.py
sobolevn:
```suggestion
    Settings.auth_token_default_expiry: dt.timedelta(days=TOKEN_DEFAULT_EXPIRY_DAYS),
```

let's rename this field

docs/examples/auth/token/issue_token.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class TokenOut(BaseModel):
+    id: int
+    name: str
+    expires_at: dt.datetime | None
+    raw_token: str
+
+
+class IssueTokenController(Controller[PydanticSerializer]):
```
sobolevn, line 10: We must provide pre-existing views for this. You can do it in the next PR, but then this example must be removed, because it shows an incorrect code.

docs/examples/auth/token/revoke_token.py
```diff
@@ -0,0 +1,19 @@
+from django.contrib.auth.models import User
+
+from dmr import Controller
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.security import AuthenticatedHttpRequest
+from dmr.security.token import TokenSyncAuth, request_token
+
+
+class RevokeTokenController(Controller[PydanticSerializer]):
```
sobolevn, line 10: This example can stay.

docs/examples/auth/token/revoke_token.py
```diff
@@ -0,0 +1,19 @@
...
+    """Revoke the token used to make this request."""
+
+    request: AuthenticatedHttpRequest[User]
+    auth = (TokenSyncAuth(),)
+
+    def delete(self) -> None:
+        request_token(self.request, strict=True).revoke()
```
sobolevn: First: `token_auth = TokenSyncAuth()`
Then: `auth = (token_auth,)`

```suggestion
        token_auth.revoke_token(request_token(self.request, strict=True))
```

sobolevn: We must use business objects to do things, not Django ones. This way we can easily customize how `.remove_token()` works. 

You can't customize the `TokenModel.revoke_token` at all in the currect design.

docs/examples/auth/token/using_token_cookie.py
```diff
@@ -0,0 +1,19 @@
...
+from dmr.security.token import CookieTokenSyncAuth
+
+
+class APIController(Controller[PydanticSerializer]):
+    request: AuthenticatedHttpRequest[User]
+    auth = (CookieTokenSyncAuth(),)
+
+    def get(self) -> str:
+        # Let's test that `User` has the correct type:
+        assert self.request.user.username
```
sobolevn:
```suggestion
        assert self.request.user.is_authenticated
```

sobolevn (commented):
Thanks a lot for your work! It starts to look good!

docs/examples/auth/token/using_token_header.py
```diff
@@ -0,0 +1,19 @@
...
+from dmr.security.token import TokenSyncAuth
+
+
+class APIController(Controller[PydanticSerializer]):
+    request: AuthenticatedHttpRequest[User]
+    auth = (TokenSyncAuth(),)
+
+    def get(self) -> str:
+        # Let's test that `User` has the correct type:
+        assert self.request.user.username
```
milssky: You dont test type there =)

dmr/security/token/auth/header.py
Dresdn: Missed during some refactoring work

dmr/security/token/auth/base.py
Dresdn: Let me know your thoughts with a redesign in https://github.com/wemake-services/django-modern-rest/pull/1051/commits/4a67679af11a6bfd71437281b94aefcea14cf741

I did ignore `WPS214` in the base classes as doing a refactor would just make it harder to read - adding a mixin just would read weird. If you feel strongly otherwise, I can pull out the checks or db logic into a mixin.

dmr/security/token/auth/cookie.py
Dresdn: Reworked inheritance in https://github.com/wemake-services/django-modern-rest/commit/4a67679af11a6bfd71437281b94aefcea14cf741

dmr/security/token/auth/header.py
Dresdn: Renamed for consistency in https://github.com/wemake-services/django-modern-rest/pull/1051/commits/6a1ae37e98d97eaf096f49d31fca161f2f7f759e

dmr/settings.py
Dresdn: Renamed in https://github.com/wemake-services/django-modern-rest/pull/1051/commits/c0ac57e9a3b259e2c00002c2b1f08891126d1a00

dmr/security/token/models.py
Dresdn: Resolved in https://github.com/wemake-services/django-modern-rest/pull/1051/commits/bed97a5bad13071724fe7667df28186637cff8ef

dmr/security/token/models.py
Dresdn: Resolved in https://github.com/wemake-services/django-modern-rest/pull/1051/commits/bed97a5bad13071724fe7667df28186637cff8ef

Dresdn:
This isn't ready to be reviewed yet (except for some comments I left on unresolved convos). Any convos left open will be addressed, so I'll let you know when it's ready to be reviewed @sobolevn to save you time (like calling out CSRF twice :)

dmr/security/token/auth/cookie.py
```diff
@@ -0,0 +1,104 @@
...
+
+class CookieTokenSyncAuth(_BaseTokenSyncAuth):
+    """
+    Sync opaque token auth reading from a cookie.
+
+    .. warning::
+        Cookie-based authentication is vulnerable to CSRF attacks in
+        browser-facing contexts.  Ensure that
+        ``django.middleware.csrf.CsrfViewMiddleware`` is active whenever
+        this auth class is used in a browser-facing application.
```
sobolevn, line 37: Please, don't forget to enforce CSRF check in this class by default.

dmr/security/token/auth/header.py
```diff
@@ -0,0 +1,167 @@
+from django.http import HttpRequest
+from typing_extensions import override
+
+from dmr.openapi.objects import Reference, SecurityScheme
+from dmr.security.token.auth.base import (
+    _BaseTokenAsyncAuth,  # noqa: WPS450  # pyright: ignore[reportPrivateUsage]
+    _BaseTokenSyncAuth,  # noqa: WPS450  # pyright: ignore[reportPrivateUsage]
+)
+
+_AUTH_DESCRIPTION = 'Opaque token authentication'
```
sobolevn:
```suggestion
_AUTH_DESCRIPTION: Final = 'Opaque token authentication'
```

dmr/security/token/logic/token.py
```diff
@@ -0,0 +1,121 @@
...
+from django.utils.crypto import salted_hmac
+
+from dmr.types import EMPTY
+
+if TYPE_CHECKING:
+    from django.contrib.auth.base_user import AbstractBaseUser
+
+    from dmr.security.token.models import Token
+
+_RAW_TOKEN_SIZE = 32
```
sobolevn:
```suggestion
_RAW_TOKEN_SIZE: Final = 32
```

dmr/security/token/logic/token.py
```diff
@@ -0,0 +1,121 @@
...
+        secret=settings.SECRET_KEY,
+        algorithm='sha256',
+    ).hexdigest()
+
+
+def token_create(
+    *,
+    user: 'AbstractBaseUser',
+    name: str,
+    expires_at: object = EMPTY,
```
sobolevn:
```suggestion
    expires_at: dt.datetime | None | EmptyType = EMPTY,
```

dmr/security/token/logic/token.py
```diff
@@ -0,0 +1,121 @@
...
+        expires_at=resolved_expires_at,
+    )
+    return token, raw_token
+
+
+async def token_acreate(
+    *,
+    user: 'AbstractBaseUser',
+    name: str,
+    expires_at: object = EMPTY,
```
sobolevn: same here

docs/examples/auth/token/using_token_cookie.py
```diff
@@ -0,0 +1,19 @@
...
+        return 'authed'
```
sobolevn, line 16: Don't forget to use `assert self.request.user.is_authenticated` here.

docs/examples/auth/token/issue_token.py
```diff
@@ -0,0 +1,39 @@
...
+    """Issue a new token for the authenticated user."""
+
+    request: AuthenticatedHttpRequest[User]
+    auth = (HeaderTokenSyncAuth(),)
```
sobolevn: This does not seem correct to me: user needs some other way of auth to get the auth token :)
Otherwise, there would be no token to auth with to get the token :)

docs/examples/auth/token/using_token_query.py
```diff
@@ -0,0 +1,19 @@
...
+
+from dmr import Controller
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.security import AuthenticatedHttpRequest
+from dmr.security.token import QueryTokenSyncAuth
+
+
+class APIController(Controller[PydanticSerializer]):
+    request: AuthenticatedHttpRequest[User]
+    auth = (QueryTokenSyncAuth(),)
```
sobolevn, line 11: Please add a warning here, that it is not really safe :)

docs/pages/auth/token.rst
```diff
@@ -0,0 +1,152 @@
...
+  :language: python
+
+
+Token lifecycle
+---------------
+
+Tokens are created via :class:`~dmr.security.token.models.Token`
+manager methods:
+
+- :meth:`~dmr.security.token.models.TokenManager.create_token`
```
sobolevn: This is no longer true

docs/pages/auth/token.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Both methods return ``(token_instance, raw_token)``.
+Only token hash is stored in the database.
+
+You can revoke tokens with:
+
+- :meth:`~dmr.security.token.models.Token.revoke`
+- :meth:`~dmr.security.token.models.Token.arevoke`
+
+.. literalinclude:: /examples/auth/token/token_lifecycle.py
```
sobolevn: this can be a mermaid chart

docs/pages/auth/token.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Opaque tokens can be sent by clients in three common ways:
+
+1. Query parameter
+2. Header
+3. Cookie
+
+Example request shapes:
+
+.. code-block:: text
```
sobolevn: 1. Use headers near the examples, not above it. It should be like so: `1. Header\n .. code-block:: text\n GET /api/thing HTTP/1.1 ...`
2. Do not use `Query` as the first one, use it as the last one and add a warning

docs/pages/auth/token.rst
```diff
@@ -0,0 +1,152 @@
...
+  :class:`~dmr.security.token.CookieTokenAsyncAuth`
+
+By default, header auth expects:
+
+- ``X-API-Token: <raw_token>``
+
+You can customize behavior per auth instance,
+for example to support ``Authorization`` header styles:
+
+- ``HeaderTokenSyncAuth(header_name='Authorization', prefix='Token')``
```
sobolevn: Please, document DRF compatible way here as well.

sobolevn (commented):
Thanks a lot! Great work! Just some minor polishing is required right now, we are almost there.

Please, resolve all my older comments that were fixed :)
It is rather hard to navigate through them.

dmr/security/token/auth/base.py
sobolevn: Awesome!

Dresdn:
Thanks for doing another review!

> Please, resolve all my older comments that were fixed :) It is rather hard to navigate through them.

Yes, I have been resolving them as I go using them as a TODO. All comments will be resolved before final review!

dmr/security/token/logic/token.py
Dresdn: @sobolevn - I'm not quite sure how to type this. I tried:

```python
expires_at: dt.datetime | EMPTY | None = EMPTY,
```

but that gives a:

```bash
"dmr.types.EMPTY" is not valid as a type  [valid-type]
```

Asking AI to help and it suggested doing this which seems really wrong given `Sentinel` ...

```python
# dmr/types.py

class EmptyType:
    """Sentinel type for unset/missing optional parameters."""
    __slots__ = ()

    def __repr__(self) -> str:
        return 'EMPTY'

EMPTY: Final[EmptyType] = EmptyType()
```

Thoughts on how to do this?

Dresdn:
@sobolevn - All your comments except for the typing issue are resolved. I went through and built the docs too.

dmr/security/token/admin.py
```diff
@@ -0,0 +1,78 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Generic, TypeVar, override
```
sobolevn:
```suggestion
from typing import TYPE_CHECKING, Generic, TypeVar

from typing_extensions import override
```

dmr/security/token/logic/token.py
```diff
@@ -0,0 +1,121 @@
...
+        secret=settings.SECRET_KEY,
+        algorithm='sha256',
+    ).hexdigest()
+
+
+def token_create(
+    *,
+    user: 'AbstractBaseUser',
+    name: str,
+    expires_at: object = EMPTY,
```
sobolevn: My bad :)

We are using `Sentinel` type for this now

```suggestion
    expires_at: dt.datetime | Sentinel | None = EMPTY,
```

It might require a `# type: ignore` commen somewhere, this happens because it is not fully supported by `mypy` / `pyrefly` yet :(

Dresdn: Typing it just as `Sentinel` makes `mypy` and `pyrefly` happy - just `pylance` doesn't like it.

Dresdn:
Should be good now. Renamed the `test_token_auth/test_admin.py` to have a unique basename, but looking at the `test_jwt/test_blocklist` modules, there's a few common names in there that could conflict in the future (`test_admin.py` and `test_models.py`). Should probably rename them to prevent anything in the future of the project conflicting with it, but curious your thoughts.

Dresdn:
D'oh! Rebase got me as `test_token/test_token_auth.py` was supposed to be deleted when I refactored it - accidentally left the module in.

However, looks like there's an issue with python 3.11 and a test (https://github.com/wemake-services/django-modern-rest/actions/runs/28523994471/job/84556495229#step:9:3360) - let me run tests locally on that version and see what's going on before you trigger another CI run.

dmr/security/token/auth/base.py
```diff
@@ -113,12 +112,11 @@ def get_token(self, raw_token: str) -> 'Token':
...
-                token_hash=hashed_token,
-            )
-        except ObjectDoesNotExist:
-            raise NotAuthenticatedError from None
+        token = model.objects.select_related('user').filter(
+            token_hash=hashed_token,
+        ).first()
+        if token is None:
+            raise NotAuthenticatedError
         return token
```
Dresdn: Apparently this is a bug in python 3.11. Claude came back with the following explanation:

> Python 3.11's sys.settrace drops the trace on a suspended parent
coroutine frame when an exception is raised *during* an await and
caught in the same coroutine. This caused the two lines following
`await dmr_async_rf.wrap(...)` in test_async_token_auth_unknown_token
to appear uncovered under 3.11 only (3.12+ use sys.monitoring).
>
> Replace aget()/get() + try/except ObjectDoesNotExist with
afirst()/first() + if None check. The await now always completes
normally; NotAuthenticatedError is raised synchronously after,
which 3.11's tracer handles correctly.

My preferred pattern is the `try .get()/raise`, but given this, the SQL is essentially the same and solves the 3.11 coverage issue. I can leave a comment in the file for any future devs if you'd prefer.

sobolevn (approved):
Thank you! This is an incredible feature, very well delivered!
It looks really promising!

There are several follow-ups that we can address in the future PRs (to polish this even more):
- [x] Provide an alternative `SECRET_KEY` if needed to all auth classes, just like JWT does
- [x] `_csrf.py` module can be public -> `protocols/csrf.py`
- [ ] We can probably reuse some existing Django logic instead of `_set_request_attrs` function
- [x] We can move common `.filter(token_hash=hashed_token)` query to a helper function instead of duplicating it
- [x] Refactor `CookieTokenSyncAuth.provide_response_specs` to not duplicate the async one
- [x] The same for `HeaderTokenAsyncAuth` and `HeaderTokenSyncAuth`
- [x] And `QueryTokenSyncAuth`
- [x] Return `Token` from `token_revoke` and similar functions
- [x] Improve the docs with examples: `IssueTokenController` needs rework
- [x] `RevokeTokenController` needs a return value
- [x] We create a temp token in `run_examples.py` and pass them for demos
- [x] Provide existing views to obtain tokens, this is a very common use-case

Would you like to continue this amazing work?

Dresdn:
More than happy to @sobolevn. I see the ticket you opened and will go off of that. Thanks!

PR #1075, opened 2026-06-13
Implement DynamicThrottle with sync/async

### LINKED ISSUE #1056: Unable to override global AsyncThrottle with controller-level SyncThrottle
### What's wrong?

Hi,

I may be misunderstanding how throttling is intended to work, so I wanted to ask whether this behavior is expected or if it might be a bug.

I have configured an `AsyncThrottle` in my global `DMR_SETTINGS`:

```python
DMR_SETTINGS = {
    Settings.throttling: [
        AsyncThrottle(...)
    ]
}
```

Then I created a synchronous controller and explicitly defined a `SyncThrottle` on the controller itself:

```python
class MyEndpoint(Controller):
    throttling = [
        SyncThrottle(...)
    ]
```

I expected the controller-level throttle to be used for this endpoint. However, DMR raises the following error:

```text
dmr.exceptions.EndpointMetadataError:
All throttling instances must be subtypes of
<class 'dmr.throttling.base.SyncThrottle'>
for self.endpoint_name='MyEndpoint.post'
```

From the error and my testing, it appears that the global throttles and controller throttles are being combined, rather than the controller throttles overriding the global ones. As a result, the endpoint ends up with both an `AsyncThrottle` and a `SyncThrottle`, which causes the validation error.

Is this the intended behavior? If so, what is the recommended way to define endpoint-specific throttling when global throttling is configured? Or is this a bug in the throttling resolution logic?

Thanks.

### How it should be?

I expected the controller-level `throttling` configuration to take precedence over the global throttling configuration for that endpoint.

If the intended behavior is to merge global and controller throttles, it would be helpful if this were documented and if there were a way to explicitly override global throttles when needed.

### Used versions

"django-modern-rest[msgspec,pydantic,jwt]>=0.10.0"

### OS information

Fedora Linux 6.17.12-300.fc43.x86_64

azad-mammedov:
Add DynamicThrottle to automatically resolve to SyncThrottle or AsyncThrottle based on the target endpoint type. This simplifies global throttling configuration by allowing a single dynamic throttle to work for both sync and async controllers.

Update endpoint metadata validation so DynamicThrottle resolves at build time, and add regression tests covering sync/async resolution, settings-level throttling, schema generation, and reporting behavior.

- Closes #1056


dmr/throttling/base.py
```diff
@@ -334,6 +334,66 @@ async def _check(
...
 
+class DynamicThrottle(_BaseThrottle[Any]):
+    """Throttle that resolves to SyncThrottle or AsyncThrottle.
+
+    Automatically resolves based on the endpoint it is applied to.
+    Use this in global settings to avoid manually matching
+    sync and async endpoints with their respective throttle types.
+    """
+
+    def __init__(
```
sobolevn, line 348: Let's use a different design:

```python
if not TYPE_CHECKING:
     def __init__(self, *args, **kwargs) -> None:
          self._args = args
          self._kwargs = kwargs
```

dmr/throttling/base.py
```diff
@@ -334,6 +334,66 @@ async def _check(
...
+                ``AsyncThrottle`` for async endpoints.
+                This is derived from ``base_type`` in ``_build_throttling``.
+
+        Returns:
+            A concrete :class:`SyncThrottle` or :class:`AsyncThrottle`
+            instance with the same configuration as this
+            :class:`DynamicThrottle`.
+
+        """
+        return throttle_cls(
```
sobolevn: And then: `return throttle_cls(*self._args, **self._kwargs)`

dmr/validation/endpoint_metadata.py
```diff
@@ -528,19 +528,26 @@ def _build_throttling(  # noqa: WPS231
...
+        # to each request, so it would not be possible to mutate it by accident.
+        # We resolve `DynamicThrottle` to get the actual throttle instance.
+        throttling = [
+            throttle.resolve(base_type)
+            if isinstance(throttle, DynamicThrottle)
+            else throttle
+            for throttle in (
+                *payload_throttling,
+                *(self.controller_cls.throttling or ()),
+                *settings_throttling,
```
sobolevn: Nope, we should only allow `DynamicThrottling` in `settings`, not in controller / endpoint. They must use regular classes.

dmr/endpoint.py
```diff
@@ -699,7 +708,10 @@ def validate(  # noqa: WPS211  # pyright: ignore[reportInconsistentOverload]
     renderers: Sequence[Renderer] | None = None,
     validate_negotiation: bool | None = None,
     auth: Sequence[AsyncAuth] | Sequence[SyncAuth] | None = (),
-    throttling: Sequence[AsyncThrottle] | Sequence[SyncThrottle] | None = (),
+    throttling: Sequence[AsyncThrottle]
```
sobolevn:
```suggestion
    throttling: ( Sequence[AsyncThrottle]
```

will this be possible here? And also adding `)` in the end of the type annotation expression? It might produce a better formatting result.

dmr/endpoint.py
```diff
@@ -669,7 +675,10 @@ def validate(
...
```
sobolevn:
```suggestion
    throttling: ( Sequence[AsyncThrottle]
```

will this be possible here? And also adding `)` in the end of the type annotation expression? It might produce a better formatting result.

dmr/endpoint.py
```diff
@@ -860,7 +872,10 @@ def modify(
...
```
sobolevn: And here. If not, please create a type alias. Right now it is hard to read.

dmr/endpoint.py
```diff
@@ -923,7 +941,10 @@ def modify(
...
```
sobolevn: and here

dmr/endpoint.py
```diff
@@ -954,7 +975,10 @@ def modify(  # noqa: WPS211
...
```
sobolevn: and here

tests/test_unit/test_throttling/test_throttling.py
```diff
@@ -377,3 +381,333 @@ def get(self) -> str:
...
+    freezer: FrozenDateTimeFactory,
+    *,
+    serializer: type[BaseSerializer],
+) -> None:
+    """Ensures DynamicThrottle resolves to SyncThrottle for sync endpoints."""
+
+    class _SyncController(
+        Controller[serializer],  # type: ignore[valid-type]
+    ):
+        @modify(throttling=[DynamicThrottle(1, Rate.second)])
```
sobolevn: Please, modify this test to use `DynamicThrottle` from settings.

Test that it is impossible to be used on controller / endpoint levels.

Test that the same instance is reused for different sync and async controllers.

Test that endpoint metadata never has `DynamicThrottle` instances in final endpoints.

tests/test_unit/test_throttling/test_throttling.py
```diff
@@ -377,3 +381,333 @@ def get(self) -> str:
...
+    assert response.status_code == HTTPStatus.OK, response.content
+    assert response.headers == {'Content-Type': 'application/json'}
+    assert json.loads(response.content) == 'inside'
+
+
+@pytest.mark.parametrize('serializer', serializers)
+@pytest.mark.filterwarnings(
+    'ignore::dmr.throttling.backends.django_cache.UnsafeCacheBackendWarning',
+)
+def test_dynamic_throttle_multiple_sources(
```
sobolevn: You can remove this test.

tests/test_unit/test_throttling/test_throttling.py
```diff
@@ -377,3 +381,333 @@ def get(self) -> str:
...
+        'detail': [{'msg': 'Too many requests', 'type': 'ratelimit'}],
+    })
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize('serializer', serializers)
+@pytest.mark.filterwarnings(
+    'ignore::dmr.throttling.backends.django_cache.UnsafeCacheBackendWarning',
+)
+async def test_dynamic_throttle_multiple_sources_async(
```
sobolevn: You can remove this test.

tests/test_unit/test_throttling/test_throttling_and_auth.py
```diff
@@ -251,3 +251,191 @@ def test_throttle_sync_before_auth(
         response.content
     )
     assert response.headers['RateLimit'] == '"per-minute";r=0;t=59'
+
+
+# DynamicThrottle
+
+
+class _DynamicSyncController(Controller[PydanticSerializer]):
```
sobolevn: these tests are not needed, please remove them.

tests/test_unit/test_throttling/test_throttling_and_parsing.py
```diff
@@ -96,3 +96,90 @@ async def test_throttle_before_negotiation_async(
         request_endpoint(request, strict=True)
         is _AsyncController.api_endpoints['GET']
     )
+
+
+class _DynamicSyncController(Controller[PydanticSerializer]):
```
sobolevn: these tests are not needed, please remove them.

tests/test_unit/test_throttling/test_throttling_real.py
```diff
@@ -86,3 +86,83 @@ async def test_throttle_async_per_controller(
     assert json.loads(response.content) == snapshot({
         'detail': [{'msg': 'Too many requests', 'type': 'ratelimit'}],
     })
+
+
+class _DynamicSyncEndpointController(Controller[PydanticFastSerializer]):
```
sobolevn: these tests are not needed, please remove them.

tests/test_unit/test_throttling/test_throttling_reports.py
```diff
@@ -437,3 +443,258 @@ async def test_leaky_bucket_async(
         'Content-Type': 'application/json',
     }
     assert json.loads(response.content) == 'inside'
+
+
+class _DynamicReportsController(Controller[PydanticSerializer]):
```
sobolevn: And these tests as well.

tests/test_unit/test_throttling/test_throttling_snapshots.py
```diff
@@ -115,3 +115,108 @@ def test_throttled_schema_with_errors(snapshot: SnapshotAssertion) -> None:
         )
         == snapshot
     )
+
+
+class _DynamicDefaultController(Controller[PydanticSerializer]):
```
sobolevn, line 127: You can keep this test, but change it to use settings. And only use a single controller.

tests/test_unit/test_throttling/test_throttling_validation.py
```diff
@@ -35,3 +35,39 @@ class _AsyncEndpointController(
 
             async def get(self) -> str:
                 raise NotImplementedError
+
+
+def test_dynamic_throttle_no_error_on_sync(
```
sobolevn: These tests are awesome, just refator them.

sobolevn (commented):
I would recommend to move `DynamicThrottle` tests to a different test file.

Please, also add docs.

sobolevn (changes requested):
One more idea: let's refactor `DynamicThrottle` to be created like: `SyncOrAsyncThrottle(SyncThrottle(...), AsyncThrottle(...))`.

This way people can customize the types they pass and we won't have to inherit `DynamicThrottle` from `_BaseThrottle`.
And it would be consistent with the future `SyncOrAsyncAuth` API.

azad-mammedov:
Thanks for the review and feedback. I am working on applying these changes now and will let you know once they are ready for another look.

dmr/endpoint.py
```diff
@@ -62,6 +62,10 @@
     from dmr.routing import Router
     from dmr.validation.response import ValidatedModification
 
+_ThrottlingT: TypeAlias = (
```
sobolevn: I would prefer to name it: `_ThrottlingDef`

tests/test_unit/test_throttling/test_throttling_dynamic.py
```diff
@@ -0,0 +1,181 @@
...
+    dmr_rf: DMRRequestFactory,
+    freezer: FrozenDateTimeFactory,
+    settings: LazySettings,
+    *,
+    serializer: type[BaseSerializer],
+) -> None:
+    """Ensures SyncOrAsyncThrottle in settings resolves and rate-limits sync."""
+    settings.DMR_SETTINGS = {
+        Settings.throttling: [
+            SyncOrAsyncThrottle(
```
sobolevn: Can you please define this instance on the module level and reuse it for all tests here? This way we would simulate its real use as a singleton.

sobolevn (commented):
Great stuff! Two minor comments and we are ready to go 🎉

dmr/throttling/base.py
```diff
@@ -334,6 +334,36 @@ async def _check(
...
+
+    .. versionadded:: 0.10.0
+    """
+
+    __slots__ = ('_async', '_sync')
+
+    def __init__(
+        self,
+        sync: SyncThrottle,
+        async_: AsyncThrottle,
```
sobolevn, line 351:
```suggestion
        async_: AsyncThrottle,
        /,
```

dmr/endpoint.py
sobolevn: I would prefer to name it: `_ThrottlingDef`

tests/test_unit/test_throttling/test_throttling_dynamic.py
sobolevn: Can you please define this instance on the module level and reuse it for all tests here? This way we would simulate its real use as a singleton.

sobolevn (commented):
GitHub is lagging, it does not show my last review :(

Thanks, great work! Several minor nitpicks are left.

azad-mammedov:
Thanks so much for the kind words and the detailed feedback — really appreciate the time you put into this! 😊 Working on the remaining nitpicks now and will push the updates as soon as they're ready!

sobolevn (approved):
Amazing! 🎉

sobolevn:
Oups, we forgot about the CHANGELOG entry, can you please send one more pr with that?

azad-mammedov:
Of course! No worries  I'll open one more PR with the CHANGELOG entry shortly.

PR #1153, opened 2026-07-17
feat: add JWT access token verification flow

### LINKED ISSUE #1129: Add a built-in JWT access token verification flow.
# Feature

## Thesis
Add a built-in JWT access token verification flow.

For example, django-modern-rest could provide a reusable controller:

```python
from dmr.security.jwt.views import VerifyTokenSyncController


class TokenVerifyController(VerifyTokenSyncController):
    pass
```

Expected behavior:

- accept an encoded token from the request body
- decode and validate the JWT
- reject malformed tokens
- reject expired tokens
- reject refresh tokens when an access token is expected
- verify that the token subject belongs to an existing active user
- return a successful empty response when the token is valid

The exact public API can be different. It could be implemented as a controller,
auth class, or reusable JWT verification component.

## Reasoning
django-modern-rest already provides JWT obtain and refresh controllers, but
applications commonly also need a token verification endpoint.

Without built-in support, each project has to implement the same logic manually:

```python
from django.conf import settings
from dmr.exceptions import NotAuthenticatedError
from dmr.security.jwt.token import JWToken


token = JWToken.decode(
    encoded_token=encoded_token,
    secret=str(settings.SECRET_KEY),
    algorithm="HS256",
)

if token.extras.get("type") != "access":
    raise NotAuthenticatedError

user = User.objects.get(pk=token.sub)

if not user.is_active:
    raise NotAuthenticatedError
```

This logic is not application-specific. It is part of the common JWT API
surface.

Providing it in django-modern-rest would reduce duplication and make JWT support
more complete. It would also make error handling more consistent with the
existing obtain and refresh controllers.

vyhuholl:
# I have made things!

* Add VerifyTokenSyncController/VerifyTokenAsyncController to dmr/security/jwt/views.py (modeled after RefreshTokenAsyncController/RefreshTokenSyncController).
* Implement token decoding and validation logic (204 if token is valid, 401 if token is incorrect or expired, user is unauthenticated or it is a refresh token).
* Add tests and docs, update changelog.

## AI Policy

## Checklist

## Related issues

Closes #1129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ of requirements for an API to count as public.
 ### Features
 
 - Adds "Opaque Token" auth backend, #1051
+- Added `VerifyTokenSyncController` and `VerifyTokenAsyncController`
+  reusable controllers to verify JWT access tokens, #1129
 
 
 ## Version 0.11.0 (2026-06-27)
--- a/django_test_app/server/apps/jwt_auth/urls.py
+++ b/django_test_app/server/apps/jwt_auth/urls.py
@@ -25,6 +25,16 @@ router = Router(
             name='jwt_refresh_async',
         ),
         path(
+            'jwt-verify-sync/',
+            views.VerifySyncController.as_view(),
+            name='jwt_verify_sync',
+        ),
+        path(
+            'jwt-verify-async/',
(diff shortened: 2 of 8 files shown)


sobolevn (approved):
Great stuff! Thank you!

PR #1154, opened 2026-07-17
fix: compile and add missing django.mo file for the Turkish language

vyhuholl:
# I have made things!
`django.mo` file for the Turkish language was missing (and so the Turkish translation would not work). I've compiled and added it.

## AI Policy

## Checklist


sobolevn (approved):
Thanks!

PR #1165, opened 2026-07-21
Add per-controller request validation error status

### LINKED ISSUE #1130: Add allow controllers to customize request validation error status code
# Feature

## Thesis
Add a framework-level way to customize request validation error status codes
per controller.

For example:

```python
from http import HTTPStatus

from dmr.controller import Controller
from dmr.plugins.pydantic import PydanticSerializer


class TokenObtainController(Controller[PydanticSerializer]):
    validation_error_status = HTTPStatus.BAD_REQUEST
```

Or via an explicit controller option / hook:

```python
from http import HTTPStatus

from dmr.controller import Controller
from dmr.plugins.pydantic import PydanticSerializer


class TokenObtainController(Controller[PydanticSerializer]):
    def get_validation_error_status(self) -> HTTPStatus:
        return HTTPStatus.BAD_REQUEST
```

The exact public API can be different. The main goal is to avoid replacing the
global error handler just to customize validation error status for a specific
controller.

## Reasoning

Some endpoints need endpoint-specific request validation semantics.

For example, JWT token endpoints often return `400 Bad Request` for malformed
request payloads. This can be required for compatibility with existing clients
or with already published API contracts.

Right now, applications have to solve this outside of django-modern-rest by
adding custom controller attributes and replacing the global error handler:

```python
from http import HTTPStatus


class TokenObtainController(...):
    validation_error_status = HTTPStatus.BAD_REQUEST
```

Then the custom error handler has to inspect this attribute manually:

```python
status_code = getattr(
    controller,
    "validation_error_status",
    HTTPStatus.UNPROCESSABLE_ENTITY,
)
```

This works, but it is not framework-level API. It also duplicates logic that
belongs close to request validation handling.

A supported per-controller setting would make this behavior explicit,
documented, and reusable.

azad-mammedov:
Add a `request_validation_error_status` class attribute on `Controller` (default `400`, so existing behavior is unchanged) and read it when raising the request `ValidationError`. Controllers that need a different code, e.g. `422 Unprocessable Entity`, can now set it explicitly.

Adds tests, docs, and a changelog entry.

Closes #1130

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ of requirements for an API to count as public.
 - Adds "Opaque Token" auth backend, #1051
 - Added `VerifyTokenSyncController` and `VerifyTokenAsyncController`
   reusable controllers to verify JWT access tokens, #1129
+- Adds `request_validation_error_status` controller attribute
+  to customize the HTTP status code returned on request
+  validation errors, #1130
 
 ### Bugfixes
 
--- a/dmr/controller.py
+++ b/dmr/controller.py
@@ -114,6 +114,8 @@ class Controller(Generic[_SerializerT_co], View):  # noqa: WPS214
             unsafe throttle Django cache backends?
         error_model: Schema type that represents
             and validates common error responses.
+        request_validation_error_status: HTTP status code returned when an
+            incoming request fails validation. Defaults to ``400``.
         is_abstract: Whether or not this controller is abstract.
             We consider controller "abstract" when it does not have
             exact serializer type or exact ``api_endpoints`` instances.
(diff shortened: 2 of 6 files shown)


azad-mammedov:
Hey @sobolevn  This implements #1130 — a per-controller request_validation_error_status attribute on Controller (default 400, behavior unchanged), read at the raise site instead of the hardcoded 400.

Kept it minimal, matching existing knobs like validate_responses. Tests, docs, and changelog included 

If you have ideas on how to improve it further, I'd be happy to take them on in a follow-up PR.

milssky:
it seems to me that the 400 status code is not the most successful. It shows errors in the structure, not the content. For example, an open parenthesis or quotation mark.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status

sobolevn:
I am not sure about the design though :(
I would have to think about for a bit, maybe I will come up with a solution.

Current design does not follow our layered architecture, it treats this exception specially. The alternative is also quite straight forward.

azad-mammedov:
Thanks for the feedback, @sobolevn and @milssky! ❤️

I opened this PR as a minimal, temporary fix for #1130, mainly to unblock the immediate need for a per-controller status code without touching the broader architecture. I understand the concern about it treating the ValidationError case specially instead of following the layered architecture properly.

If it's okay with you, I'd like to open a separate feature issue to discuss the proper design with @sobolevn — or we can discuss it here on this PR if that's easier. Happy to close or rework this PR once we land on an approach that fits better.

sobolevn:
Let's discuss it on the issue, thanks! :)

Issue #1131, opened 2026-07-06
Provide a way to test throttling


sobolevn:
Currently it is rather hard to write tests that assert that throttling is reached. It would be awesome to think about the proper API for that.

sesav:
I believe I'm capable of this. An approach I propose is  to add two small, composable helpers in dmr.test:

1. `throttle_state(Controller).exhaust(request)` (.aexhaust as well)
    Pre-fills the throttle state for the request's identity so that the next real request is guaranteed to be throttled. 
    No loop, no knowing max_requests, handles all throttles on the endpoint at once.

2. `assert_throttled(response, *, limit=…, reset=…, retry_after=…)`
    Combines the 429 status, header check, and ratelimit body check into a single call.
    Header args accept int, str, or a matcher such as `dirty_equals.IsStr()` for real-time (non-frozen) tests.

```python
def test_throttled(dmr_rf):
    throttle_state(MyController).exhaust(dmr_rf.get('/x/'))

    response = MyController.as_view()(dmr_rf.get('/x/'))
    assert_throttled(response, limit=5)
```

Also expose a `dmr_throttle_state` pytest fixture, mirroring `dmr_rf`. 

I'd love to hear your opinion.

sobolevn:
Nice! I would only want to highlight that throttling might be rather big, for example 1000/h. The test will be really slow for 1000 real requests. So, we still need to patch it with 1 or 2 real requests to make the test really managable.

sesav:
Got it. I'll give it a shot.

PR #1167, opened 2026-07-23
 Add a test API for asserting that throttling is reached #1131

sesav:
Ref https://github.com/wemake-services/django-modern-rest/issues/1131.

## Solution

Instead of seeding cache state, the helpers mock the count: they temporarily lower the endpoint's first throttle and drive a few real requests through `as_view()`, so the `429` is genuine and nothing is coupled to a backend's storage.

New in `dmr.test`:

- `assert_throttling` and `assert_async_throttling` — one call that lowers the limit, drives the requests and asserts the `429`, returning the response for further checks.

- `reduced_throttling` — the context manager underneath, for tests that need custom bodies or auth. `line` picks which throttle line to lower; the reduced copy also gets an hour-long window, otherwise a `2/second` throttle would reset between the requests and never reject.

- `assert_throttled` — status, headers and the `ratelimit` body in a single check. Header names are never hardcoded: given the `throttle`, it asks its providers what they report, so `XRateLimit`, `RateLimitIETFDraft` and custom providers work alike. Expected values go into `headers` as an `int`, a `str` or a matcher like `dirty_equals.IsStr()`.

Only the endpoint under test is affected, and its throttling is restored on exit. All four helpers are re-exported as `dmr_*` fixtures, with public protocols in `dmr_pytest` to annotate them, plus the `HeaderValue` and `ThrottlingLine` types used in their signatures.

## Examples

```python
_URL: Final = '/reports/'


class ReportsController(Controller[PydanticSerializer]):
    # A large real budget: reaching it with real requests would be slow.
    throttling = (SyncThrottle(1000, Rate.hour),)

    def get(self) -> str:
        return 'inside'


class AsyncReportsController(Controller[PydanticSerializer]):
    throttling = (AsyncThrottle(1000, Rate.hour),)

    async def get(self) -> str:
        return 'inside'


class IETFReportsController(Controller[PydanticSerializer]):
    # Reports `RateLimit` and `RateLimit-Policy`, no `X-RateLimit-*`:
    throttling = (
        SyncThrottle(1000, Rate.hour, response_headers=[RateLimitIETFDraft()]),
    )

    def get(self) -> str:
        return 'inside'


class PerUserReportsController(Controller[PydanticSerializer]):
    throttling = (
        SyncThrottle(
            1000,
            Rate.hour,
            cache_key=RemoteAddr(runs_before_auth=False),
        ),
    )

    def get(self) -> str:
        return 'inside'


# 1. One call: lower the limit, drive the requests, assert the `429`.
def test_reports_are_throttled(
    dmr_rf: DMRRequestFactory,
    dmr_assert_throttling: AssertThrottlingFixture,
) -> None:
    response = dmr_assert_throttling(
        ReportsController,
        dmr_rf,
        _URL,
        headers={'X-RateLimit-Limit': 2},
    )
    assert response.status_code == HTTPStatus.TOO_MANY_REQUESTS


# 2. The same for async controllers.
@pytest.mark.asyncio
async def test_reports_are_throttled_async(
    dmr_async_rf: DMRAsyncRequestFactory,
    dmr_assert_async_throttling: AssertAsyncThrottlingFixture,
) -> None:
    await dmr_assert_async_throttling(
        AsyncReportsController,
        dmr_async_rf,
        _URL,
    )


# 3. Any provider works: header names come from the throttle, not from us.
def test_ietf_headers(
    dmr_rf: DMRRequestFactory,
    dmr_assert_throttling: AssertThrottlingFixture,
) -> None:
    dmr_assert_throttling(
        IETFReportsController,
        dmr_rf,
        _URL,
        headers={'RateLimit': IsStr(regex=r'"[^"]+";r=0;t=\d+')},
    )


# 4. Target a specific throttle line.
def test_per_user_limit(
    dmr_rf: DMRRequestFactory,
    dmr_assert_throttling: AssertThrottlingFixture,
) -> None:
    dmr_assert_throttling(
        PerUserReportsController,
        dmr_rf,
        _URL,
        line='after_auth',
    )


# 5. Custom error models: skip the `ratelimit` body check.
def test_custom_error_model(
    dmr_rf: DMRRequestFactory,
    dmr_assert_throttling: AssertThrottlingFixture,
) -> None:
    dmr_assert_throttling(ReportsController, dmr_rf, _URL, detail=False)


# 6. Fine control: drive the requests yourself, then assert.
def test_reports_throttling_manually(
    dmr_rf: DMRRequestFactory,
    dmr_reduced_throttling: ReducedThrottlingFixture,
    dmr_assert_throttled: AssertThrottledFixture,
) -> None:
    view = ReportsController.as_view()
    with dmr_reduced_throttling(ReportsController, max_requests=1) as throttle:
        for _ in range(throttle.max_requests):
            allowed = view(dmr_rf.get(_URL))
            assert allowed.status_code == HTTPStatus.OK

        rejected = view(dmr_rf.get(_URL))
        assert isinstance(rejected, HttpResponse)
        dmr_assert_throttled(
            rejected,
            throttle=throttle,  # every header it reports must be present
            headers={'X-RateLimit-Limit': 1, 'Retry-After': Rate.hour},
        )
```

`django.test.TestCase` works the same way through the plain `dmr.test` functions, both styles are documented with runnable examples in `docs/pages/testing.rst`.

## Changes to existing code

- `SyncThrottle.replace` / `AsyncThrottle.replace` — general, keyword-only, also backing `copy.replace` on Python 3.13+ via `__replace__`;

- `SyncThrottle.response_headers` / `AsyncThrottle.response_headers` — public access to the providers, so the assertion helper can reuse them instead of hardcoding names;

- `dmr/test.py` is split into a package (the client code moved verbatim);

- `dmr_pytest.py` became a package with `py.typed`, without which the published protocols are `import-untyped` for downstream users.

## How it works

The helpers stay black-box: no cache seeding, no coupling to storage. The reduced throttle is installed with `dataclasses.replace` on a fresh `EndpointMetadata`, so only the endpoint under test is touched and no shared state is mutated. Real requests go through `as_view()`, producing genuine `429` responses and headers.


dmr/test.py
```diff
@@ -179,3 +191,184 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+
+def _assert_ratelimit_detail(response: HttpResponse) -> None:
+    body = json_loads(response.content.decode(response.charset or 'utf8'))
+    problems = body['detail']
+    assert problems, (  # noqa: S101
+        f'Expected a non-empty `detail` in throttled response, got {body!r}'
+    )
+    assert all(  # noqa: S101
+        problem.get('type') == 'ratelimit' for problem in problems
+    ), body
```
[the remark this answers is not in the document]
sesav: I tried it, too. No, thanks!

sesav:
Currently, the PR is set to Draft. I'll make some more additions or changes later on.

dmr/test.py
```diff
@@ -1,13 +1,25 @@
...
+    from dmr.controller import Controller
+    from dmr.endpoint import Endpoint
+    from dmr.serializer import BaseSerializer
 
 _ThingT = TypeVar('_ThingT')
 
+#: Value we compare a response header against: an ``int`` (compared as its
+#: ``str`` form), a plain ``str``, or any object with a custom ``__eq__``
+#: such as ``dirty_equals.IsStr`` for real-time (non-frozen) assertions.
+_HeaderValue = int | str | object
```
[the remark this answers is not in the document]
sesav: Protocol is not an option. I tested it. This would represent a regression in ergonomics, not an improvement.

dmr/test.py
```diff
@@ -179,3 +191,184 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+def _assert_header(
+    headers: Any,
+    name: str,
+    expected: _HeaderValue | None,
+) -> None:
+    if expected is None:
+        return
+    if isinstance(expected, int):
+        expected = str(expected)
+    assert headers[name] == expected, {name: headers.get(name)}  # noqa: S101
```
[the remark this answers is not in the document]
sesav: Okay, I'll make the fix and add a test.

dmr/test.py
```diff
@@ -179,3 +191,188 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+
+def _assert_header(
+    headers: Any,
+    name: str,
+    expected: _HeaderValue | None,
+) -> None:
+    if expected is None:
+        return
+    if isinstance(expected, int):
+        expected = str(expected)
```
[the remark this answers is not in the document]
sesav: Copilot's described behavior above is extremely odd. This issue occurs only in Python versions 3.10 and earlier, and this project does not support those versions.

tests/test_unit/test_throttling/test_throttling_helpers.py
```diff
@@ -0,0 +1,320 @@
...
+    throttle_state(_SyncController).exhaust(dmr_rf.get('/whatever/'))
+
+    response = _SyncController.as_view()(dmr_rf.get('/whatever/'))
+    assert isinstance(response, HttpResponse)
+    assert_throttled(
+        response,
+        limit=5,
+        reset=Rate.minute,
+        retry_after=Rate.minute,
+    )
```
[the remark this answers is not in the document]
sesav: Just for context: bare `IsStr()` only checks present + is a string, and it would happily pass on `reset='abc'`. 

Only two tests assert time-dependent headers, and both are in this module (reset/retry_after), and for those, I already added freezer (latest commit). Non-flaky and asserts the exact 60, strictly stronger, at the cost of the freezer fixture.

sobolevn (commented):
I don't think that this approach is correct :(
Because it triggers real amount of requests for backend. Imagine that there's a limit of 500/m, it will create 500 requests to increment the state.

I am 100% sure that we need to mock the number of requests.

I see this as:
- We get the endpoint to test
- We ensure that there's at least one throttling instance
- We change the first's throttling rate to be the `max_requests` which should be 2 by default
- We recalculate everything
- We run the checks with proper body / headers / etc.

Do you agree? :)

sesav:
> I don't think that this approach is correct :( Because it triggers real amount of requests for backend. Imagine that there's a limit of 500/m, it will create 500 requests to increment the state. 

Yes, this represents a real number of requests, but django-cache seed writes the saturated blob in a single cache write. It does not make 500 requests on the default backend. 

Anyway, my strategy is no good. I get it, and I'm trying to look at it differently. 


> I am 100% sure that we need to mock the number of requests.
> 
> I see this as:
> 
> * We get the endpoint to test
> * We ensure that there's at least one throttling instance
> * We change the first's throttling rate to be the `max_requests` which should be 2 by default
> * We recalculate everything
> * We run the checks with proper body / headers / etc.
> 
> Do you agree? :)

Yeah, I'm with you. Working on the code and your feedback have improved my understanding of this idea.

I will adjust this PR a bit..

dmr/throttling/base.py
```diff
@@ -152,6 +152,25 @@ def full_cache_key(
             f'{self.max_requests}::{self.duration_in_seconds}'
         )
 
+    def replace(self, *, max_requests: int) -> Self:
```
sobolevn: Can we better convert types to dataclasses? This way it will support `dataclasses.replace` and `copy.replace` automatically.

dmr/test.py
```diff
@@ -1,13 +1,26 @@
...
+if TYPE_CHECKING:
+    from dmr.controller import Controller
+    from dmr.serializer import BaseSerializer
 
 _ThingT = TypeVar('_ThingT')
 
+#: Value we compare a response header against: an ``int`` (compared as its
+#: ``str`` form), a plain ``str``, or any object with a custom ``__eq__``
+#: such as ``dirty_equals.IsStr`` for real-time (non-frozen) assertions.
+_HeaderValue = int | str | object
```
sobolevn: Please, use explicit type alias.
And `int | str | object` is the same as `object`

dmr/test.py
```diff
@@ -179,3 +192,202 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
     Uses ``async`` API.
     Requires you to ``await`` calls to ``.get``, ``.post``, etc.
     """
+
+
+@contextlib.contextmanager
+def reduced_throttling(
+    controller_cls: 'type[Controller[BaseSerializer]]',
+    *,
+    method: str = 'GET',
```
sobolevn: Maybe better `HTTPMethod`?

dmr/test.py
```diff
@@ -179,3 +192,202 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+        raise ValueError(
+            f'Endpoint {metadata.operation_id} has no throttling to test',
+        )
+
+    original = throttles[0]
+    reduced = original.replace(max_requests=max_requests)
+    # Swap the throttle in a fresh metadata; restored in `finally` below:
+    endpoint.metadata = dataclasses.replace(
+        metadata,
+        throttling_before_auth=_swap(
```
sobolevn: This should be a onfiguration. Which types we want to check. Any? Before? After?

dmr/test.py
```diff
@@ -179,3 +192,202 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+    Header expectations accept an ``int`` (matched as its string form), a
+    ``str``, or any matcher object (e.g. ``dirty_equals.IsStr()``) for
+    real-time tests where the exact reset value is unknown.
+
+    .. versionadded:: 0.12.0
+    """
+    assert response.status_code == HTTPStatus.TOO_MANY_REQUESTS, (
+        response.content
+    )
+    _assert_header(response.headers, 'X-RateLimit-Limit', limit)
```
sobolevn: This is not correct. Because different reports can produce different results. Better reuse them :)

dmr/test.py
```diff
@@ -179,3 +192,202 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+    assert isinstance(candidate, HttpResponse), candidate
+    return candidate
+
+
+def _assert_ok(candidate: HttpResponseBase) -> None:
+    response = _as_response(candidate)
+    assert response.status_code == HTTPStatus.OK, response.content
+
+
+def assert_throttling(
```
sobolevn: First: put all public function, then all protected ones.

dmr/test.py
```diff
@@ -179,3 +192,202 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+        max_requests=max_requests,
+    ):
+        for _ in range(max_requests):
+            _assert_ok(view(build(path)))
+        throttled = _as_response(view(build(path)))
+    assert_throttled(throttled, limit=limit)
+    return throttled
+
+
+async def aassert_throttling(
```
sobolevn:
```suggestion
async def assert_async_throttling(
```

docs/examples/throttling/testing.py
```diff
@@ -0,0 +1,41 @@
...
+from django.http import HttpResponse
+from django.test import TestCase
+from typing_extensions import override
+
+from dmr import Controller
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.test import DMRRequestFactory, assert_throttling
+from dmr.throttling import Rate, SyncThrottle
+
+_URL = '/reports/'
```
sobolevn:
```suggestion
_URL: Final = '/reports/'
```

docs/pages/throttling.rst
```diff
@@ -539,6 +539,42 @@ This way both your successful responses
...
+  :language: python
+
+Only the endpoint under test is affected; its throttling is restored afterwards.
+Use ``aassert_throttling`` for async controllers.
+
+For finer control -- custom bodies or auth, or asserting exact headers -- use
+the underlying ``reduced_throttling`` context manager and pair it with
+``assert_throttled``:
+
+.. code-block:: python
```
sobolevn: We generally prefer not to have any non-runnable code in the docs :)

dmr/throttling/base.py
sesav: We could, and I thought about it, but is a real refactor of core production code, not a drop-in, and I'd keep it out of this test-helpers PR to keep it simpler. And I noticed that `copy.replace` [was introduced](https://docs.python.org/3.13/library/copy.html#copy.replace) only in python 3.13 and calls `__replace__`. The project support python 3.11 and later, where `copy.replace` is not yet available, so even a complete transition to dataclasses won't provide `copy.replace` across all supported versions. While I believe this is not a big deal and I implemented `replace` already, but I'm bringing it to your attention just in case.

But that's not all there is to it. I mean, init params != stored fields. The constructor takes `backend`, `algorithm`, `response_headers`, but stores them as `_backend`, `_algorithm`, `_response_headers` (private). Dataclass fields are the init params, so `dataclasses.replace(t, max_requests=2)` would round-trip through `backend=…`, not `_backend=….` We'd have to rename the private attributes to public (or add init aliases), which ripples through the whole throttling package.

And also  `__init__` does non-trivial resolution that dataclass field defaults can't express directly, or I'm not sure how it works. And finnaly, it's `Generic[_BackendT]` with `__slots__` and two subclasses all doable as dataclasses, but it adds up.

Let's build a prototype of the dataclass version in a separate branch for comparison and decision-making because it touches equality/repr/hashing and the private-field naming across the package. And all of this seems to go beyond the scope of the issue we're currently working on. 🥲

dmr/test.py
sesav: Okay, I will fix it. 

A hint within the annotation itself seemed like a good idea to me.

dmr/test.py
sesav: Got it, name's changed and docs are updated.

dmr/test.py
sesav: Thanks for the tip, I updated it everywhere. :ok_hand:

dmr/throttling/base.py
```diff
@@ -152,6 +152,25 @@ def full_cache_key(
...
+        .. versionadded:: 0.12.0
+        """
+        return type(self)(
+            max_requests,
+            self.duration_in_seconds,
+            cache_key=self.cache_key,
+            backend=self._backend,
+            algorithm=self._algorithm,
+            response_headers=self._response_headers,
+        )
```
sobolevn, line 189: Ok, at least provide `__replace__` alias for `copy.replace` support. 

And support all parameters as kw-only.

dmr/test.py
```diff
@@ -1,13 +1,28 @@
...
+    from dmr.serializer import BaseSerializer
 
 _ThingT = TypeVar('_ThingT')
 
+#: Expected value for a header checked by ``assert_throttled``: an ``int``
+#: (rendered to its ``str`` form), a ``str``, or a matcher like
+#: ``dirty_equals.IsStr``. Typed as ``object`` because the header is compared
+#: with ``==``, so no narrower static type is correct -- matchers like
+#: ``IsStr`` are deliberately unhashable, so a ``Protocol`` won't help.
+_HeaderValue = object
```
sobolevn:
```suggestion
_HeaderValue: TypeAlias = object
```

dmr/test.py
sesav: That's a valid point. I added a `line` selector and a case statement.

dmr/test.py
```diff
@@ -179,3 +194,214 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+            reduced,
+        ),
+    )
+    try:
+        yield reduced
+    finally:
+        endpoint.metadata = metadata
+
+
+def _swap(
```
sobolevn: Please, move all protected functions to the bottom of the file.

sobolevn (commented):
Please, also re-export these as fixtures in `dmr_pytest.py`

dmr/test.py
sesav: Done!

dmr/throttling/base.py
sesav: Okay. I added `__replace__` alias and rebuilt all parameters as kw-only.

dmr/test.py
sesav: Done!

dmr/test.py
sesav: Fixed.

sesav:
@sobolevn I think the pull request is ready. Please check it when you have a moment. 

I apologize for the Copilot reviews.

dmr/test.py
```diff
@@ -179,3 +194,223 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+    ``str``, or any matcher object (e.g. ``dirty_equals.IsStr()``) for
+    real-time tests where the exact reset value is unknown.
+
+    .. versionadded:: 0.12.0
+    """
+    assert response.status_code == HTTPStatus.TOO_MANY_REQUESTS, (
+        response.content
+    )
+    headers = response.headers
+    _assert_header(headers, 'X-RateLimit-Limit', limit)
```
sobolevn: Please test this assert with all possible header formats :)

docs/pages/throttling.rst
```diff
@@ -539,6 +539,32 @@ This way both your successful responses
 and error responses will have the needed ratelimiting headers.
 
 
+Testing
```
sobolevn, line 542: I would just leave a not here with a `:ref:` to the testing page, where all this information must be located ;)

tests/test_unit/test_throttling/test_throttling_helpers.py
```diff
@@ -0,0 +1,298 @@
...
+    assert_throttling,
+    reduced_throttling,
+)
+from dmr.throttling import AsyncThrottle, Rate, SyncThrottle
+from dmr.throttling.algorithms import LeakyBucket
+from dmr.throttling.backends import SyncDjangoCache
+from dmr.throttling.cache_keys import RemoteAddr
+from dmr.throttling.headers import RateLimitIETFDraft
+
+_URL = '/whatever/'
```
sobolevn:
```suggestion
_URL: Final = '/whatever/'
```

tests/test_unit/test_throttling/test_throttling_helpers.py
```diff
@@ -0,0 +1,298 @@
...
+    with reduced_throttling(_SyncController) as throttle:
+        assert throttle.max_requests == 2
+        for _ in range(throttle.max_requests):
+            ok = view(dmr_rf.get(_URL))
+            assert isinstance(ok, HttpResponse)
+            assert ok.status_code == HTTPStatus.OK, ok.content
+        throttled = view(dmr_rf.get(_URL))
+
+    assert isinstance(throttled, HttpResponse)
+    assert_throttled(
```
sobolevn: Please, test pytest fixtures directly, not functions.

tests/test_unit/test_throttling/test_throttling_helpers.py
```diff
@@ -0,0 +1,298 @@
...
+    def get(self) -> str:
+        return 'inside'
+
+
+def _throttled_response(
+    controller_cls: type[Controller[PydanticFastSerializer]],
+    dmr_rf: DMRRequestFactory,
+) -> HttpResponse:
+    view = controller_cls.as_view()
+    with reduced_throttling(controller_cls) as throttle:
```
sobolevn: Why don't you use `_throttled_response` here in other similar tests?

dmr_pytest/__init__.py
```diff
@@ -59,6 +64,40 @@ def dmr_async_rf() -> 'DMRAsyncRequestFactory':
...
+
+    return reduced_throttling
+
+
+@pytest.fixture
+def dmr_assert_throttled() -> 'Callable[..., None]':
+    """Provides :func:`dmr.test.assert_throttled`."""
+    from dmr.test import assert_throttled
+
+    return assert_throttled
```
sobolevn: Do we need to expose this as a fixture?

sobolevn (commented):
Two more comments. 
1. We can split `test.py` into a package now
2. Please, don't forget to update the docs :)

dmr/test.py
sesav: Uh, I believed you requested that the value not be reused. :smiling_face_with_tear:  Ok. I've included new tests using parametrization.

docs/pages/throttling.rst
sesav: Absolutely. Moved.

tests/test_unit/test_throttling/test_throttling_helpers.py
sesav: Done!

tests/test_unit/test_throttling/test_throttling_helpers.py
sesav: Yep, I added them before they went into the fixtures.

tests/test_unit/test_throttling/test_throttling_helpers.py
sesav: Well spotted. The preview implementation wasn't updated because I added tests, reworked them, and then added new ones.

dmr_pytest/__init__.py
sesav: Yes, I'd keep it as a fixture for a consistent, discoverable `dmr_*` testing surface, and it's exactly what you asked for, right?

sesav:
@sobolevn The pull request looks ready to me. Check it out when you have time.

I've updated the PR description to reflect the revisions and fixes, and we should also remember to create a new task for converting `ResponseSpecProvider` or `_BaseThrottle` to a new dataclass.

dmr/test.py
```diff
@@ -179,3 +194,223 @@ class DMRAsyncClient(_DMRMixin, AsyncClient):
...
+        case _:
+            throttles = metadata.throttling
+    if not throttles:
+        raise ValueError(
+            f'Endpoint {metadata.operation_id} has no throttling '
+            f'to test for line {line!r}',
+        )
+
+    original = throttles[0]
+    reduced = original.replace(max_requests=max_requests)
```
sobolevn: Please, also replace the time scale to `Hour`

docs/pages/throttling.rst
```diff
@@ -539,6 +539,16 @@ This way both your successful responses
...
+-------
+
+.. note::
+
+  ``dmr.test`` ships helpers to drive an endpoint to its limit and assert the
```
sobolevn:
```suggestion
  ``dmr.test`` and our ``pytest`` plugin 
   ship helpers to drive an endpoint to its limit and assert the
```

tests/test_unit/test_throttling/test_throttling_helpers.py
```diff
@@ -0,0 +1,368 @@
...
+
+# Types of the `dmr.test` helpers as provided by their pytest fixtures:
+_ReducedThrottling: TypeAlias = Callable[
+    ...,
+    contextlib.AbstractContextManager[SyncThrottle | AsyncThrottle],
+]
+_AssertThrottled: TypeAlias = Callable[..., None]
+_AssertThrottling: TypeAlias = Callable[..., HttpResponse]
+_AssertAsyncThrottling: TypeAlias = Callable[..., Awaitable[HttpResponse]]
+_ThrottledResponse: TypeAlias = Callable[..., HttpResponse]
```
sobolevn: These aliases must be public in our pytest plugin, so users can import them.

tests/test_unit/test_throttling/test_throttling_helpers.py
```diff
@@ -0,0 +1,368 @@
...
+    ('reset', 'X-RateLimit-Reset'),
+    ('retry_after', 'Retry-After'),
+)
+
+# Types of the `dmr.test` helpers as provided by their pytest fixtures:
+_ReducedThrottling: TypeAlias = Callable[
+    ...,
+    contextlib.AbstractContextManager[SyncThrottle | AsyncThrottle],
+]
+_AssertThrottled: TypeAlias = Callable[..., None]
```
sobolevn: Make sure that types are actually accurate, not just `...`, we enforce strict type-safety :)

sobolevn (commented):
- `assert_throttled` is not tested with all possible response headers, it just hardcodes specific headers
- `test.py` is not split into a package as I proposed in the previous review :( 
- Please, add a note about time freezing / time gaps in such cases. What is important? It won't really work on `2/s`, we should decrease the scale as well
- Docs do not have a note about `pytest` fixtures, there's no example on how to use it with `pytest`. I propose adding a `.. tab:` with `unittest` and `pytest` tabs with proper examples

sesav:
@sobolevn It's all good to go. 

I've updated the feature based on your feedback, focusing on a more streamlined design.

```python
assert_throttled(response, *, throttle=None, headers=None, detail=True)
assert_throttling(Controller, rf, path, *, method=HTTPMethod.GET, max_requests=2, line='any', headers=None, detail=True)
assert_async_throttling(...)  # same, async
reduced_throttling(Controller, *, method=HTTPMethod.GET, max_requests=2, line='any')
```

- PR description has been updated;

- No hardcoded header names: given the `throttle`, `assert_throttled` asks its providers what they report and checks that all of them are present, so `XRateLimit`, `RateLimitIETFDraft` and custom providers work alike. The drivers pass the throttle themselves. `limit` / `reset` / `retry_after` are gone, expected values go into `headers` as an `int`, a `str`, or a matcher;

- `reduced_throttling` also widens the window to `Rate.hour`, so `2/second` endpoints became testable, this is the `time gaps` point from your comment. Note that `X-RateLimit-Reset` and `Retry-After` now report that window. The docs note keeps the time-freezing advice for the case where you assert those two headers exactly;

- Fixture types are public protocols in `dmr_pytest` with full signatures, pinned by `typesafety/assert_type/test_pytest_plugin.py` (the same way as the existing ones);

- Docs: `unittest` / `pytest` tabs, both examples runnable via `just example`.


**In existing code**

- `dmr/test.py` split into a package (client code moved verbatim);

- `dmr_pytest.py` became a package with `py.typed` without the marker downstream mypy reports `import-untyped`, which makes publishing the protocols pointless;

- `throttling/base.py` got a `response_headers` property, needed for the check above.

PR #1182, opened 2026-07-30
fix: use Generator return type for contextmanager helpers

### LINKED ISSUE #1181: Fix all `@contextmanager` entries with `Iterator` return type
Using `@contextmanager` with `Iterator` return type is deprecated.
It must use `Generator` return type.

Image

We need to fix all places where we use `-> Iterator` together with `contextmanager` decorator.

arimu1:
## Description

`@contextmanager` helpers annotated with `Iterator[...]` return types are deprecated.
Update them to `Generator[...]` to match typing guidance and the existing pattern in
`tests/test_unit/test_compiled/test_negotiate_compiled.py`.

## Changes

- `benchmarks/tests/test_compiled/conftest.py`
- `docs/tools/sphinx_ext/run_examples.py`

Fixes #1181

## Verification

Scanned the tree for `@contextmanager` + `-> Iterator[` — none remaining.

### DIFF
--- a/benchmarks/tests/test_compiled/conftest.py
+++ b/benchmarks/tests/test_compiled/conftest.py
@@ -1,5 +1,5 @@
 import sys
-from collections.abc import Callable, Iterator, Set
+from collections.abc import Callable, Generator, Set
 from contextlib import AbstractContextManager, contextmanager
 from types import ModuleType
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thank you!

Issue #1193, opened 2026-08-01
Support migration from `dj-rest-auth`


sobolevn:
This is a critical component in DRF's world: https://github.com/iMerica/dj-rest-auth
We need to provide a clean migration path from `dj-rest-auth` to `django-modern-rest`.

Without this component many project's can really migrate, because they would have to re-invent the most critical part: auth. This is not something we can allow for our users.

How?

- [x] Support default login views
- [x] Support registration default views
- [x] Support password change default views
- [x] Support OAuth flow #1012 
- [x] Support MFA
- [x] Support Passkeys
- [x] Support cookie-based JWT auth flow (currently, we only support headers)
- [x] Support AI skill for migration from `dj-rest-auth` to `django-modern-rest`

For all features we would require:
- Strict AI review for the best practices of the method (because I am not an expert in several things here, so we need someone else's opinion), security folks are welcome!
- Docs 
- Schemes, how everything works, like `dj-rest-auth` has
- Lots of tests :) 

This is a big project, so any help / contributions are welcome! 
Let's do it together, friends :)

Lancetnik:
Assign me please

sobolevn:
😆 

https://www.youtube.com/watch?v=laJQcNAqxc0

kstypik:
Do we need to reinvent this? 

`dj-rest-auth` relies on django-allauth for some of its features. Allauth has all of the features required (excluding an AI skill which is irrelevant for the package) and has its own `headless` mode for use in APIs nowadays that's not coupled with neither DRF nor django-ninja, because it uses custom django views (see: https://docs.allauth.org/en/latest/headless/index.html )

Considering you don't feel confident in this field, maybe adding a slim integration of `django-allauth`'s headless mode into `django-modern-rest` is a better idea? For example `django-allauth` generates its own OpenAPI schema, which we could integrate into the one generated by `dmr`.

`django-allauth` is a well-established, full-featured, well-maintained library. This task looks like a monumental work to be done.

sobolevn:
Yes, sorry for not being explicit about this. Thin wrapper around `django-allauth` is what I want :)

kstypik:
First thing that comes to my mind if these are pre-defined views: how should the data be serialized if `dmr` supports pluggable serializers? Configurable, dataclasses or maybe something else?

In my new project I'm using django-modern-rest for the API and pure headless mode from django-allauth with session cookies.

sobolevn:
[...]

PR #1198, opened 2026-08-02
Document plain Django views in the OpenAPI specification

Lancetnik:
# I have made things!

A plain Django view can now be documented in the generated OpenAPI
specification without being rewritten as a controller. This is the piece
needed to describe endpoints the framework does not own — the immediate
motivation being allauth's headless endpoints, but it stands on its own for
anyone migrating to `dmr` incrementally.

Three commits, each one leaving the tree green:

1. **`adapt_django_view`** returns a subclass of any Django view that the
   schema collector can consume. The path item is supplied statically, either
   as a `PathItem` or as a plain mapping, so a fragment of an existing
   document can be pasted in without translating it to typed objects.

2. **A component contribution channel.** Any imported third-party description
   references shared components, and a path item referencing them produced
   dangling references before this. An adapted view can now contribute
   `schemas`, `responses`, `parameters`, `examples` and `requestBodies`
   through the `OpenAPIContext` during `get_path_item` — the same channel
   controllers already use to register schemas.

3. **Namespacing and collision reporting.** `component_prefix` namespaces the
   whole description: the component names and every reference to them, in the
   path item and nested inside other components, in both spellings. Names
   imported from a foreign document are rarely unique enough to share a
   document with the schemas the framework generates.

Two collisions were resolved silently before and now raise: a second,
*different* definition of a name already contributed, and a contributed schema
clashing with a generated one (previously the generated schema won, leaving the
adapted view's `$ref` pointing at a different definition than intended).
Re-contributing an *identical* definition stays legal, because one description
is commonly shared by several adapted views.

A discriminator's `mapping` is handled specially — it is the one place a
reference is not spelled as a `$ref`, so a rename would otherwise leave it
pointing at a name that no longer exists, without tripping schema validation.

Adapting a view is deliberately documentation-only: it does **not** enrol the
view into the request pipeline. No parsers, renderers, negotiation, problem
details, throttling, response validation or controller-level auth apply, and
its errors keep the shape Django gives them. That is the design, not an
omission, and it is called out in a `Warning::` block on the function.

## AI Policy

## Checklist

Tests live in `tests/test_unit/test_adapters/`, following the existing
schema-snapshot suite: the views and the router are declared in the test
module and the assertions are on the document that comes out, not on the
adapter's shape or on registry internals. `docs/pages/deep-dive/public-api.rst`
picks the new API up through autodoc.

## Related issues

- Refs #1193

Two pre-existing defects were found while building this and are **not**
addressed here; they deserve their own issues:

- `SchemaRegistry` silently discards a newly registered schema on a name
  collision when no annotation is supplied. Contributed components never pass
  through it, so this work sidesteps rather than fixes it.
- The demo project's URL configuration builds the specification before
  assigning url patterns, so any URL reversing during the build fails.
  Harmless today, but it constrains what the build is allowed to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ of requirements for an API to count as public.
 - Added `dmr.adapters.adapt_django_view` to document plain Django views
   in the generated OpenAPI schema without rewriting them
   as controllers, #1193
+- Added the `components` argument to `dmr.adapters.adapt_django_view`,
+  so an adapted view can contribute the reusable `schemas`, `responses`,
+  `parameters`, `examples` and `requestBodies` its path item
+  references, #1193
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,12 @@ of requirements for an API to count as public.
   so an adapted view can contribute the reusable `schemas`, `responses`,
   `parameters`, `examples` and `requestBodies` its path item
   references, #1193
+- Added the `component_prefix` argument to `dmr.adapters.adapt_django_view`,
+  which namespaces the components a view contributes and every reference
+  to them, so names imported from a foreign specification cannot displace
+  the project's own, #1193
+- Contributing two different definitions of one component name now raises,
+  instead of silently keeping one of them, #1193
 
 ### Bugfixes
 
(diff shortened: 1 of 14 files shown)


sobolevn:
It is now supported: https://django-modern-rest.readthedocs.io/en/latest/pages/external-views.html

PR #1200, opened 2026-08-03
Add pytest-timeout to fail hanging tests fast (#1195)

### LINKED ISSUE #1195: Add `pytest-timeout` to fail hanging tests fast
# Feature

## Thesis

Add [`pytest-timeout`](https://pypi.org/project/pytest-timeout/) to the test dependencies and enable a global per-test timeout, so a hanging test fails fast and names itself instead of stalling the job.

```toml
[dependency-groups]
unit-test = [
  # ...
  "pytest-timeout>=2.4,<3",
]

[tool.pytest.ini_options]
# Fail hanging tests instead of blocking CI:
timeout = 30
timeout_method = "thread"
```

Tests that legitimately need longer raise their own budget with the marker:

```python
@pytest.mark.timeout(120)
def test_something_genuinely_slow() -> None: ...
```

## Reasoning

- **Nothing currently bounds a hang.** Only `.github/workflows/relator.yml` sets `timeout-minutes`; `test.yml`, `test-extras.yml`, `build-wheels.yml` and `codspeed.yml` do not, so a single deadlocked test burns the GitHub Actions default of **6 hours per job** — multiplied across the Python/Django matrix.
- **The suite has real hang-prone surface.** `asyncio_mode = 'auto'` means every async test can deadlock on an unawaited future; `tests/test_integration/test_throttling/test_backends/test_redis_backend/` talks to a live Redis/Valkey; the streaming tests consume generators; the smoke tests spawn subprocesses. Today a wedged backend looks like "CI is slow", not "CI is broken".
- **Attribution.** A job-level `timeout-minutes` kills the run without telling you which test hung. `pytest-timeout` dumps the stack of the offending test and keeps going, which is the difference between a five-minute fix and a bisect.
- **Locally too.** The same protection applies to `just unit`, where a hang currently requires a manual Ctrl-C and offers no traceback.
- **Cheap.** One pure-Python dev dependency, no runtime impact on the shipped package.

### Notes / open questions

- It has to go in the `unit-test` group specifically: that group is the minimal one installed by `cibuildwheel`, and with `--strict-config` in `addopts` an unknown `timeout` ini key would make pytest error out anywhere the plugin isn't installed.
- `timeout_method = "thread"` behaves the same on every platform (relevant for the Windows wheel jobs); `signal` gives a slightly nicer traceback but is POSIX-only. Happy to go with either.
- The `30`s default is a starting point — it should be set above the slowest current test so the change is a no-op for a green suite. Worth a `--durations` run to pick the number.

AsilbekKha:
# I have made things!

## AI Policy

## Checklist

## Related issues
Closes #1195
Added pytest-timeout as a dev dependency and set a global 5s timeout in
the pytest config. Went with the `thread` method since it behaves the
same on Windows and POSIX. Signal-based timeouts don't work the same
way on Windows, and this repo builds Windows wheels, so thread keeps
things consistent.

To pick the 5s number, I ran the full suite locally with `--durations=10`
first to see what "normal" looks like. The slowest regular test was
under 3 seconds, so 5s still leaves some margin, though less than the
30s I originally had. The maintainer suggested 5s directly, which lines
up with what I saw in practice. The one exception is `test_schemathesis`,
which is an aggregate fuzz test and legitimately runs 25-30s, so I gave
that one its own `@pytest.mark.timeout(60)` override instead of letting
it hit the global limit.

Ran `just test` locally (lint, type-check, unit) and everything passes
clean, coverage's still at 100%.

Didn't add a new test for this since it's more of a CI/config change
than new behavior. Also didn't update the docs, since this only affects the internal test suite and not anything a library user would interact with.

### DIFF
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -1,7 +1,9 @@
 import pytest
 
 
-def pytest_collection_modifyitems(session: pytest.Session, config: pytest.Config, items: list[pytest.Item]) -> None:
+def pytest_collection_modifyitems(
+    session: pytest.Session, config: pytest.Config, items: list[pytest.Item]
+) -> None:
     """Automatically disable timeout for all of benchmarks"""
     for item in items:
         item.add_marker(pytest.mark.timeout(0))
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -1,0 +1,6 @@
+import pytest
+
+def pytest_collection_modifyitems(items):
+    """Automatically disable timeout for all of benchmarks"""
+    for item in items:
+        item.add_marker(pytest.mark.timeout(0))
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -1,5 +1,6 @@
 import pytest
 
+
 def pytest_collection_modifyitems(items):
     """Automatically disable timeout for all of benchmarks"""
     for item in items:
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -1,7 +1,7 @@
 import pytest
 
 
-def pytest_collection_modifyitems(items):
+def pytest_collection_modifyitems(session: pytest.Session, config: pytest.Config, items: list[pytest.Item]) -> None:
     """Automatically disable timeout for all of benchmarks"""
     for item in items:
         item.add_marker(pytest.mark.timeout(0))
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -2,7 +2,9 @@ import pytest
 
 
 def pytest_collection_modifyitems(
-    session: pytest.Session, config: pytest.Config, items: list[pytest.Item]
+    session: pytest.Session,
+    config: pytest.Config,
+    items: list[pytest.Item],
 ) -> None:
     """Automatically disable timeout for all of benchmarks"""
     for item in items:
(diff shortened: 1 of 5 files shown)


sobolevn (approved):
Great! Thank you!

AsilbekKha:
Saw the errors, trying to fix them. I'll notify you when im done with commits

AsilbekKha:
I've disabled the timeout for test_schematesis using pytest.mark.timeout(0) since fuzzing tests can naturally take longer than the strict global limit.

For the CodSpeed benchmarks, what is the preferred approach here? Should I add pytest.mark.timeout(0) to the benchmark tests as well to exclude them from the timeout?

sobolevn:
Yes, please disable benchmarks from this rule.

benchmarks/tests/test_router/test_path.py
```diff
@@ -10,6 +11,8 @@
 
 from dmr.routing import path as dmr_path
 
+pytestmark = pytest.mark.timeout(0)
```
sobolevn: please, ignore all benchmarks, not only this module

AsilbekKha: Got it!  Let me handle the whole benchmarks/ folder globally in a conftest.py hook instead so we don't have to touch every file individually. I'll update it now.

benchmarks/conftest.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+def pytest_collection_modifyitems(items):
```
sobolevn:
```suggestion
def pytest_collection_modifyitems(session: pytest.Session, config: pytest.Config, items: list[pytest.Item]) -> None:
```

sobolevn (approved):

sobolevn:
Awesome work, it should be good now :)

PR #1230, opened 2026-08-09
Fix \OpenAPIFormat.IRI\ value

### LINKED ISSUE #1226: Fix `OpenAPIFormat.IRI` value
See the spec https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.3.5

It must be `'iri'`, not `'iri-reference'`: https://github.com/wemake-services/django-modern-rest/blob/ea7d1d008383de9c976cf4db4b2a67934a7fe320/dmr/openapi/objects/enums.py#L33-L34

We also must mark these enums as `@enums.unique`

ChickenisLegit:
Fixes #1226

### DIFF
--- a/dmr/openapi/objects/enums.py
+++ b/dmr/openapi/objects/enums.py
@@ -30,7 +30,7 @@ class OpenAPIFormat(enum.StrEnum):
     URI_TEMPLATE = 'uri-template'
     JSON_POINTER = 'json-pointer'
     RELATIVE_JSON_POINTER = 'relative-json-pointer'
-    IRI = 'iri-reference'
+    IRI = 'iri'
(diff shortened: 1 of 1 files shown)


dmr/openapi/objects/enums.py
sobolevn, line 5:
```suggestion
@final
@enum.unique
```

sobolevn, line 49:
```suggestion
@enum.unique
class OpenAPIType(enum.StrEnum):
```

sobolevn (approved):
Thank you!

sobolevn:
Oups, https://github.com/wemake-services/django-modern-rest/pull/1228 was first :( 
Sorry!

PR #1241, opened 2026-08-10
Add missing JSON Schema keywords to Schema

### LINKED ISSUE #1232: Fix missing `Schema` fields from the spec
We currently miss these fields from `Schema` type definition: https://github.com/wemake-services/django-modern-rest/blob/03e7b141878dc35692f364558c4b349dce1f4b4f/dmr/openapi/objects/schema.py#L17-L18

- `$ref` — https://json-schema.org/draft/2020-12/json-schema-core#section-8.2.3.1
"$ref" — References another schema by URI.

- `$anchor` — https://json-schema.org/draft/2020-12/json-schema-core#section-8.2.2
"$anchor" — Defines a plain-name fragment anchor within a schema resource.

- `$comment` — https://json-schema.org/draft/2020-12/json-schema-core#section-8.3
"$comment" — A developer comment; has no effect on validation.

- `$schema` — https://json-schema.org/draft/2020-12/json-schema-core#section-8.1.1
"$schema" — Declares which JSON Schema dialect the schema is written against.

It would require a CHANGELOG entry and a test case, since we don't use them as of right now.

ShlokShar:
Fixes #1232

Add the missing `$ref`, `$anchor`, `$comment`, and `$schema` fields to `Schema`, plus regression coverage and a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ of requirements for an API to count as public.
 
 - Fixed a bug that `OpenAPIConfig.components` were silently
   ignored when defined with custom user's data, #1229
+- Fixed missing `$ref`, `$anchor`, `$comment`, and `$schema` fields in `Schema`, #1232
 - Fixed `OpenAPIFormat.IRI` value, #1228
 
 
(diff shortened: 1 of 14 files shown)


sobolevn (approved):
Thanks!

PR #1256, opened 2026-08-12
ci: remove unnecessary tomli from codespell additional_dependencies

### LINKED ISSUE #1255: Remove `tomli` from`additional_dependencies` in pre-commit
We don't need it, since we run py3.11+ with `tomllib`: https://github.com/wemake-services/django-modern-rest/blob/d1f2b6c3d56bb64a4d6fe9f8349209c4663d68bf/.pre-commit-config.yaml#L61-L62

Pavankalyan32:
Python 3.11+ includes `tomllib` in the standard library (PEP 680), making the `tomli` backport redundant for this project which requires Python ">=3.11" (see `pyproject.toml:20`).

Removing `tomli` from `additional_dependencies` in the codespell pre-commit hook simplifies the dependency chain and avoids installing an unnecessary package.

Fixes #1255


sobolevn:
Sorry, but @danfimov was first :)
You can work on other issues, we have a lot of things to help with.

PR #1261, opened 2026-08-12
feat: add ty type checker support (#1257)

### LINKED ISSUE #1257: Support `ty` for type-checking
Very interesting

Things to do:
- Install specific `ty` version
- Enable its strict mode
- Type-check `dmr/` and `typesafety/` directories
- Fix all real problems found by it
- Ignore all false-positives
- Add a changelog entry
- Update list of support type-checkers in README.md

If it produces way to many errors for one specific problem: disable it in config instead.

Pavankalyan32:
Adds support for the  type checker as requested in #1257.

Changes:
- Add  to the  dependency group
- Update README to include  in the list of supported type checkers
- Add changelog entry for 0.13.1

Note:  currently produces many false-positives with the Django codebase (1700+ errors), so it's added as an optional manual check rather than a CI gate. Run with:
```bash
ty check --project . dmr typesafety
```

Fixes #1257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,8 +47,14 @@ on the releases page: https://github.com/wemake-services/django-modern-rest/rele
 - Renamed `dmr.controller.Controller.get_path_item` to
   `get_schema`, so all methods will be consistent, #1238
 - Removed `dmr_assert_throttling` and `dmr_assert_async_throttling`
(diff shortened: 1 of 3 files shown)


sobolevn (commented):
This is not correct, you never actually run `ty`

PR #1274, opened 2026-08-14
fix: validate email format in UserCreateSchema

### LINKED ISSUE #945: UserCreateSchema.email accepts any string — no email format validation
### What's wrong?

The User model uses email = EmailField(unique=True), which implies a valid email address. The Pydantic schema does not enforce this:

```python
class UserCreateSchema(pydantic.BaseModel):
      email: str <!--------------------------------------HERE
```
Invalid email addresses pass Pydantic validation and either cause a database-level error (PostgreSQL does not validate email format) or get stored as garbage data. There is no user-facing 422 response  explaining the format requirement. 

### How it should be?

Use pydantic.EmailStr to validate email format at the serializer layer:
```python

from pydantic import EmailStr
                                                                                                                                                                                                            class UserCreateSchema(pydantic.BaseModel):
    email: EmailStr                                                                                                                                                                                       
    role: RoleSchema                                  
    tags: list[TagSchema]
```

### Used versions

0.7.0

### OS information

MacOS

aryansk:
## Problem

Closes #945. `UserCreateSchema.email` was typed as `str`, so malformed addresses passed Pydantic validation and either caused a database-level error (PostgreSQL does not validate email format) or were stored as garbage data, with no user-facing error explaining the format requirement.

## Change

Type `UserCreateSchema.email` as `pydantic.EmailStr` so email format is validated at the serializer layer and invalid input returns a proper 4xx with a clear message. No new dependencies: the existing `pydantic[email,timezone]` extra already ships `email-validator`.

## Why this approach

The Django `User` model already uses `models.EmailField(unique=True)`, so enforcing the same constraint in the schema mirrors the model contract at the API boundary, exactly as the issue proposes. The change is one type annotation plus a regression test.

## Testing

```text
uv run python -m pytest tests/test_integration/test_model_fk/test_model_fk.py -o addopts="" -q
result: 10 passed (2 new: invalid email rejected with value_error)

uv run python -m pytest tests/ -o addopts="" -q --ignore=.../redis_backend
result: 2803 passed, 11 skipped, 47 subtests passed

uv run python -m ruff check ... && uv run python -m ruff format --check
result: All checks passed
```

## Documentation and release impact

- [x] User-facing documentation updated (CHANGELOG entry)
- [ ] Changelog/release note needed
- [ ] Migration or compatibility note needed
- [ ] No documentation impact

## Review notes

- Known limitations: `SimpleUserCreateSchema` in the model_simple example keeps `email: str`; the issue scope is `UserCreateSchema` only, happy to extend if desired.
- Follow-up issue, if any: none
- Security/licensing considerations: none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@ Later on we will make the API more stable and decrease the amount
 of requirements for an API to count as public.
 
 
+## Unreleased
+
+### Bugfixes
+
+- `UserCreateSchema.email` now validates the email format via
+  `pydantic.EmailStr`, so malformed addresses fail at the serializer
+  layer instead of reaching the database, #945
+
+
 ## 0.14.0 (2026-08-14)
 
 ### Breaking changes
--- a/tests/test_integration/test_model_fk/test_model_fk.py
+++ b/tests/test_integration/test_model_fk/test_model_fk.py
@@ -98,6 +98,32 @@ def test_user_create_name_too_long_rejected(
 
 
 @pytest.mark.django_db
+def test_user_create_invalid_email_rejected(
+    dmr_client: DMRClient,
+    faker: Faker,
+) -> None:
+    """Invalid email addresses are rejected at the serializer layer.
+
+    Regression for #945: the Pydantic schema accepted any string, so
+    malformed emails passed validation and either failed at the database
+    layer with an unhandled error or were stored as garbage data.
+    """
+    response = dmr_client.post(
+        reverse('api:model_fk:user'),
+        data={
+            'email': 'not-an-email',
+            'role': {'name': faker.name()},
+            'tags': [{'name': faker.name()}],
+        },
+    )
+
+    assert response.status_code == HTTPStatus.BAD_REQUEST, response.content
+    payload = response.json()
+    assert payload['detail'][0]['type'] == 'value_error'
+    assert 'value is not a valid email address' in payload['detail'][0]['msg']
+
+
+@pytest.mark.django_db
 def test_user_create_unique_email_error(
     dmr_client: DMRClient,
     faker: Faker,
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,15 @@ Later on we will make the API more stable and decrease the amount
 of requirements for an API to count as public.
 
 
+## Unreleased
```
sobolevn, line 21: changelog entry is not needed, this is a test-only change.

tests/test_integration/test_model_fk/test_model_fk.py
```diff
@@ -97,6 +97,32 @@ def test_user_create_name_too_long_rejected(
...
+def test_user_create_invalid_email_rejected(
+    dmr_client: DMRClient,
+    faker: Faker,
+) -> None:
+    """Invalid email addresses are rejected at the serializer layer.
+
+    Regression for #945: the Pydantic schema accepted any string, so
+    malformed emails passed validation and either failed at the database
+    layer with an unhandled error or were stored as garbage data.
+    """
```
sobolevn, line 110:
```suggestion
    """Ensures that non-email fails the schema validation."""
```

tests/test_integration/test_model_fk/test_model_fk.py
```diff
@@ -97,6 +97,32 @@ def test_user_create_name_too_long_rejected(
...
+            'email': 'not-an-email',
+            'role': {'name': faker.name()},
+            'tags': [{'name': faker.name()}],
+        },
+    )
+
+    assert response.status_code == HTTPStatus.BAD_REQUEST, response.content
+    payload = response.json()
+    assert payload['detail'][0]['type'] == 'value_error'
+    assert 'value is not a valid email address' in payload['detail'][0]['msg']
```
sobolevn, line 123:
```suggestion
    assert payload == snapshot()
```

use `inline_snapshot` please, reuse parts of the passed request data.

PR #1282, opened 2026-08-19
docs: add django_modern_schemas in docs

Izcarmt95:
# I have made things!

Adds a `django-modern-schemas` section to the "Serializing models and Querysets"
page, next to the existing `django-mantle` one, for readers who want their
serializer schemas derived from the models they already have.

It links the library, lists what it does (schemas generated from the model,
`fields`/`exclude`/`optional` for read, create and `PATCH`, `Source` and
`MethodSource`, `ModelSchema` generic in its model, nothing in
`INSTALLED_APPS`), shows a `schemas.py` + `views.py` example with a
list/create controller and a single-object one plus the `curl` response, and
links a complete app built with both projects:
https://github.com/open-byte/dmr-with-django-modern-schemas

## AI Policy

## Checklist


docs/pages/queryset.rst
```diff
@@ -365,3 +365,126 @@ It’s your type-safe layer around Django’s liquid core.
...
+- ``ModelSchema`` is generic in its model, so ``.create()``, ``.update()``
+  and ``.save()`` are typed as returning that exact model.
+- Needs nothing in ``INSTALLED_APPS``: schemas are regular Python classes.
+
+Schemas are declared from the model, plus any extra fields they need:
+
+.. code-block:: python
+  :caption: schemas.py
+
+  from typing import Annotated
```
sobolevn: Please, use `.. literalinclude` from `examples/` folder.

You can add `django-modern-schemas` as `dev-and-docs` dependency.

We need this, so we can verify our example is actually correct :)

See similar example with `django-mantle`

Izcarmt95: Done!

pyproject.toml
```diff
@@ -111,6 +111,7 @@ dev = [
 docs = [
   "django-filter>=25.2,<27",
   "django-mantle>=26.6,<27",
+  "django-modern-schemas>=0.1.0,<0.2.0",
```
```suggestion
  "django-modern-schemas>=0.1.0,<0.2.0; python_version >= 3.14",
```

docs/pages/queryset.rst
```diff
@@ -365,3 +365,51 @@ It’s your type-safe layer around Django’s liquid core.
   :caption: views.py
   :language: python
   :linenos:
+
+
+django-modern-schemas
+---------------------
+
+If you want to automate the schema definition part and derive your
```
sobolevn, line 373: Please, add a note that it requires python 3.14+

sobolevn (commented):
please, also update `readthedocs.yml` to require python 3.14

sobolevn:
Oh, ok. I got what's wrong :(

```
# Capped below 3.14: PEP 649 defers class annotations, so the metaclass sees an empty
# `__annotations__` and silently drops Source/MethodSource fields. Lift once that is fixed.
requires-python = ">=3.10,<3.14"
```

Sorry, I can't allow this to be in the docs, while this problem is not solved.
Let's wait for the second release, when you will fix this problem.

Izcarmt95:
> Oh, ok. I got what's wrong :(
> 
> ```
> # Capped below 3.14: PEP 649 defers class annotations, so the metaclass sees an empty
> # `__annotations__` and silently drops Source/MethodSource fields. Lift once that is fixed.
> requires-python = ">=3.10,<3.14"
> ```
> 
> Sorry, I can't allow this to be in the docs, while this problem is not solved. Let's wait for the second release, when you will fix this problem.

Hi, today I published the new version , this version support python3.14+ !

Izcarmt95:
@sobolevn can you check this PR? 👀

sobolevn (approved):
Looks amazing!

PR #1284, opened 2026-08-20
Return 401 instead of 500 for unusable JWT subjects

vyhuholl:
# I have made things!

A signed token can carry anything in `sub`, but the user lookup field is typed. Looking up a non-numeric subject against the default integer `pk` made Django raise `ValueError`, which escaped as a 500:

    ValueError: Field 'id' expected a number but got 'admin'.

Only `ObjectDoesNotExist` was caught. Fields that validate on conversion, such as `UUIDField`, raise `ValidationError` and had the same problem.

All of these mean the same thing for auth: no such user. They are now collected in `USER_LOOKUP_ERRORS` and turned into `NotAuthenticatedError`, in the auth classes as well as in the refresh and verify controllers, which all shared the lookup.
## AI Policy

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@ Later on we will make the API more stable and decrease the amount
 of requirements for an API to count as public.
 
 
+## WIP
+
+### Bugfixes
+
+- JWT auth, refresh, and verify now return `401` instead of `500`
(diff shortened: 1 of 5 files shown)


sobolevn (approved):
Great catch! Thank you!

PR #1287, opened 2026-08-20
Add django-allauth headless session token auth

vyhuholl:
# I have made things!

Adds `XSessionTokenSyncAuth` and `XSessionTokenAsyncAuth`, which resolve the `X-Session-Token` header that `django-allauth`'s headless mode hands out into an authed `request.user`.

`django-allauth` keeps owning the whole login flow, we only consume its tokens. Available via the new `django-modern-rest[allauth]` extra.

No CSRF check here, unlike Django session auth: the token travels in a header, which browsers do not attach on their own.

`django-allauth` has no async API, so the async class does the lookup through `sync_to_async`. This is documented rather than hidden.

We call `allauth.headless.internal.sessionkit`, the same entry point allauth's own DRF and Ninja integrations use. It is not a public API contract, and the docs say so.
## AI Policy

## Checklist

## Related issues

Refs #1193


dmr/security/allauth/auth.py
```diff
@@ -0,0 +1,254 @@
...
+        Return the raw session token for the given request.
+
+        By default we look up the ``X-Session-Token`` header.
+        Override this if you want to extract the token from somewhere else,
+        for example from the ``Authorization`` header.
+        """
+        return request.headers.get(self.header_name)
+
+
+def _authenticate(token: str) -> 'tuple[AbstractBaseUser, SessionBase]':
```
sobolevn:
```suggestion
def _authenticate(token: str) -> tuple['AbstractBaseUser', 'SessionBase']:
```

docs/pages/auth/allauth.rst
sobolevn, line 1: Ideas about this file:
- We should provide an example of how to obtain auth using `external_path` and pre-existing headless views

tests/test_unit/test_security/test_allauth/test_allauth_schema.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+from inline_snapshot import snapshot
+
+from dmr.openapi.objects import SecurityScheme
+from dmr.security.allauth import XSessionTokenAsyncAuth, XSessionTokenSyncAuth
+
+_AuthType = type[XSessionTokenSyncAuth] | type[XSessionTokenAsyncAuth]
```
sobolevn:
```suggestion
_AuthType: TypeAlias = type[XSessionTokenSyncAuth] | type[XSessionTokenAsyncAuth]
```

tests/test_unit/test_security/test_allauth/test_allauth_schema.py
sobolevn, line 1: I would say that we need a new integration test for this, something like `django_test_app/apps/allauth` with two external views that provide an auth (we may reuse them in the docs as well). And two views that require this auth.

tests/test_unit/test_security/test_allauth/test_allauth_session_auth.py
```diff
@@ -0,0 +1,238 @@
...
+@pytest.mark.asyncio
+@pytest.mark.django_db(transaction=True)
+async def test_async_session_token_auth(
+    dmr_async_rf: DMRAsyncRequestFactory,
+    admin_user: User,
+    make_session_token: Callable[[User], str],
+) -> None:
+    """Ensures XSessionTokenAsyncAuth authenticates a valid token."""
+    # Session creation touches the DB, we are in an async context here:
+    token = await sync_to_async(make_session_token)(admin_user)
```
sobolevn: we can just use the sync function in tests, it would be even faster.

sobolevn (commented):
Awesome! Looks almost ready.

Some extras:
- Please add `[allauth]` extra to the list of extras in the readme
- Please add the same extra to the `getting_started.rst` guide with the link

pyproject.toml
```diff
@@ -69,6 +69,7 @@ pydantic = [ "pydantic[email,timezone]>=2.12", "pydantic-extra-types" ]
 msgspec = [ "msgspec>=0.21.1" ]
 attrs = [ "attrs>=26", "msgspec>=0.21.1" ]
 jwt = [ "pyjwt[crypto]>=2.11" ]
+allauth = [ "django-allauth>=65.4,<66" ]
```
```suggestion
allauth = [ "django-allauth>=65.4" ]
```

This would be hard for users to update `allauth` with the pinned version on our end. They can add `allauth<70` (or whatever) to their own `pyproject.toml` in case they need to limit the version.

CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ of requirements for an API to count as public.
 
 ## WIP
 
+### Features
+
+- Added `XSessionTokenSyncAuth` and `XSessionTokenAsyncAuth`
+  to authenticate `django-allauth` headless session tokens,
+  available via the new `django-modern-rest[allauth]` extra, #1193
```
sobolevn, line 34:
```suggestion
  available via the new `django-modern-rest[allauth]` extra, #1193

```

docs/pages/auth/allauth.rst
vyhuholl: Added allauth_external_views example.

docs/examples/auth/allauth/allauth_external_views.py
```diff
@@ -0,0 +1,78 @@
...
+from allauth.headless.constants import Client
+
+from dmr import Controller
+from dmr.openapi import load_schema
+from dmr.openapi.objects import PathItem
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.routing import Router, external_path, path
+from dmr.security.allauth import XSessionTokenSyncAuth
+
+# `django-allauth` serves these two endpoints, we only describe them.
```
sobolevn: I propose to hide it from the user-facing example, let's just create a new python file in the same `example/` folder and import these two from there. With a comment near their usage, what these two objects are and the same link to the docs.

tests/test_unit/test_security/test_allauth/test_allauth_schema.py
vyhuholl: Added `django_test_app/apps/allauth_auth` app (because `allauth` app name is already taken) + 18 tests.

docs/examples/auth/allauth/allauth_external_views.py
```diff
@@ -0,0 +1,78 @@
...
+    "summary": "Log out and invalidate the session token",
+    "responses": {
+      "401": {"description": "Logged out, the token no longer resolves"}
+    }
+  }
+}
+"""
+
+
+class MeController(Controller[PydanticSerializer]):
```
sobolevn: I don't think that we need one more usage conrtoller here. We already have one in another example.

docs/examples/auth/allauth/allauth_external_views.py
```diff
@@ -0,0 +1,78 @@
...
+        ),
+        external_path(
+            'session/',
+            SessionView.as_api_view(client=Client.APP),
+            name='current_session',
+            openapi=load_schema(json.loads(_SESSION_OPENAPI), PathItem),
+        ),
+        path('me/', MeController.as_view(), name='me'),
+    ],
+)
```
sobolevn, line 34: Can we at least show `# openapi:` here? It would require to define `urlpatterns` and `use_urlpatterns` option.

tests/test_unit/test_security/test_allauth/test_allauth_session_auth.py
vyhuholl: Moved session token creation to fixture.

sobolevn (commented):
Almost there! 🎉

docs/examples/auth/allauth/allauth_external_views.py
vyhuholl: Moved allauth login and session endpoint descriptions to `allauth_openapi.py` file in the same folder.

docs/examples/auth/allauth/allauth_external_views.py
vyhuholl: Removed controller.

sobolevn (approved):
Fantastic! 

Btw, are you interested in joining the DMR team? :)

docs/examples/auth/allauth/allauth_external_views.py
vyhuholl: Added OpenAPI schema directive via `urlpatterns` + `use_urlpatterns`. Because `# openapi:` headless views require `allauth.account`, I've added allauth to example runner's `INSTALLED_APPS`. Side effect — `AccountMiddleware` is now active for all examples, not just allauth examples.

vyhuholl:
> Btw, are you interested in joining the DMR team? :)

Thank you for the invitation! I would like to.

sobolevn:
Invite sent, welcome :)

PR #1305, opened 2026-08-24
Add Parser.validate method for import-time validation, #1304

### LINKED ISSUE #1304: Add `Parser.validate` method to perform import-time checks
While working on #1275, I realized that proposed `OctetStreamParser` is kinda unique,
because it must not contain `Body[...]` near `FileMetadata[...]` defintions.
Why? Because it only accepts a single stream of bytes, there can't be a body nearby.

So, what should we do for similar cases? I propose adding `validate` method to the `Parser` class, empty by default. It would be similar to `ComponentParser.validate` method. And then call it in validation process.

This test parser should check that it only works on endpoints with `FileMetadata` and no `Body`.

SammySN-car:
Closes #1304

Adds a Parser.validate() method (empty by default) and wires it into the existing import-time validation flow via _validate_parsers().

Parsers can now override validate() to enforce constraints at import time, similar to ComponentParser.validate().

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ of requirements for an API to count as public.
   to authenticate `django-allauth` headless session tokens,
   available via the new `django-modern-rest[allauth]` extra, #1193
 - Added `query` method support for `PathItem` OpenAPI 3.2.0 spec, #1300
-- Added ``Parser.validate`` method for import-time validation of parser
+- Added `Parser.validate` method for import-time validation of parser
   configuration, #1304
 
 ### Bugfixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ of requirements for an API to count as public.
   to authenticate `django-allauth` headless session tokens,
   available via the new `django-modern-rest[allauth]` extra, #1193
 - Added `query` method support for `PathItem` OpenAPI 3.2.0 spec, #1300
+- Added ``Parser.validate`` method for import-time validation of parser
+  configuration, #1304
 
 ### Bugfixes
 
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -44,7 +44,7 @@ per-file-ignores =
   dmr/openapi/objects/*.py: WPS201, WPS110
   # We have complex modules with a lot of quircks:
   dmr/endpoint.py: WPS201, WPS203, WPS402, WPS404
-  dmr/validation/endpoint_metadata.py: WPS201, WPS203, WPS402
+  dmr/validation/endpoint_metadata.py: WPS201, WPS203, WPS214, WPS402
```
```suggestion
  dmr/validation/endpoint_metadata.py: WPS201, WPS203, WPS402
```

Do not ignore it inline for a class

dmr/parsers.py
```diff
@@ -41,6 +41,13 @@ class Parser(ResponseSpecProvider):
     Must be defined for all subclasses.
     """
 
+    def validate(
+        self,
+        controller_cls: type['Controller[BaseSerializer]'],
+        metadata: EndpointMetadata,
+    ) -> None:
+        """Validate parser configuration at import time."""
```
sobolevn: Document that it should raise `EndpointMetadataError` for invalid cases.

sobolevn (commented):
Please, add a test case that it works as expected for a custom parser with a custom validation rule.

Please, update the changelog with a new feature.

dmr/parsers.py
SammySN-car: ok ill fix and and push again

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ of requirements for an API to count as public.
   to authenticate `django-allauth` headless session tokens,
   available via the new `django-modern-rest[allauth]` extra, #1193
 - Added `query` method support for `PathItem` OpenAPI 3.2.0 spec, #1300
+- Added ``Parser.validate`` method for import-time validation of parser
```
sobolevn, line 38:
```suggestion
- Added `Parser.validate` method for import-time validation of parser
```

SammySN-car:
6 test failed , ill push in a minute

sobolevn (approved):

PR #1312, opened 2026-08-25
Fix Redis throttling reset values using consistent time source, #1308

### LINKED ISSUE #1308: `test_throttle_multiple_headers_async` reports reset values above the configured window
### What's wrong?

`test_throttle_multiple_headers_async` intermittently fails because the reported Redis throttling reset values are one second larger than the configured windows.
This is a rare and difficult-to-reproduce failure. Re-running the test suite with the same `pytest-randomly` seed does not reproduce it reliably, so the failure does not appear to depend on test ordering or randomization.

The failure occurred during a full test run:

```bash
uv run python -m pytest -n auto --max-worker-restart=1 --inline-snapshot=disable
```

Observed failure:

```text
{'X-RateLimit-Reset': '3601'} !=
{'X-RateLimit-Reset': IsOneOf('3600', '3599')}

{'RateLimit': '"RemoteAddr";r=0;t=61'} !=
{'RateLimit': IsOneOf('"RemoteAddr";r=0;t=60', '"RemoteAddr";r=0;t=59')}
```

### How it should be?

The reported reset values should not exceed the configured 60-second and 3600-second throttling windows, and the test should pass consistently.

### Used versions

```
platform darwin -- Python 3.13.13
pytest-9.1.1
pluggy-1.6.0
codspeed-5.0.3
randomly-seed=4206430425
```

### OS information

macOS 26.5.2
Redis started in a container by testcontainers

SammySN-car:
Closes #1308

Root cause: Lua script used 
edis.call('TIME') to compute expire_at using Redis server time, while Python used 	ime.time() for the reset calculation. When these clocks differed by 1 second, the reported reset value was off by 1.

Fix: Return 	tl directly from Lua instead of expire_at. Add is_ttl flag to CachedRateLimit so the algorithm knows whether 	ime is a relative 	tl (from Redis) or an absolute expire_at (from Django cache). The algorithm uses 	tl directly for reset when is_ttl=True, and expire_at - now when is_ttl=False.

This eliminates the cross-system clock skew entirely. The Redis backend now uses a single time source (Redis TTL) with no conversion.


dmr/throttling/lua.py
```diff
@@ -40,22 +40,17 @@
 end
 
 local ttl = redis.call("TTL", key)
--- Use Redis server clock so the timestamp is always consistent
--- with the data stored in Redis (no Python <-> Redis skew).
```
sobolevn, line 44: Looks like that we just moved this problem to another place 🤔

dmr/throttling/backends/redis.py
```diff
@@ -228,5 +229,5 @@ async def get(
         )
         return CachedRateLimit(
             history=[script_result[1]],
-            time=script_result[2],
+            time=int(time.time()) + script_result[2],
```
sobolevn, line 232: This likely be very imprecise :(

SammySN-car: @sobolevn, you're right — my fix just moved the skew from Lua to Python.

The core issue: `cache_object['time'] - now` mixes Redis time (in `expire_at`) with Python time (in `now`).

Two approaches I see:

A) Return `ttl` from Lua, use it directly as `reset` in the algorithm without converting to absolute timestamp. Requires changing `SimpleRate._report_usage` to handle `ttl` differently for Redis vs Django cache.

B) Return `ttl` from Lua, compute `expire_at = int(time.time()) + ttl` in Python (my current fix). Both sides use Python time consistently.

Which direction would you prefer?

SammySN-car:
You're right — I was just moving the skew around. Let me think about this more carefully.

After looking at the code, I realize Django cache already has the same behavior:

`python
# _process_cache:
time = int(time.time()) + duration  # call A

# _report_usage:
now = int(time.time())  # call B
reset = cache_object['time'] - now  # = duration + (A - B)
`

If calls A and B happen in different seconds, reset is off by 1. The test already accepts this with IsOneOf.

My fix makes Redis behave identically — both use Python time, both have the same precision. The skew between Redis and Python is eliminated.

Would you prefer a different approach, or is this acceptable?

SammySN-car:
The idea: return ttl from Lua and use it directly as reset, without converting to absolute timestamp at all.
What changes:
1. Lua returns ttl instead of expire_at
2. Redis backend stores ttl directly as cache_object['time']
3. Django cache backend stores duration directly (instead of now + duration)
4. Algorithm changes:
- reset = cache_object['time'] (no subtraction)
- Expiration check: cache_object['time'] <= 0 (instead of cache_object['time'] <= now)
This way, ttl flows through without any time conversion. No expire_at, no time.time(), no mixing time sources.
More changes (4 files instead of 2), but eliminates the problem entirely instead of just reducing it.

is this ok?

sobolevn:
Let's try and see :)

SammySN-car:
> Let's try and see :)

Actually after looking deeper, the TTL-direct approach won't work. Here's why:
Django cache can't get remaining TTL from the cache backend — it doesn't have a Lua script like Redis. So it needs expire_at (absolute timestamp) to compute reset = expire_at - now. If we store ttl directly, the algorithm can't compute the correct reset for Django cache.
If we went with TTL-direct, you'd need to either:
1. Add a creation_time field to CachedRateLimit and compute remaining TTL as duration - (now - creation_time)
2. Or store both ttl and creation_time in CachedRateLimit
3. Or change the algorithm to not rely on cache_object['time'] for reset at all
So the fix is: make Redis behave identically to Django cache. Both use Python time for expire_at, both have the same ±1 second precision (already accepted by tests). The cross-system skew (Redis time vs Python time) is eliminated.
Changes:
- Lua: returns ttl instead of expire_at
- Redis backend: computes expire_at = int(time.time()) + ttl in Python
- Algorithm: unchanged (already uses expire_at - now)

SammySN-car:
@sobolevn any update on this issue, you were travelling so i thought to wait for you to comeback and then ill continue this

sobolevn:
Yeap, I am back :) 

Maybe we should use different types for different backends? So backend can define the source of truth?

SammySN-car:
> Yeap, I am back :)
> 
> Maybe we should use different types for different backends? So backend can define the source of truth?

Hey @sobolevn, could you clarify what you mean by "different types for different backends"? Are you suggesting each backend returns its own CachedRateLimit variant (e.g., Redis with ttl, Django cache with expire_at), and the algorithm handles both?

sobolevn:
The type can be the same, but the content inside the `time` field can be different. It can be the absolute time or relative.

SammySN-car:
> The type can be the same, but the content inside the `time` field can be different. It can be the absolute time or relative.

did some changes , should be gooood i think hmmm

dmr/throttling/algorithms.py
```diff
@@ -146,11 +148,15 @@ def _report_usage(
         *,
         report_all: bool = True,
     ) -> dict[str, str]:
+        if cache_object.get('is_ttl'):
```
sobolevn, line 158: Please, add a comment on what is going on here. In the future we will forget all this context :)

pyproject.toml
```diff
@@ -340,6 +340,7 @@ select = [
 ignore = [
   "A005",     # allow to shadow stdlib and builtin module names
   "ASYNC119", # `contextlib.aclosing` is not detected
+  "COM812",   # conflicts with formatter
```
sobolevn: please remove this

dmr/throttling/algorithms.py
```diff
@@ -123,7 +123,9 @@ def _process_cache(
         cache_object: CachedRateLimit | None,
     ) -> tuple[CachedRateLimit, int]:
         now = int(time.time())
-        if cache_object is None or cache_object['time'] <= now:
+        if cache_object is None or (
+            not cache_object.get('is_ttl') and cache_object['time'] <= now
```
sobolevn, line 130: I don't understand this logic :)

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,81 @@
...
+        'is_ttl': True,
+    }
+    headers = algorithm._report_usage(
+        None,  # type: ignore[arg-type]
+        None,  # type: ignore[arg-type]
+        throttle,
+        cache_object,
+        now=_DUMMY_NOW,
+    )
+    assert headers['X-RateLimit-Reset'] == '45'
```
sobolevn:
```suggestion
    assert headers['X-RateLimit-Reset'] == str(cache_object['time'])
```

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,81 @@
...
+    assert headers['X-RateLimit-Remaining'] == '3'
```
sobolevn:
```suggestion
    assert headers['X-RateLimit-Remaining'] == str(_MAX_REQUESTS - len(cache_object['history']) - 1)
```

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,81 @@
...
+        'time': 100,
+    }
+    headers = algorithm._report_usage(
+        None,  # type: ignore[arg-type]
+        None,  # type: ignore[arg-type]
+        throttle,
+        cache_object,
+        now=_EXPIRE_AT_NOW,
+    )
+    assert headers['X-RateLimit-Reset'] == '30'
```
sobolevn: same here

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,81 @@
...
+_MAX_REQUESTS: Final = 5
+_WINDOW: Final = Rate.minute
+_DUMMY_NOW: Final = 999_999
+_EXPIRE_AT_NOW: Final = 70
+
+
+def test_ttl_reset_uses_ttl_directly() -> None:
+    """TTL mode: reset equals ttl directly, no time subtraction."""
+    algorithm = SimpleRate()
+    throttle = SyncThrottle(_MAX_REQUESTS, _WINDOW, algorithm=SimpleRate())
```
sobolevn:
```suggestion
    throttle = SyncThrottle(_MAX_REQUESTS, _WINDOW, algorithm=algorithm)
```

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,81 @@
...
+        now=_DUMMY_NOW,
+    )
+    assert headers['X-RateLimit-Reset'] == '45'
+    assert headers['X-RateLimit-Remaining'] == '3'
+
+
+def test_expire_at_subtracts_now() -> None:
+    """Expire_at mode: reset = time - now."""
+    algorithm = SimpleRate()
+    throttle = SyncThrottle(_MAX_REQUESTS, _WINDOW, algorithm=SimpleRate())
```
sobolevn:
```suggestion
    throttle = SyncThrottle(_MAX_REQUESTS, _WINDOW, algorithm=algorithm)
```

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,81 @@
...
+        now=_EXPIRE_AT_NOW,
+    )
+    assert headers['X-RateLimit-Reset'] == '30'
+    assert headers['X-RateLimit-Remaining'] == '4'
+
+
+def test_ttl_skips_window_expiry() -> None:
+    """TTL mode: _process_cache does not reset the window."""
+    algorithm = SimpleRate()
+    throttle = SyncThrottle(_MAX_REQUESTS, _WINDOW, algorithm=SimpleRate())
```
sobolevn: and here

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,81 @@
...
+    cached, _now = algorithm._process_cache(throttle, cache_object)
+    assert cached.get('is_ttl') is True
+    assert cached['history'] == [3]
+    assert cached['time'] == 1
+
+
+def test_expire_at_resets_expired_window() -> None:
+    """Expire_at mode: _process_cache resets when window expired."""
+    algorithm = SimpleRate()
+    throttle = SyncThrottle(_MAX_REQUESTS, _WINDOW, algorithm=SimpleRate())
```
sobolevn: and here

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,81 @@
...
+        None,  # type: ignore[arg-type]
+        throttle,
+        cache_object,
+        now=_DUMMY_NOW,
+    )
+    assert headers['X-RateLimit-Reset'] == '45'
+    assert headers['X-RateLimit-Remaining'] == '3'
+
+
+def test_expire_at_subtracts_now() -> None:
```
sobolevn, line 42: this test also needs a bit more explanation about what is going on

sobolevn (commented):
Thank you!

SammySN-car:
> Thank you!

no worries :) , also yeah as per your suggestions i have added those explanations and have fixed those small code stuff,  do lemme know if any change needed

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,106 @@
...
+    algorithm = SimpleRate()
+    throttle = SyncThrottle(_MAX_REQUESTS, _WINDOW, algorithm=algorithm)
+
+    cache_object: CachedRateLimit = {
+        'history': [3],
+        'time': 1,
+        'is_ttl': True,
+    }
+    cached, _now = algorithm._process_cache(throttle, cache_object)
+    assert cached.get('is_ttl') is True
```
sobolevn: shouldn't this just be `assert cached is cache_object`?

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,106 @@
...
+    algorithm = SimpleRate()
+    throttle = SyncThrottle(_MAX_REQUESTS, _WINDOW, algorithm=algorithm)
+
+    cache_object: CachedRateLimit = {
+        'history': [3],
+        'time': 1,
+    }
+    cached, now = algorithm._process_cache(throttle, cache_object)
+    assert cached['history'] == [0]
+    assert cached['time'] == now + 60
```
sobolevn, line 110: assert that `is_ttl` is not in `cached`

tests/test_unit/test_throttling/test_algorithms/test_simple_rate_is_ttl.py
```diff
@@ -0,0 +1,106 @@
...
+        'time': 45,
+        'is_ttl': True,
+    }
+    headers = algorithm._report_usage(
+        None,  # type: ignore[arg-type]
+        None,  # type: ignore[arg-type]
+        throttle,
+        cache_object,
+        now=_DUMMY_NOW,
+    )
```
sobolevn, line 34: add a new line before the `assert` sections in all tests.

CHANGELOG.md
```diff
@@ -72,6 +72,10 @@ https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
 
 ### Bugfixes
 
+- Fixed Redis throttling reset values sometimes being 1 second above the
+  configured window due to clock skew between Redis server time and Python
+  time. The Redis backend now uses the ``ttl`` returned by Redis directly
```
```suggestion
  time. The Redis backend now uses the `ttl` returned by Redis directly
```

SammySN-car:
done applied those changes

sobolevn (approved):
Thank you!

SammySN-car:
> Thank you!

finally its done , it took me and my llm agent 3 weeks lol

PR #1314, opened 2026-08-26
Add dmr-from-dj-rest-auth agent skill

### LINKED ISSUE #1193: Support migration from `dj-rest-auth`

vyhuholl:
# I have made things!

`dj-rest-auth` is a DRF layer over `django-allauth`, so the migration is not shaped like the DRF one. Account management stays in allauth through its headless mode, while dmr takes over the API surface and the auth classes that read allauth's credentials.

The skill says so up front: paths and payloads change, strict transport parity is not achievable, and every API client has to be updated. An agent that cannot change clients is told to stop rather than quietly build a compatibility shim.

Because of that the default mode is behavioral parity rather than transport parity, and the report has a fourth section for security posture changes, so a token becoming readable by JavaScript cannot get lost among path changes.

The reference map is built from the sources of both libraries: endpoints, MFA and social flows, auth classes, all the REST_AUTH settings, and the serializers. It calls out the two things that vanish silently, the missing user-details endpoint and side effects hidden in overridden serializers.
## AI Policy

## Checklist

## Related issues

Closes #1193, I guess?..


sobolevn (commented):
Please, don't forget to list this skill in `getting-started.rst`

sobolevn (approved):

Issue #1290, opened 2026-08-20
Reusable controllers to issue JWT tokens as cookies


vyhuholl:
# Feature

## Thesis

Follow-up to #1287. That PR added `CookieJWTSyncAuth` / `CookieJWTAsyncAuth`, which *read* a JWT from a cookie. Nothing in the framework *writes* one, so the issuing half is still left to every user.

`docs/pages/auth/jwt.rst` currently carries a `.. todo::` in place of an example, and this issue is that todo.

### What is needed

1. Obtain: authenticate and set the access and refresh cookies
2. Refresh: read the refresh token from its cookie and rotate both
3. Log out: clear both cookies, ideally blocklisting the access token (there is no logout controller today at all)

### The design obstacle

This is the part that needs a decision, and it is why the todo is still a todo rather than a patch.

Cookie values are only known at request time, but both ways of declaring cookies fix them at decoration time:

- `@modify(cookies=...)` takes `NewCookie` instances, and `ModifyEndpointPayload.actionable_cookies()` returns exactly those objects. Static values only.
- `@validate` takes `ResponseSpec(cookies=...)`, which accepts only `CookieSpec`, that is a description and not a value. Runtime values then go through `self.to_response(..., cookies={...: NewCookie(value=...)})`.

The second one works, and it is what a hand-written controller does today. But in a *reusable* controller the decorator runs once, on the base class. The cookie names and flags would be frozen there, and a subclass could only change them by redefining `post` completely, which removes the reason to have a reusable controller in the first place.

Note that per-subclass data does already reach endpoint metadata: `Controller.__init_subclass__` builds an `Endpoint` per concrete subclass, and `ResponseSpecProvider.provide_response_specs` receives `controller_cls`. So a hook is feasible. The open question is its shape, not whether it can exist.

### Open questions

- Where do the cookie names and flags come from? `ClassVar`s on the controller like the existing `jwt_*` settings, dmr settings, or a dedicated spec object?
- Should the tokens still appear in the response body? Putting them in both places is convenient and undoes the point of `httponly`.
- Does logout blocklist the access token, or stay transport-only?
- How is `Set-Cookie` represented in the generated OpenAPI schema when the names are configurable?

### Requirements for whatever we build

- `httponly=True` and `secure=True` by default, and `samesite` no weaker than `'lax'`
- The refresh cookie scoped by `path` to the refresh endpoint, so it is not sent to the rest of the API
[...]

PR #1316, opened 2026-08-26
Add `FromController` for response headers and cookies, #1296

### LINKED ISSUE #1296: Think about a way to modify response specs and metadata for endpoints in reusable controllers
Currently we can customize parsing details and response models for endpoints via generics. But, we can't do the same with response headers and response cookies, especially with `@validate`.

I will figure this out and send a PR tomorrow.
Ideas are welcome in the meantime!

Part of the https://github.com/wemake-services/django-modern-rest/issues/1290

vyhuholl:
# I have made things!

Reusable controllers apply `@validate` and `@modify` once, on the base class, so response headers and cookies were frozen there. `FromController` wraps a `classmethod` and resolves it against each concrete subclass while its endpoint metadata is built.

Resolution happens in the metadata builder for `@modify` and in `_resolve_all_responses` for `@validate`, both before anything reads the mappings. An unresolved mapping raises instead of looking empty.

Also adds `CookieSpec.to_new()`, so a real cookie can be derived from its own spec: response validation compares every cookie attribute, and the two definitions have to agree.
## AI Policy

## Checklist

## Related issues

Closes #1296
Refs #1290


sobolevn (commented):
I have some ideas about this PR, but I am traveling right now, so I wont' be able to implement them for several days :)

sobolevn:
Replaced with full-featured solution: https://github.com/wemake-services/django-modern-rest/pull/1409

Thanks for your work!

PR #1355, opened 2026-09-01
Fix empty response body checks for 1xx, 205, and HEAD

### LINKED ISSUE #1340: Incorrect empty response check in `_check_empty_response_body`
https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/validation/endpoint_metadata.py#L151-L160

There are several things that are broken here:
1. `response.status_code < HTTPStatus.CONTINUE` must be `response.status_code < HTTPStatus.OK`, because there are no status codes less than `100`. See [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-informational-1xx)
2. `205 Reset Content` must be in the no-body set: https://www.rfc-editor.org/rfc/rfc9110.html#name-205-reset-content
3. The set should not be created on each call, create a frozenset per-class once
4. `HEAD` methods must not return a body. This needs an extra rule

Shaisolaris:
Fixes #1340

`_check_empty_response_body` used `status < 100` as the 1xx bound, so that branch never ran. It also skipped 205 Reset Content, and HEAD could still advertise a body on 2xx.

This:
- treats 1xx as no-body (`status < 200`)
- keeps 204, 205, and 304 in a class-level frozenset
- rejects a body on successful HEAD responses


sobolevn (commented):
Amazing! Two notes:
1. Please add a CHANGELOG entry for a new feature
2. Add tests for `head` method (both successful and failed responses) :)
3. Add tests for `205` response

Shaisolaris:
Added a Bugfixes CHANGELOG line for #1340. Tests cover 205, successful HEAD with None, and failed HEAD with a body. The collect-metadata extra 205 example is now 206, because 205 cannot carry a body.

PR #1356, opened 2026-09-01
Protect auth views from leaking credentials

### LINKED ISSUE #1323: `@sensitive_variables` decorator is missing on auth views
We already have `@endpoint_decorator(sensitive_post_parameters())` on all API views that work with auth, but I forgot about `sensitive_variables`: https://docs.djangoproject.com/en/6.1/howto/error-reporting/#django.views.decorators.debug.sensitive_variables

We need to add vars that must not leak into the logs / error reporting middlewares / etc.

### LINKED ISSUE #1335: All pre-existing auth views must provide `Cache-Control: no-store` header
Refs https://github.com/wemake-services/django-modern-rest/issues/1323

It should better be explicit. By default `POST` methods are never cached anyway, but we should be better safe here.

vyhuholl:
## `@sensitive_variables` on auth views (#1323)

All auth views already had `@sensitive_post_parameters()`, but that only hides the request's POST data. The local variables of the views themselves were still shown in full in error reports, and `parsed_body` holds the raw password or the raw token. So any unhandled exception in a login view leaked credentials into whatever error reporting middleware a project has installed.

`@sensitive_variables()` is now applied to every method of the shipped auth views that keeps credentials or tokens in its locals:

* django session: `post`, `login`
* opaque token: `post`, `login`, `issue_token`
* jwt: `post`, `login`, `refresh`, `verify`, `create_jwt_token`,
  `_decode_and_validate_refresh_token`, `_decode_and_validate_access_token`

It is used without arguments, the same way `@sensitive_post_parameters()` already is here, and it has to be the topmost decorator. For sync functions Django walks up the stack looking for the `sensitive_variables_wrapper` frame, so one decorator on the endpoint also covers everything below it, including the `endpoint_decorator` wrappers and `django.contrib.auth.authenticate`. Placing it under `@endpoint_decorator` instead would leave those wrappers holding `parsed_body` in their `kwargs`.

While writing the tests it turned out the leak was not limited to the views. `Endpoint._sync_endpoint` and `Endpoint._async_endpoint` kept the parsed request in a local `context` variable, and that frame sits above any view decorator, so the password was visible in error reports of *any* endpoint, not just the auth ones. Inlining the call keeps the parsed data off `f_locals`. Behaviour and performance are unchanged.

### `Cache-Control: no-store` (#1335)

Added `dmr.security.NO_STORE_HEADERS` and passed it to `@modify(headers=...)` on all ten auth endpoints. The header is set at runtime and is documented in the OpenAPI schema. The constant is public so that the same thing is easy to do in custom auth views, and it avoids repeating the literal in three modules.

### Known limitation for async controllers

Async controllers are protected less thoroughly than sync ones, and this is a Django limitation rather than something specific to this change. Coroutines have no shared call stack, so `sensitive_variables` registers them in a global `file:lineno` registry and only covers the frame of the decorated coroutine itself.

After the `Endpoint` fix, an async login still exposes the parsed body in three wrapper frames: the `endpoint_decorator` wrapper, the `_unwrap_request_param` wrapper and Django's own `sensitive_post_parameters_wrapper`. Their code objects are shared by every endpoint that uses `endpoint_decorator`, so marking them sensitive would hide locals for unrelated endpoints and make debugging much worse. The same applies to user-written async `convert_auth_payload` and `make_api_response` overrides, which the docs now call out explicitly.

If full async coverage is wanted, it needs a different mechanism — for example an opt-in flag on the endpoint metadata that `Endpoint` acts on. Happy to look at that separately if you think it is worth it.

### Tests

* `tests/test_unit/test_security/test_sensitive_variables.py` runs `SafeExceptionReporterFilter` over the real traceback of a failing auth view and asserts the credentials come back cleansed. The sync tests additionally assert that no frame of the whole traceback shows the secret, which is what pins the `Endpoint` fix.
* Integration tests assert `Cache-Control: no-store` for session login, opaque token obtain, and the jwt obtain / refresh / verify flow.

### Docs

New "Security of auth views" section in `docs/pages/auth/common.rst` covering both topics, including the sync/async difference and the decorator ordering.

## AI Policy

## Checklist

## Related issues

Closes #1323
Closes #1335


docs/pages/auth/common.rst
```diff
@@ -126,6 +126,94 @@ all auth in further layers.
...
+.. code:: python
+
+  >>> from http import HTTPStatus
+
+  >>> from dmr import Controller, modify
+  >>> from dmr.plugins.pydantic import PydanticSerializer
+  >>> from dmr.security import NO_STORE_HEADERS
+
+  >>> class MyLoginController(Controller[PydanticSerializer]):
+  ...     @modify(status_code=HTTPStatus.OK, headers=NO_STORE_HEADERS)
```
sobolevn: I am not quite sure about this example. Since it does not show ALL the things we need, maybe we can just show the `modify()` call here?

docs/pages/auth/common.rst
```diff
@@ -126,6 +126,94 @@ all auth in further layers.
...
+
+.. note::
+
+  Sync and async functions are protected differently by Django.
+
+  A sync ``@sensitive_variables()`` also hides local variables
+  of everything the decorated function calls,
+  so one decorator on the endpoint covers the whole call chain.
+
+  Async functions do not have a shared call stack,
```
sobolevn, line 206: Looks like we don't follow our own advice here. Because we don't protect `async` functions with credentials in `django_test_app`

docs/pages/auth/common.rst
```diff
@@ -126,6 +126,94 @@ all auth in further layers.
...
+  >>> from dmr.security import NO_STORE_HEADERS
+
+  >>> class MyLoginController(Controller[PydanticSerializer]):
+  ...     @sensitive_variables()
+  ...     @endpoint_decorator(sensitive_post_parameters())
+  ...     @modify(headers=NO_STORE_HEADERS)
+  ...     def post(self, parsed_body: Body[dict[str, str]]) -> str:
+  ...         return self.login(parsed_body)
+  ...
+  ...     @sensitive_variables()
```
sobolevn: Do we need this? 🤔

tests/test_integration/test_jwt_auth/test_jwt_views.py
```diff
@@ -718,3 +718,54 @@ def test_verify_non_numeric_subject(
...
+            reverse('api:jwt_auth:jwt_verify_sync'),
+        ),
+        (
+            reverse('api:jwt_auth:jwt_obtain_access_refresh_async'),
+            reverse('api:jwt_auth:jwt_refresh_async'),
+            reverse('api:jwt_auth:jwt_verify_async'),
+        ),
+    ],
+)
+def test_jwt_views_are_never_stored(
```
sobolevn: I think that it would be better to update existing tests, not add a new one.

tests/test_integration/test_token_auth/test_token_obtain.py
```diff
@@ -126,3 +126,28 @@ def test_obtain_failures(
...
+    ],
+)
+def test_obtain_is_never_stored(
+    dmr_client: DMRClient,
+    user: User,
+    password: str,
+    *,
+    url: str,
+) -> None:
+    """Ensures the issued token response is never written to any cache."""
```
sobolevn: Same here. Also, test both success and failures.

tests/test_unit/test_security/test_sensitive_variables.py
sobolevn, line 1: I vote to remove this test :)
It way too complex for a single decorator.

sobolevn (commented):
Thank you!

docs/pages/auth/common.rst
vyhuholl: Removed everything except the `modify()` call.

docs/pages/auth/common.rst
vyhuholl: Chose to fix `django_test_app` instead of dropping the advice in the documentation — added `@sensitive_variables()` decorator to async functions.

The async payload converters (`convert_auth_payload`, `convert_refresh_payload`, `convert_verify_payload`) receive the raw credentials as an argument, so each one now carries its own decorator. Async `make_api_response` overrides are deliberately left alone: in the test app they never bind a token to a local variable, so a decorator there would only hide unrelated locals.

tests/test_unit/test_security/test_sensitive_variables.py
vyhuholl: Removed.

docs/pages/auth/common.rst
vyhuholl: No, the second decorator is redundant and contradicts the warning below. Thank you for noticing!

Removed the second decorator and also explained why login function does not need it.

tests/test_integration/test_jwt_auth/test_jwt_views.py
vyhuholl: Removed new test, added `Cache-Control` asserts to existing ones — one success and one failure to each endpoint group (session login, token obtain, jwt obtain/refresh/verify).

tests/test_integration/test_token_auth/test_token_obtain.py
vyhuholl: Removed new test, added `Cache-Control` asserts to existing ones — one success and one failure to each endpoint group (session login, token obtain, jwt obtain/refresh/verify).

sobolevn (approved):
Thanks!

PR #1359, opened 2026-09-01
Support `WWW-Authenticate` header for auth classes, #1334

### LINKED ISSUE #1334: Support `WWW-Authenticate` header for auth classes
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/WWW-Authenticate
RFC: https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.2

Quote:

> The server generating a 401 response MUST send a [WWW-Authenticate](https://www.rfc-editor.org/rfc/rfc9110.html#field.www-authenticate) header field ([Section 11.6.1](https://www.rfc-editor.org/rfc/rfc9110.html#field.www-authenticate)) containing at least one challenge applicable to the target resource.
> If the request included authentication credentials, then the 401 response indicates that authorization has been refused for those credentials. The user agent MAY repeat the request with a new or replaced [Authorization](https://www.rfc-editor.org/rfc/rfc9110.html#field.authorization) header field ([Section 11.6.2](https://www.rfc-editor.org/rfc/rfc9110.html#field.authorization)). If the 401 response contains the same challenge as the prior response, and the user agent has already attempted authentication at least once, then the user agent SHOULD present the enclosed representation to the user, since it usually contains relevant diagnostic information.

Currently, we don't support this. But, we probably should. 
Ideas on how to support this are welcome.

Criteria:
1. All auth classes should be supported, if some can't be supported, we need to figure out - why
2. It should be documented as a feature and what is supported
3. There should be a way to disable this (probably?)

vyhuholl:
# I have made things!

RFC 9110 requires every `401` to carry at least one challenge. This PR adds the header to any response that `NotAuthenticatedError` produces: the auth chain running out of options, an auth instance rejecting the credentials, or an endpoint raising it by hand.

A challenge can only name a scheme sent in the `Authorization` header, so cookie-based and custom-header auth send nothing. The header-based classes take a new `www_authenticate=` flag to opt out, and HTTP Basic also takes a `realm=`.

## AI Policy

## Checklist

## Related issues

Closes #1334


dmr/security/jwt/auth.py
```diff
@@ -194,6 +198,23 @@ class _HeaderJWTAuth:
...
+
+        Returns ``None`` for a custom *auth_header*, because a challenge
+        can only ask the client for the ``Authorization`` header.
+        ``realm`` is optional for bearer tokens, see :rfc:`6750#section-3`,
+        and we don't send it.
+        """
+        if not self.www_authenticate or not self.auth_scheme:
+            return None
+        if self.auth_header != _AUTHORIZATION_HEADER:
+            return None
```
sobolevn: can we join all the conditions here?

dmr/security/base.py
```diff
@@ -16,16 +17,67 @@
     from dmr.endpoint import Endpoint
     from dmr.serializer import BaseSerializer
 
+#: Name of the header that carries auth challenges in ``401`` responses.
+WWW_AUTHENTICATE: Final = 'WWW-Authenticate'
+
+#: Returned by auth that has no ``WWW-Authenticate`` challenge to advertise.
+NO_CHALLENGE: Final[str | None] = None
```
sobolevn: Do we even need this value? Why just not return `None`?

dmr/security/base.py
```diff
@@ -55,6 +107,24 @@ def security_requirement(self) -> SecurityRequirement:
         """Provides a security schema usage requirement."""
         raise NotImplementedError
 
+    @property
+    def www_authenticate_challenge(self) -> str | None:
```
sobolevn, line 137: Design question: I guess that this probably needs to be an abstract method with no default implementation. Otherwise, people might forget about it. But, just returning `None` is simple enough, if they don't need it.

dmr/security/http.py
```diff
@@ -13,19 +13,69 @@
     from dmr.endpoint import Endpoint
     from dmr.serializer import BaseSerializer
 
+#: Default protection space that we advertise in ``WWW-Authenticate``.
+DEFAULT_BASIC_REALM: Final = 'api'
+
+
+def quote_auth_param(value: str) -> str:  # noqa: WPS110
```
sobolevn: Should it be public?

dmr/security/http.py
```diff
@@ -13,19 +13,69 @@
     from dmr.endpoint import Endpoint
     from dmr.serializer import BaseSerializer
 
+#: Default protection space that we advertise in ``WWW-Authenticate``.
+DEFAULT_BASIC_REALM: Final = 'api'
```
sobolevn: should this be public?

dmr/endpoint.py
```diff
@@ -521,6 +528,24 @@ async def _run_async_throttle_after(
 
     # Utils:
 
+    def _add_auth_challenge(self, exc: Exception) -> None:
```
sobolevn: My vote would be to turn this into a public function and hide `combined_www_authenticate` and make it protected.

docs/pages/auth/common.rst
```diff
@@ -126,6 +126,137 @@ all auth in further layers.
...
+  WWW-Authenticate: Basic realm="api", charset="UTF-8"
+
+When an endpoint has several auth instances, we join their challenges
+into a single header value, because
+`RFC 9110 <https://www.rfc-editor.org/rfc/rfc9110.html#section-11.6.1>`_
+allows a challenge list:
+
+.. code-block:: text
+
+  WWW-Authenticate: Basic realm="api", charset="UTF-8", Bearer
```
sobolevn, line 174: question: is this correct? why does it contain `, charset="UTF-8"` part? what is this?

CHANGELOG.md
```diff
@@ -27,9 +27,21 @@ of requirements for an API to count as public.
...
 ### Features
 
+- Added `WWW-Authenticate` support for auth classes that read
+  the `Authorization` header: `HttpBasicSyncAuth`, `HttpBasicAsyncAuth`,
+  `HeaderJWTSyncAuth`, `HeaderJWTAsyncAuth`, `HeaderTokenSyncAuth`,
+  and `HeaderTokenAsyncAuth`. Cookie-based and custom-header auth
+  send no challenge, because there is none to express.
+  Configurable via the new `www_authenticate=` and `realm=` arguments
+  and the `SyncAuth.www_authenticate_challenge` property, #1334
+
```
sobolevn: extra new line

dmr/endpoint.py
```diff
@@ -521,6 +528,24 @@ async def _run_async_throttle_after(
...
+        and all of them need the same challenge.
+        An explicitly passed ``headers=`` always wins.
+        """
+        if not isinstance(exc, NotAuthenticatedError):
+            return
+        if exc.headers is not None:
+            return
+        challenge = combined_www_authenticate(self.metadata.auth)
+        if challenge is not None:
+            exc.headers = {WWW_AUTHENTICATE: challenge}
```
sobolevn: We can also make `WWW_AUTHENTICATE` protected

sobolevn (commented):
Thanks!

dmr/security/jwt/auth.py
vyhuholl: Done.

dmr/security/base.py
vyhuholl: Made `www_authenticate_challenge` abstract. The classes that have nothing to advertise say so explicitly.

dmr/security/http.py
vyhuholl: Made public.

dmr/security/http.py
vyhuholl: Made public.

CHANGELOG.md
vyhuholl: Fixed.

dmr/endpoint.py
vyhuholl: Replaced `_add_auth_challenge` with `add_auth_challenge(exc, auth)` from `dmr/security/base.py`. Made `combined_www_authenticate` protected.

dmr/endpoint.py
vyhuholl: Done.

dmr/security/base.py
vyhuholl: Removed.

docs/pages/auth/common.rst
vyhuholl: This is the `Basic` scheme from [RFC 7617 §2.1](https://www.rfc-editor.org/rfc/rfc7617.html#section-2.1) auth param, which tells the client which encoding to use for the username and password. ``UTF-8`` is its only allowed value, and it matches what we decode the credentials as, so we always send it.

Added a note about this to docs.

dmr/security/http.py
```diff
@@ -13,19 +13,71 @@
     from dmr.endpoint import Endpoint
     from dmr.serializer import BaseSerializer
 
+# Default protection space that we advertise in `WWW-Authenticate`:
+_DEFAULT_BASIC_REALM: Final = 'api'
+
+
+def _quote_auth_param(value: str) -> str:  # noqa: WPS110
```
sobolevn:
```suggestion
def _quote_auth_param(auth_param: str) -> str:
```

dmr/security/http.py
```diff
@@ -13,19 +13,71 @@
...
+    .. code:: python
+
+      >>> _quote_auth_param('api')
+      '"api"'
+
+      >>> _quote_auth_param('say "hi"')
+      '"say \\"hi\\""'
+
+    """
+    escaped = value.replace('\\', r'\\').replace('"', r'\"')
```
sobolevn:
```suggestion
    escaped = auth_param.replace('\\', r'\\').replace('"', r'\"')
```

dmr/security/http.py
```diff
@@ -13,19 +13,71 @@
...
+        """
+        if (
+            not self.www_authenticate
+            or not self._uses_standard_http_basic_auth()
+        ):
+            return None
+        # `charset` is defined by RFC 7617 and tells the client to encode
+        # credentials as UTF-8, which is what we decode them as
+        # in `_get_username_and_password`.
+        return f'Basic realm={_quote_auth_param(self.realm)}, charset="UTF-8"'
```
sobolevn, line 63: wait, this is actually really strange to see in a RFC 😒 
they separate both VALUES and CHANLANGES with `,`. This is kinda unexpected. Usually `;` and `,` are utilized to tell the difference. Ok, TIL

dmr/endpoint.py
```diff
@@ -260,6 +265,11 @@ async def handle_async_error(
         Override this method to add custom async error handling.
         """
         # NOTE: if you change something here, also change in `handle_error`
+        # We do this here, and not in `_run_auth`, because a `401` can also
+        # come from an auth instance itself or be raised by hand
+        # from an endpoint body. All of them need the same challenge.
+        if isinstance(exc, NotAuthenticatedError):
```
sobolevn: I propose moving this to `global_error_handler`. This way we would have a default handler with `WWW-Authenticate` header addition. But, it would be possible to easily override this behavior.

docs/pages/auth/common.rst
```diff
@@ -126,6 +126,173 @@ all auth in further layers.
...
+
+  >>> HeaderJWTSyncAuth(www_authenticate=False).www_authenticate_challenge
+
+.. warning::
+
+  Browsers show their own native login prompt when they see a ``Basic``
+  challenge on a ``401``. If your API is called from a browser and you do
+  not want that popup, turn the challenge off for HTTP Basic auth.
+
+Writing your own auth
```
sobolevn: Please, remove this section for now. Let's create a new issue for this. We need to show how to do in the docs. But, we need a full guide. Not just `www_authenticate_challenge` part.

docs/pages/auth/common.rst
```diff
@@ -324,6 +491,8 @@ API Reference
 
 .. autofunction:: dmr.security.request_auth
 
+.. autofunction:: dmr.security.add_auth_challenge
```
sobolevn:
```suggestion
.. autofunction:: dmr.security.add_www_authenticate
```

what do you think about this name instead?

tests/test_unit/test_security/test_www_authenticate.py
```diff
@@ -0,0 +1,309 @@
...
+        (_RejectingBasicAuth(www_authenticate=False), None),
+        (HeaderJWTSyncAuth(auth_scheme='JWT'), 'JWT'),
+        (HeaderJWTSyncAuth(auth_header='X-Api-Auth'), None),
+        (HeaderJWTSyncAuth(www_authenticate=False), None),
+        (
+            HeaderTokenSyncAuth(header_name='Authorization', prefix='Token'),
+            'Token',
+        ),
+        # Without a prefix there is no scheme name to build a challenge from:
+        (HeaderTokenSyncAuth(header_name='Authorization'), None),
```
sobolevn, line 140: Please, test `HeaderTokenAsyncAuth` as well

vyhuholl: Done.

dmr/security/http.py
```diff
@@ -13,19 +13,71 @@
     from dmr.endpoint import Endpoint
     from dmr.serializer import BaseSerializer
 
+# Default protection space that we advertise in `WWW-Authenticate`:
+_DEFAULT_BASIC_REALM: Final = 'api'
+
+
+def _quote_auth_param(auth_param: str) -> str:
```
sobolevn: Please, move this function to the bottom of the file. Files must start with important parts and then lead to less important ones.

sobolevn (commented):
Thanks a lot for your awesome work! One last thing :)

dmr/endpoint.py
vyhuholl: Moved this block to `global_error_handler` in `dmr/security/errors.py`.

docs/pages/auth/common.rst
vyhuholl: Done, and also, created an issue #1366

dmr/security/http.py
vyhuholl: Done)

sobolevn (approved):

PR #1360, opened 2026-09-01
Document removing expired blocklisted and opaque tokens

### LINKED ISSUE #1336: Document that it is a good idea to remove expired tokens from `BlocklistedJWToken`
Since they are expired anyway, they won't be usable for the auth.

So, there's no point in storing them.
The same can be said about `Token` model from `dmr/security/token/app/models.py`.

Document that adding a cron-job to remove old ones is probably worth it.
Show examples, what we can remove in both cases.

vyhuholl:
# I have made things!

Expired rows in `BlocklistedJWToken` and `Token` cannot change any auth decision anymore, so they are pure dead weight.

Both auth pages now explain what is safe to delete and what must stay, and ship a management command example to schedule with cron.

## AI Policy

## Checklist

## Related issues

Closes #1336

### DIFF
--- a/docs/pages/auth/jwt.rst
+++ b/docs/pages/auth/jwt.rst
@@ -360,16 +360,7 @@ We recommend deleting them with a periodic job:
   :linenos:
   :language: python
 
-And then schedule it, for example with ``cron``:
-
-.. code-block:: bash
-  :caption: crontab
-
-  0 4 * * * /path/to/venv/bin/python /path/to/manage.py cleanup_blocklist
-
-Anything else that runs code on a schedule will do:
-Celery beat, ``django-q``, a ``systemd`` timer,
-or whatever your hosting provides.
+Then run this task as a periodic job.
 
 .. warning::
 
--- a/docs/pages/auth/jwt.rst
+++ b/docs/pages/auth/jwt.rst
@@ -333,6 +333,58 @@ We provide two mixin types:
 
 If this app is installed, we would provide an admin panel by default.
 
+.. _cleaning-up-blocklisted-tokens:
+
+Cleaning up expired tokens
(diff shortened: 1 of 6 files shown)


docs/pages/auth/jwt.rst
```diff
@@ -333,6 +333,58 @@ We provide two mixin types:
...
+And then schedule it, for example with ``cron``:
+
+.. code-block:: bash
+  :caption: crontab
+
+  0 4 * * * /path/to/venv/bin/python /path/to/manage.py cleanup_blocklist
+
+Anything else that runs code on a schedule will do:
+Celery beat, ``django-q``, a ``systemd`` timer,
+or whatever your hosting provides.
```
sobolevn: Let's remove this part.

You can just say: "Run this task as a periodic job"

docs/pages/auth/token.rst
```diff
@@ -175,6 +175,64 @@ Here's an example with the default model:
...
+:meth:`~dmr.security.token.token.TokenLikeSync.revoke` is what stops
+a token that is still active. Cleanup is just housekeeping afterwards,
+and it is worth doing periodically:
+
+.. literalinclude:: /examples/auth/token/token_cleanup.py
+  :caption: myapp/management/commands/cleanup_tokens.py
+  :linenos:
+  :language: python
+
+Note that ``expires_at__lt`` never matches ``NULL``,
```
sobolevn, line 208: Same here

docs/pages/auth/jwt.rst
vyhuholl: Done.

docs/pages/auth/token.rst
vyhuholl: Done.

sobolevn (approved):

PR #1361, opened 2026-09-01
Validate `exp` and `iat` in a single place, #1324

### LINKED ISSUE #1324: `leeway` / `iat` checks are not consistent in JWT token
Currently we do double validation of `leeway` and `iat` for tokens:
1. https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/security/jwt/token.py#L97-L104 Our way
2. https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/security/jwt/token.py#L154-L162 pyjwt way

The question is: do we really need to do this second validation here: https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/security/jwt/token.py#L257

Because it was just checked here: https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/security/jwt/token.py#L236-L242

Context:
- `_validate_iat` in `jwt/api_jwt.py` in pyjwt

This probably needs a rework to have a single source of truth. API breakage is allowed.

vyhuholl:
# I have made things!

The double validation mentioned in #1324  did not merely duplicate work — it **overrode** the pyjwt settings. Verified against the live code:

| Case | Before | After |
|---|---|---|
| `verify_exp=False` + expired token | `ValueError` → **500** | token is decoded |
| `verify_iat=False` + `iat` in the future | `ValueError` → **500** | token is decoded |
| `iat` in the future, within `leeway` | `ValueError` → **500** | token is decoded |
| `sub: ''` in a signed token | `ValueError` → **500** | `NotAuthenticatedError` → 401 |

The cause: `__post_init__` checked `exp` / `iat` by its own rules (`leeway` was not applied to `iat`, the `verify_*` flags were not taken into account), and it was called at the end of `decode()` — after pyjwt had already checked everything.

**Solution:** a single source of truth **per direction**, instead of a "skip validation" flag:

- **issuing a token** → our own checks, now in `encode()` (`dmr/security/jwt/token.py:92`): `exp` not in the past, `iat` not in the future, `ValueError`;
- **parsing a token** → entirely pyjwt with its `options` / `leeway`, no second check.

In `__post_init__` I kept only what does not depend on time: a non-empty `sub` and datetime normalization. The init-only `leeway` argument of `JWToken` was removed — it existed solely for the sake of that repeated check (breakage is allowed by the issue). In `decode`, the constructor is wrapped in `except ValueError → NotAuthenticatedError`, so that any remaining field validation yields a 401 rather than a 500.

**A risk worth being aware of:** `verify_exp=False` now genuinely disables the expiry check. Previously an expired token was rejected anyway — accidentally, by the second check. This is the declared behaviour of the flag, but the semantics have changed.

## AI Policy

## Checklist

## Related issues

Closes #1324

### DIFF
--- a/dmr/security/jwt/token.py
+++ b/dmr/security/jwt/token.py
@@ -29,7 +29,7 @@
 
 import datetime as dt
 from collections.abc import Sequence
-from dataclasses import InitVar, asdict, dataclass, field, fields
+from dataclasses import asdict, dataclass, field, fields
 from typing import Any, Self
 
 import jwt
@@ -53,6 +53,14 @@ class JWToken:
         jti: JWT ID - a unique identifier of the JWT between different issuers.
         extras: Extra fields that were found on the JWT token.
 
+    .. versionchanged:: 0.15.0
+
+        Init-only ``leeway`` argument was removed.
+        Time-based claims are now validated in a single place
+        for each direction: :meth:`encode` checks that a token
+        can be issued, while :meth:`decode` fully relies
+        on ``pyjwt`` and its options.
+
     """
 
     sub: str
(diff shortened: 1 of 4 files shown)


dmr/security/jwt/token.py
```diff
@@ -122,8 +108,16 @@ def encode(
             An encoded token string.
 
         Raises:
+            ValueError: If this token cannot be issued right now.
             InternalServerError: If encoding fails.
```
sobolevn, line 131: while we are at it. it seems strange to me that `Token.encode` produces `InternalServerError`?! 

what was I thinking? can you please create a new issue about it?

we need to raise some semantic error here and convert it to `ServerInternalError` somewhere else.

tests/test_unit/test_security/test_jwt/test_jwt_token.py
```diff
@@ -337,3 +340,121 @@ def test_invalid_datetime_claim_raises_auth_error() -> None:
...
+                'sub': '',
+                'iat': dt.datetime.now(dt.UTC),
+                'exp': dt.datetime.now(dt.UTC) + dt.timedelta(days=1),
+            },
+            id='empty-sub',
+        ),
+    ],
+)
+def test_invalid_claims_raise_auth_error(
+    raw_token_data: dict[str, Any],
```
sobolevn, line 465:
```suggestion
    *,
    raw_token_data: dict[str, Any],
```

this is our style to tell apart parametrized values from fixtures :)

dmr/security/jwt/token.py
```diff
@@ -318,6 +326,15 @@ def _decode_datetime_claim(
             raise NotAuthenticatedError from None
 
 
+def _validate_issued_claims(token: JWToken) -> None:
```
sobolevn: should this be a method, so it would be easier to customize if needed?

sobolevn (commented):
Thanks!

dmr/security/jwt/token.py
vyhuholl: Made `validate_issued_claims` a customizable method.

dmr/security/jwt/token.py
vyhuholl: Done, see #1364

sobolevn (approved):

PR #1362, opened 2026-09-01
Require `auth_scheme` prefix in http basic auth

### LINKED ISSUE #1330: `HttpBasicSyncAuth` and `HttpBasicAsyncAuth` must strictly configure `auth_scheme` prefix
Currently HTTP-Basic auth is not strict about its `auth_scheme` prefix.
This would be a breaking change, I allow it.

Currently, we allow both header values: `some-login:password` and `Basic some-login:password`. Which is not really cool: https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/security/http.py#L56-L76

We must make this configurable, as all other auth classes do. This must require `auth_scheme: str = 'Basic'` prefix by default. 

Also, we must update the OpenAPI description as well: https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/security/http.py#L82-L89

vyhuholl:
# I have made things!
**`dmr/security/http.py`**

- `_HttpBasicAuth.__init__` gains `auth_scheme: str = 'Basic'` (new slot), documented alongside `header` and `security_scheme_name`.
- New `_split_encoded_credentials()` mirrors JWT's `split_encoded_token`: the value must be exactly two space-separated parts and `parts[0]` must equal `auth_scheme`. A bare `<base64>` value is no longer accepted.
- `dmr/security/http.py:82-85`: once the prefix matches, an undecodable payload now `raise NotAuthenticatedError from None` instead of returning `None` — so it fails the request immediately and does not fall through to the next auth in the chain.
- `_uses_standard_http_basic_auth()` additionally requires `auth_scheme.casefold() == 'basic'`, so a custom scheme is emitted as an `apiKey` scheme; the OpenAPI description is now `` `{auth_scheme} <base64(username:password)>` format `` (the "or bare value" alternative is gone).
- `basic_auth()` is unchanged in behaviour; its second doctest switched from `prefix=''` to `prefix='Custom '`, and it notes that `prefix` must match the reader's `auth_scheme`.

**Tests** — `tests/test_unit/test_security/test_http_basic_auth.py` grew from 2 to 15 tests (sync + async): scheme/case-insensitive-`basic` schema, custom-header and custom-scheme schema, accepted vs rejected header values (no prefix, `basic`, `Bearer`, prefix-only, extra parts), and two chained-auth tests that pin the new raise semantics — a missing header still falls through to a second auth (200), while broken credentials with the right prefix return 401 even though the fallback auth would have succeeded. Integration: added `'basic dGVzdDpwYXNz'` and `'dGVzdDpwYXNz'` to the rejected headers in `tests/test_integration/test_contollers/test_auth.py:64-69`, and `tests/test_integration/test_contollers/test_invalid_requests.py:109` no longer uses `prefix=''`.

**Docs/changelog** — new "Customization" section in `docs/pages/auth/http-basic.rst:48-67`, `.. versionchanged:: 0.15.0` on both classes, two breaking-change entries in `CHANGELOG.md:30-37`.

## AI Policy

## Checklist

## Related issues

Closes #1330

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,9 @@ of requirements for an API to count as public.
 - `HttpBasicSyncAuth` and `HttpBasicAsyncAuth` now require
   the `auth_scheme` header prefix, it is `Basic` by default
   and is matched exactly, credentials sent without it
-  are not accepted anymore, #1330
+  are not accepted anymore.
+  Pass `auth_scheme=''` to keep reading prefixless
+  credentials like the older versions did, #1330
 - `HttpBasicSyncAuth` and `HttpBasicAsyncAuth` now raise
   `NotAuthenticatedError` when credentials have the right
   `auth_scheme` prefix, but cannot be decoded,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,14 @@ of requirements for an API to count as public.
 - Throttling cache keys are now hashed to keep their length bounded, #1337
 - HTTP Basic Auth credentials are no longer URL-decoded,
   so percent-encoded characters such as `%40` are preserved as-is, #1363
+- `HttpBasicSyncAuth` and `HttpBasicAsyncAuth` now require
+  the `auth_scheme` header prefix, it is `Basic` by default
+  and is matched exactly, credentials sent without it
+  are not accepted anymore, #1330
+- `HttpBasicSyncAuth` and `HttpBasicAsyncAuth` now raise
+  `NotAuthenticatedError` when credentials have the right
+  `auth_scheme` prefix, but cannot be decoded,
+  previously the next auth in the chain was tried, #1330
 
 ### Features
 
--- a/dmr/security/http.py
+++ b/dmr/security/http.py
@@ -154,10 +154,13 @@ class _HttpBasicAuth:  # noqa: WPS214
 
     def _get_custom_security_scheme_description(self) -> str:
         """Describe non-standard basic auth header contracts."""
+        # Empty `auth_scheme` means that the header carries
+        # the credentials alone, without a prefix and a space after it.
+        scheme_prefix = f'{self.auth_scheme} ' if self.auth_scheme else ''
         return (
             'HTTP Basic auth via '
             f'`{self.header}` header using '
-            f'`{self.auth_scheme!r} <base64(username:password)>` format'
+            f'`{scheme_prefix}<base64(username:password)>` format'
         )
 
 
--- a/dmr/security/http.py
+++ b/dmr/security/http.py
@@ -132,10 +132,7 @@ class _HttpBasicAuth:  # noqa: WPS214
 
     def _uses_standard_http_basic_auth(self) -> bool:
         """Whether the auth contract matches OpenAPI HTTP basic auth."""
-        return (
-            self.header == 'Authorization'
-            and self.auth_scheme == 'Basic'
-        )
+        return self.header == 'Authorization' and self.auth_scheme == 'Basic'
 
     def _get_custom_security_scheme_description(self) -> str:
         """Describe non-standard basic auth header contracts."""
--- a/dmr/security/http.py
+++ b/dmr/security/http.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Final, Self
 
 from typing_extensions import override
 
+from dmr.exceptions import NotAuthenticatedError
 from dmr.openapi.objects import Reference, SecurityRequirement, SecurityScheme
 from dmr.security.base import AsyncAuth, SyncAuth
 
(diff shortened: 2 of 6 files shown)


dmr/security/http.py
```diff
@@ -15,17 +16,27 @@
...
-        """Apply possible customizations."""
+        """
+        Apply possible customizations.
+
+        - *security_scheme_name* is the name
+          used in the OpenAPI security scheme map
+        - *header* selects the header to read the credentials from
+        - *auth_scheme* is the prefix that the header value must start with,
+          it is matched exactly, so ``Basic`` won't accept ``basic``
+        """
```
sobolevn, line 64: Please add `.. versionchanged` here as well.

dmr/security/http.py
```diff
@@ -57,35 +68,43 @@ def _get_username_and_password(
...
         except Exception:
-            return None
+            raise NotAuthenticatedError from None
         return unquote(username), unquote(password)
 
+    def _split_encoded_credentials(self, header: str) -> str | None:
+        """Splits string like 'Basic credentials' and returns 'credentials'."""
+        parts = header.split(' ')
+        if len(parts) != 2 or parts[0] != self.auth_scheme:
+            return None
```
sobolevn, line 148: Please, add a comment here to explain why we return `None` and do not raise

dmr/security/http.py
```diff
@@ -57,35 +68,43 @@ def _get_username_and_password(
...
+        return parts[1]
+
     def _uses_standard_http_basic_auth(self) -> bool:
         """Whether the auth contract matches OpenAPI HTTP basic auth."""
-        return self.header == 'Authorization'
+        return (
+            self.header == 'Authorization'
+            and self.auth_scheme.casefold() == 'basic'
```
sobolevn:
```suggestion
            and self.auth_scheme == 'Basic'
```

dmr/security/http.py
```diff
@@ -57,35 +68,43 @@ def _get_username_and_password(
...
+        )
 
     def _get_custom_security_scheme_description(self) -> str:
         """Describe non-standard basic auth header contracts."""
         return (
             'HTTP Basic auth via '
             f'`{self.header}` header using '
-            '`<base64(username:password)>` or '
-            '`Basic <base64(username:password)>` format'
+            f'`{self.auth_scheme} <base64(username:password)>` format'
```
sobolevn:
```suggestion
            f'`{self.auth_scheme!r} <base64(username:password)>` format'
```

tests/test_unit/test_security/test_http_basic_auth.py
```diff
@@ -1,8 +1,120 @@
...
+    auth = (
+        _SyncAuth(),
+        _SyncAuth(header=_FALLBACK_HEADER, security_scheme_name='fallback'),
+    )
+
+    def get(self) -> str:
+        return 'authed'
+
+
+class _ChainedAsyncController(Controller[PydanticSerializer]):
```
sobolevn: please, define controllers near tests that actually use them, this way it is easier to read.

docs/pages/auth/http-basic.rst
```diff
@@ -45,6 +45,27 @@ Any other authentication method will be better than the one above.
 Consider using :doc:`jwt` instead.
 
 
+Customization
+-------------
```
sobolevn, line 49:
```suggestion
-------------

.. versionadded:: 0.15.0
```

CHANGELOG.md
```diff
@@ -27,6 +27,14 @@ of requirements for an API to count as public.
   use `FileResponseSpec.return_type` instead, #1278
 - Removed `FileMetadataComponent.schema_metadata`,
   now we use `SupportsFileParsing.schema_metadata` instead, #1278
+- `HttpBasicSyncAuth` and `HttpBasicAsyncAuth` now require
```
sobolevn, line 51: Please, add a note about using `''` as a prefix, if people need the old behavior for no-prefix.

dmr/security/http.py
vyhuholl: Done.

tests/test_unit/test_security/test_http_basic_auth.py
vyhuholl: Done.

CHANGELOG.md
vyhuholl: Done.

dmr/security/http.py
```diff
@@ -57,35 +79,47 @@ def _get_username_and_password(
...
+        return self.header == 'Authorization' and self.auth_scheme == 'Basic'
 
     def _get_custom_security_scheme_description(self) -> str:
         """Describe non-standard basic auth header contracts."""
         return (
             'HTTP Basic auth via '
             f'`{self.header}` header using '
-            '`<base64(username:password)>` or '
-            '`Basic <base64(username:password)>` format'
+            f'`{self.auth_scheme!r} <base64(username:password)>` format'
```
sobolevn: hm, this does not seem right. my suggestion was not fully correct :) 

I saw that using `self.auth_scheme == ''` would produce two spaces in a row. And decided to try `!r`. But, it produces extra `''` :( 

What we need is to properly format `self.auth_scheme` not to have two spaces on `''`

dmr/security/http.py
vyhuholl: Done

dmr/security/http.py
vyhuholl: Replaced `!r` with the proper format.

sobolevn (approved):
Please, rebase

vyhuholl:
Rebased.

sobolevn (approved):

PR #1363, opened 2026-09-01
Fix HTTP basic auth URL decoding

### LINKED ISSUE #1331: `_HttpBasicAuth._get_username_and_password` must not unquote `%` chars
Currently, we incorrectly use `unquote` in the auth: https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/security/http.py#L60-L76

This is a problem, because now these two passwords are the same:

```python
>>> from base64 import b64decode, b64encode
>>> from urllib.parse import unquote

>>> unquote(b64decode(b64encode(b'test_%40')).decode())
'test_@'
>>> unquote(b64decode(b64encode(b'test_@')).decode())
'test_@'
```

But, the thing is that https://www.rfc-editor.org/info/rfc7617/ does not say that this unquoting is needed.
I just copied this errors from somewhere else. This needs to be fixed.

panguss:
# I have made things!

## AI Policy

## Checklist

## Related issues

- Closes #1331

### DIFF
--- a/tests/test_unit/test_security/test_http_basic_auth.py
+++ b/tests/test_unit/test_security/test_http_basic_auth.py
@@ -1,8 +1,18 @@
-import pytest
-from inline_snapshot import snapshot
+from http import HTTPStatus
+from typing import Final, Self, final
 
+import pytest
+from django.http import HttpResponse
+from inline_snapshot import snapshot
+from typing_extensions import override
+
+from dmr import Controller
+from dmr.endpoint import Endpoint
 from dmr.openapi.objects import SecurityScheme
-from dmr.security.http import HttpBasicAsyncAuth, HttpBasicSyncAuth
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.security.http import HttpBasicAsyncAuth, HttpBasicSyncAuth, basic_auth
+from dmr.serializer import BaseSerializer
+from dmr.test import DMRAsyncRequestFactory, DMRRequestFactory
 
 
 @pytest.mark.parametrize('typ', [HttpBasicSyncAuth, HttpBasicAsyncAuth])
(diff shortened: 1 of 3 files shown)


sobolevn (commented):
Please, add:
- a test case with explicit `%` value
- a changelog entry with a breaking change

panguss:
@sobolevn I got it

tests/test_unit/test_security/test_http_basic_auth.py
```diff
@@ -43,3 +49,34 @@ def test_custom_header_schema(
...
+
+
+@pytest.mark.parametrize('typ', [HttpBasicSyncAuth, HttpBasicAsyncAuth])
+def test_get_username_and_password_with_percent(
+    *,
+    typ: type[HttpBasicSyncAuth] | type[HttpBasicAsyncAuth],
+    dmr_rf: DMRRequestFactory,
+) -> None:
+    """Ensures that percent-encoded characters are not unquoted."""
+    credentials = b64encode(
```
sobolevn: Use `basic_auth` here :)

tests/test_unit/test_security/test_http_basic_auth.py
```diff
@@ -43,3 +49,34 @@ def test_custom_header_schema(
...
+
+
+@final
+class _Controller(Controller[PydanticSerializer]):
+    def get(self) -> str:
+        return 'test'
+
+
+@pytest.mark.parametrize('typ', [HttpBasicSyncAuth, HttpBasicAsyncAuth])
+def test_get_username_and_password_with_percent(
```
sobolevn: This test is not fully correct. What we need to do:
1. Create a real auth class with real parametrized expected login and password pairs
2. Create parametrized values for both login and password with `%`
3. Test that literal values match, but values with escaped `%` do not

tests/test_unit/test_security/test_http_basic_auth.py
```diff
@@ -43,3 +49,34 @@ def test_custom_header_schema(
...
+
+    request = dmr_rf.get(
+        '/',
+        headers={'Authorization': f'Basic {credentials}'},
+    )
+
+    controller = _Controller()
+    controller.request = request
+
+    assert typ()._get_username_and_password(controller) == (
```
sobolevn: This tests implementation, not the outcome :)

sobolevn (commented):
Thanks! Please, also fix a conflict

tests/test_unit/test_security/test_http_basic_auth.py
panguss: done!

tests/test_unit/test_security/test_http_basic_auth.py
panguss: ty for explanation, try to add

tests/test_unit/test_security/test_http_basic_auth.py
```diff
@@ -43,3 +53,103 @@ def test_custom_header_schema(
         ),
     })
     assert instance.security_requirement == snapshot({'http_basic': []})
+
+
+_USERNAME: Final = 'user%40name'
+_PASSWORD: Final = 'pass%3Aword'
```
sobolevn:
```suggestion
_PASSWORD: Final = 'pass%3Aword'  # noqa: S105
```

sobolevn (approved):
Nice! Thanks!

PR #1365, opened 2026-09-01
Document safe usage of `RedirectTo`

### LINKED ISSUE #1326: `RedirectTo` accepts protocol-relative URLs
https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/response.py#L164-L168

This might be a bug in Django as well.

```python
>>> from urllib.parse import urlsplit
>>> urlsplit('//evil.example/x').scheme
''
```

So, the `scheme` check is skipped.
Django has https://github.com/django/django/blob/73cc09f14f13fedddc14d6ba5b287cb33c24e4a4/django/utils/http.py#L274 for this case.
And this is how it is used: https://github.com/django/django/blob/73cc09f14f13fedddc14d6ba5b287cb33c24e4a4/django/contrib/auth/views.py#L43-L59

We need to add docs about `RedirectTo` usage. So, developers will know that redirects to users' paths are not always safe.

AndreyGatsuk:
`RedirectTo` validates a URL's length and scheme, but does not check whether its destination host is trusted. As a result, passing user-provided protocol-relative URLs such as `//evil.example/x` can cause an open redirect.

  This PR documents the risk and adds an example using Django's `url_has_allowed_host_and_scheme`. The example falls back to `/` when validation fails.

## AI Policy

## Checklist

## Related issues

Closes #1326

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,8 @@ of requirements for an API to count as public.
 ### Misc
 
 - Fixes AI docs and plugin install instructions, #1311
+- Documented safe use of user-provided redirect targets with `RedirectTo`,
+  #1326
 - Added `dmr-from-dj-rest-auth` agent skill to migrate `dj-rest-auth`
   installations to `django-modern-rest` and `django-allauth` headless, #1193
 
--- a/docs/examples/using_controller/redirect_safe.py
+++ b/docs/examples/using_controller/redirect_safe.py
@@ -1,0 +1,15 @@
+from typing import Never
+
+from django.http import HttpRequest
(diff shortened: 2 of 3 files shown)


sobolevn (approved):
Great stuff! Thanks!

sobolevn:
@AndreyGatsuk please, create a new issue / PR with the same change, but for the `/dmr` agent skill, it must be listed as well near the redirects section :)

AndreyGatsuk:
@sobolevn Sure, no problem :)

PR #1368, opened 2026-09-01
Document safe RedirectTo usage in DMR skill, #1367

### LINKED ISSUE #1367: Document safe RedirectTo usage in the /dmr agent skill
Follow-up to #1365, as suggested by @sobolevn.
  >
  > The /dmr agent skill currently does not mention that RedirectTo validates a URL's length and scheme, but does not check whether the destination host is trusted. Passing user-provided URLs, including protocol-relative URLs such as //evil.example/x, can therefore cause an open redirect.
  >
  > A redirects section should be added recommending that user-provided redirect targets be checked with Django's url_has_allowed_host_and_scheme, with a safe local fallback when validation fails.

AndreyGatsuk:
Follow-up to #1365, as suggested by @sobolevn.

  The `/dmr` agent skill now explains that `RedirectTo` validates a URL's length and scheme, but does not check whether the destination host is trusted. Passing user-provided URLs, including protocol-relative URLs such as `//evil.example/x`, can therefore cause an open redirect.

  This PR adds a redirects section recommending that user-provided redirect targets be checked with Django's `url_has_allowed_host_and_scheme`, with a safe local fallback when validation fails.

## AI Policy

## Checklist

## Related issues

Closes #1367

### DIFF
--- a/.agents/skills/dmr/SKILL.md
+++ b/.agents/skills/dmr/SKILL.md
@@ -314,6 +314,69 @@ Docs: https://django-modern-rest.readthedocs.io/en/latest/pages/error-handling.h
 
 ## Redirects
 
+### Choose a redirect compatible with the endpoint style
+
+Raise `RedirectTo` from `@modify` endpoints, because they cannot return
+`HttpResponse` subclasses. `@validate` endpoints support both raising
+`RedirectTo` and returning Django's `HttpResponseRedirect` equally. Returning
+`HttpResponseRedirect` from `@validate` is an exception to the general rule
+against returning Django responses directly.
+
+Whichever redirect style you use, document the `Location` header in its
+`ResponseSpec`. For `@modify`, add the spec to `extra_responses`; for
+`@validate`, pass the spec directly:
+
+```python
+from http import HTTPStatus
+from typing import Final
+
+from django.http import HttpResponse, HttpResponseRedirect
+
+from dmr import (
+    Controller,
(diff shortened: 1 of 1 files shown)


.agents/skills/dmr/SKILL.md
```diff
@@ -312,6 +312,52 @@ You can also use `self.to_error` when using `@validate` endpoints.
 Docs: https://django-modern-rest.readthedocs.io/en/latest/pages/error-handling.html
 
 
+## Redirects
```
sobolevn, line 315: Nice! But, we still need to show that `HttpResponseRedirect` is also supported for `@validate` responses. Specify that both can be used equaly.

Also, show how to define the `Location` header for both `@modify` and `@validate`.

Also, specify when to use which status codes, you can just provide a link to the spec and say that "select redirect status codes according to this spec"

sobolevn (commented):
Thanks!

.agents/skills/dmr/SKILL.md
AndreyGatsuk: Thanks! Addressed all points:
- clarified that `@validate` supports both `RedirectTo` and `HttpResponseRedirect` equally;
- showed how to document the `Location` header for both `@modify` and `@validate`;
- added a link to the HTTP Semantics specification for selecting the appropriate 3xx status code

sobolevn (approved):
Great! Thank you!

PR #1369, opened 2026-09-01
Raise `JWTokenError` from `JWToken.encode`

### LINKED ISSUE #1364: `JWToken.encode` raises `InternalServerError`, an HTTP-layer exception
### What's wrong?

`JWToken.encode` converts any `pyjwt` failure into `InternalServerError`:

https://github.com/wemake-services/django-modern-rest/blob/f7a8baebff6ae07e651ce2353637c121ca8c76f2/dmr/security/jwt/token.py#L138-L139

`InternalServerError` is an HTTP-layer exception: it carries `status_code = 500`, a translated `default_message`, is formatted by `format_error()` and is listed in `_default_handled_excs` in `dmr/errors.py`. But `JWToken` is a plain DTO, and `encode` is public API that is regularly called outside of any request cycle — a management command that issues a service token, a Celery task, a test factory, a script. In those places a "500 response" object means nothing, and to catch it the caller has to import an HTTP exception into non-HTTP code.

What actually triggers it is a developer/configuration mistake, not a broken runtime:

* an unknown or empty `algorithm` (`NotImplementedError` from `pyjwt`);
* a key that does not match the algorithm, e.g. an HMAC secret passed with `RS256` (`jwt.exceptions.InvalidKeyError`).

That is closer in meaning to `ValueError` or a dedicated JWT error than to "something is broken on our side, show a generic error page".

## Three different error types leave the same method

After #1361  `encode` can raise:

| Cause | What escapes today |
|---|---|
| `exp` in the past / `iat` in the future | `ValueError` (from `validate_issued_claims`) |
| bad algorithm or key | `InternalServerError` |
| `extras` value that is not JSON-serializable | bare `TypeError`, not handled at all |

The last one is easy to reproduce:

```python
JWToken(
    sub='a',
    exp=dt.datetime.now(dt.UTC) + dt.timedelta(days=1),
    extras={'x': object()},
).encode(secret, 'HS256')
# TypeError: Object of type object is not JSON serializable
```

So the "we always turn encoding problems into a 500" contract is not actually held.

### How it should be?

Raise a semantic error in the token layer, and convert it to `InternalServerError` at the HTTP boundary — most likely in `_BaseTokenController.create_jwt_token` in `dmr/security/jwt/views.py`, which is the only place inside the framework that encodes a token while serving a request.

Open questions for whoever picks this up:

1. **Which error?** A new dedicated exception (`JWTokenError` / `TokenEncodingError`), or a plain `ValueError` so that `encode` fails the same way for every kind of bad input?
2. **Where does it live?** `dmr/exceptions.py` next to `NotAuthenticatedError`, or inside `dmr/security/jwt/` since jwt support is an optional extra?
3. Should the same treatment apply to the `ValueError` from `validate_issued_claims`, so that the whole method has one error contract?
4. Should the non-serializable payload (`TypeError`) be caught and converted as well?

## Notes

* `raise ... from None` currently drops the original `pyjwt` exception, so the real cause is not in the traceback. Worth keeping the cause when this is reworked.
* `InternalServerError` is documented public API (`docs/pages/deep-dive/public-api.rst`), so changing what `encode` raises is a breaking change and needs a `CHANGELOG.md` entry under "Breaking changes".
* `tests/test_unit/test_security/test_jwt/test_jwt_token.py::test_encode_validation` asserts `InternalServerError` for bad algorithm/secret combinations and will have to be updated.
* Follow-up to #1361 

### Used versions

0.14.0

### OS information

macOS

vyhuholl:
# I have made things!

`JWToken.encode` no longer raises `InternalServerError`. It raises a new `JWTokenError`, and the views we ship convert that into `InternalServerError` themselves.

`JWToken` is a DTO, and tokens get created in management commands, background tasks and scripts, where a "500 response" object means nothing and forces non-HTTP code to import an HTTP exception. What actually goes wrong there is always a wrong value: an algorithm that does not exist, a key that does not match the algorithm, claims that cannot be issued.

* Added `JWTokenError` in `dmr/security/jwt/token.py`, exported from `dmr.security.jwt`.
* `JWToken.encode`, `JWToken.validate_issued_claims` and `JWToken.__post_init__` all raise it now, so the whole class has one error contract.
* `encode` also catches the `TypeError` from a payload that cannot be serialized to json — before it escaped raw, which contradicted the "encoding problems become a 500" promise the code made.
* The original error is kept as `__cause__` instead of being dropped with `from None`.
* `_BaseTokenController.create_jwt_token` catches `JWTokenError` and raises `InternalServerError` from it, so our own views behave as before from the outside.

## One change that is not strictly about the exception

`responses` on the six jwt controllers is now spelled `ClassVar[Sequence[ResponseSpec]]`.

Without it, `responses = (ResponseSpec(...),)` is inferred as a one-element tuple type and *no* subclass can add a spec — not even with an explicit annotation of its own. Since the whole point of this PR is that these views answer with a `500`, and declaring that `500` is what a user has to do to let it through response validation, leaving it unfixed would have made the new behaviour untestable and unusable. No runtime change. The same narrowing is still present in `dmr/security/token/views.py` and `dmr/security/django_session/views.py` — I left those alone and opened a separate issue, see #1371 

## Something for the reviewer to be aware of

The converted `500` does not currently reach the client anyway: response validation rejects any status code that is not declared, so it comes out as `422 "Returned status code 500 is not specified in the list of allowed status codes"`. This is not new and not caused by this PR — a plain controller that raises `InternalServerError` behaves the same way on `master`. I've opened a separate issue about it (see #1370); the test here declares the `500` on its own controller to work around it.

## AI Policy

## Checklist

## Related issues

Closes #1364
Refs #1361

### DIFF
--- a/dmr/security/jwt/token.py
+++ b/dmr/security/jwt/token.py
@@ -30,12 +30,35 @@
 import datetime as dt
 from collections.abc import Sequence
 from dataclasses import asdict, dataclass, field, fields
-from typing import Any, Self
+from typing import Any, Self, final
 
 import jwt
 from jwt.types import Options
 
-from dmr.exceptions import InternalServerError, NotAuthenticatedError
+from dmr.exceptions import NotAuthenticatedError
+
+
+@final
+class JWTokenError(ValueError):
+    """
+    Raised when a token cannot be created or encoded.
+
+    This is a semantic error about the token itself: its claims,
+    the signing algorithm, or the key. It is not an HTTP error,
+    because tokens are regularly created outside of any request:
+    in management commands, background tasks, and scripts.
+
+    We subclass :exc:`ValueError`, because it always means
+    that some value we were given is wrong.
+
+    Views that we ship convert it
+    into :exc:`dmr.exceptions.InternalServerError`,
+    since a token that our own server cannot sign
+    is a server-side problem.
+
+    .. versionadded:: 0.15.0
+
+    """
 
 
 @dataclass(frozen=True, slots=True)
@@ -82,7 +105,7 @@ class JWToken:  # noqa: WPS214
     def __post_init__(self) -> None:
         """Normalizes datetime claims and runs extra validation."""
         if len(self.sub) < 1:
-            raise ValueError(
+            raise JWTokenError(
                 'sub must be a string with a length greater than 0',
             )
 
@@ -97,16 +120,16 @@ class JWToken:  # noqa: WPS214
         the checks that we run before signing a token.
 
         Raises:
-            ValueError: If this token cannot be issued right now.
+            JWTokenError: If this token cannot be issued right now.
 
         .. versionadded:: 0.15.0
 
         """
         now = _normalize_datetime(dt.datetime.now(dt.UTC)).timestamp()
         if self.exp.timestamp() < now:
-            raise ValueError('exp value must be a datetime in the future')
+            raise JWTokenError('exp value must be a datetime in the future')
         if self.iat.timestamp() > now:
-            raise ValueError('iat must be a current or past time')
+            raise JWTokenError('iat must be a current or past time')
 
     def encode(
         self,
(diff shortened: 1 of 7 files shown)


dmr/security/jwt/token.py
```diff
@@ -30,12 +30,35 @@
...
 
 import jwt
 from jwt.types import Options
 
-from dmr.exceptions import InternalServerError, NotAuthenticatedError
+from dmr.exceptions import NotAuthenticatedError
+
+
+@final
+class JWTokenError(ValueError):
```
sobolevn, line 42: Our own exception must use `Exception` as the base type, otherwise it is possible to catch them with `except ValueError` - which is unexpected.

sobolevn:
Let's give https://github.com/wemake-services/django-modern-rest/pull/1373 a chance to contribute, since they are a first-time contributor :)

Thanks a lot for your work!

dmr/security/jwt/views.py
```diff
@@ -91,19 +91,38 @@ def create_jwt_token(  # noqa: WPS211
...
+                because here it always means that this server
+                is configured incorrectly.
+
+        .. versionchanged:: 0.15.0
+
+            Converts :exc:`dmr.security.jwt.token.JWTokenError`
+            into :exc:`dmr.exceptions.InternalServerError`,
+            previously :meth:`dmr.security.jwt.token.JWToken.encode`
+            raised it on its own.
+
```
sobolevn:
```suggestion
```

PR #1372, opened 2026-09-01
Allow changing `responses` of shipped controllers

### LINKED ISSUE #1371: `responses` of shipped controllers is narrowed to a fixed-size tuple
### What's wrong?

`Controller` declares `responses: ClassVar[Sequence[ResponseSpec]] = []`, but the controllers we ship assign a tuple literal without an annotation:

```python
responses = (
    ResponseSpec(return_type=ErrorModel, status_code=HTTPStatus.UNAUTHORIZED),
)
```

`mypy` then infers `tuple[ResponseSpec]` — a *one-element* tuple type — and any subclass that adds one more spec fails:

```
error: Incompatible types in assignment (expression has type "tuple[ResponseSpec, ResponseSpec]",
base class "ObtainTokensSyncController" defined the type as "tuple[ResponseSpec]")  [assignment]
```

Re-declaring the wider type in the subclass does not help either, because the base class has already narrowed it. Subclassing these controllers is a documented use case, and adding a response spec (a `500`, see the other issue, or any custom error) is the obvious thing to do in a subclass.

Still narrowed:

* `dmr/security/token/views.py` — `ObtainTokenSyncController`, `ObtainTokenAsyncController`
* `dmr/security/django_session/views.py` — `DjangoSessionSyncController`, `DjangoSessionAsyncController`

### How it should be?

Already fixed in the jwt views in #1369 by spelling the type out:

```python
responses: ClassVar[Sequence[ResponseSpec]] = (...)
```

The same one-line change applies to the four controllers above. It is worth checking whether other `ClassVar` options we ship (`auth`, `parsers`, `renderers`, `throttling`) have the same problem.

### Used versions

0.14.0

### OS information

macOS

vyhuholl:
# I have made things!

`responses` on the four controllers listed in the issue is now spelled out as `ClassVar[Sequence[ResponseSpec]]`, so a subclass can add or drop a spec. Nothing changes at runtime — this is purely about what type-checkers infer.

## What changed

* `dmr/security/token/views.py` — `ObtainTokenSyncController`, `ObtainTokenAsyncController`
* `dmr/security/django_session/views.py` — `DjangoSessionSyncController`, `DjangoSessionAsyncController`

## About the other `ClassVar` options

The issue asked whether `auth`, `parsers`, `renderers` and `throttling` have the same problem. **They do**, and I reproduced all four:

```python
class AuthBase(Controller[PydanticSerializer]):
    auth = (JWTSyncAuth(),)

class AuthChild(AuthBase):
    auth = (JWTSyncAuth(), CookieJWTSyncAuth())
```

```
error: Incompatible types in assignment (expression has type
"tuple[HeaderJWTSyncAuth, CookieJWTSyncAuth]", base class "AuthBase"
defined the type as "tuple[HeaderJWTSyncAuth]")  [assignment]
```

But there is nothing to fix for them on our side: no controller we ship assigns `auth`, `parsers`, `renderers` or `throttling` in a class body — `responses` is the only one. The trap is entirely in user code, in anyone's own base controller.

So instead of a code change I added a tip to the "Reusable controllers" docs, which is exactly where people build such hierarchies: annotate these options if the controller will be subclassed. The example there is executable — `.rst` files are collected as doctests (`--doctest-glob=*.rst`), and I checked that this one runs.

## The jwt views are deliberately not in this PR

They are being fixed in #1369, which is still open, and touching the same lines here would only produce a merge conflict. Once both land, every controller we ship is annotated, and the sentence saying so can be added to that docs tip — I left it out for now because it is not true on this branch yet.

## Tests

`typesafety/assert_type/test_security/test_shipped_responses.py` subclasses all four controllers and extends `responses` **without an annotation of its own**, which is what a user would write. Files under `typesafety/` are checked by all four type-checkers in `just type-check` and are excluded from pytest, so this is the kind of test that can actually catch this bug.

I checked that the test is not vacuous: with the source change reverted, `mypy` fails on both classes with the original `tuple[ResponseSpec]` error.

## AI Policy

## Checklist

## Related issues

Closes #1371

### DIFF
--- a/docs/pages/reusable-code.rst
+++ b/docs/pages/reusable-code.rst
@@ -58,6 +58,39 @@ Let's try to create two exact controllers with exact serializers:
 Basically - we just specify what kind of serializer to use. And that's it.
 But, this is just the first step. We can do much more!
 
+.. tip::
+
+  Annotate class-level options like ``responses``, ``auth``, ``parsers``,
+  ``renderers``, and ``throttling`` in controllers that will be subclassed.
+
+  Without an annotation type-checkers infer a fixed-size tuple
+  from the value you assign, and subclasses won't be able
+  to add or to remove items from it.
+
+  .. code:: python
+
+    >>> from collections.abc import Sequence
+    >>> from http import HTTPStatus
+    >>> from typing import ClassVar
+
+    >>> from dmr import Controller, ResponseSpec
+    >>> from dmr.errors import ErrorModel
+    >>> from dmr.plugins.pydantic import PydanticSerializer
+
+    >>> class MyBaseController(Controller[PydanticSerializer]):
+    ...     responses: ClassVar[Sequence[ResponseSpec]] = (
+    ...         ResponseSpec(ErrorModel, status_code=HTTPStatus.NOT_FOUND),
+    ...     )
+    ...
+    ...     def get(self) -> str:
+    ...         return 'reusable'
+
+    >>> class MyController(MyBaseController):
+    ...     responses = (
+    ...         *MyBaseController.responses,
+    ...         ResponseSpec(ErrorModel, status_code=HTTPStatus.CONFLICT),
+    ...     )
+
 
 Generic parsing and response models
 -----------------------------------
--- a/docs/pages/reusable-code.rst
+++ b/docs/pages/reusable-code.rst
@@ -67,29 +67,10 @@ But, this is just the first step. We can do much more!
   from the value you assign, and subclasses won't be able
   to add or to remove items from it.
 
-  .. code:: python
-
-    >>> from collections.abc import Sequence
-    >>> from http import HTTPStatus
-    >>> from typing import ClassVar
-
-    >>> from dmr import Controller, ResponseSpec
-    >>> from dmr.errors import ErrorModel
-    >>> from dmr.plugins.pydantic import PydanticSerializer
-
-    >>> class MyBaseController(Controller[PydanticSerializer]):
-    ...     responses: ClassVar[Sequence[ResponseSpec]] = (
-    ...         ResponseSpec(ErrorModel, status_code=HTTPStatus.NOT_FOUND),
-    ...     )
-    ...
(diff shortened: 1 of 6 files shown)


docs/pages/reusable-code.rst
```diff
@@ -58,6 +58,39 @@ Let's try to create two exact controllers with exact serializers:
...
+  Annotate class-level options like ``responses``, ``auth``, ``parsers``,
+  ``renderers``, and ``throttling`` in controllers that will be subclassed.
+
+  Without an annotation type-checkers infer a fixed-size tuple
+  from the value you assign, and subclasses won't be able
+  to add or to remove items from it.
+
+  .. code:: python
+
+    >>> from collections.abc import Sequence
```
sobolevn: this should go to examples, because we type check them. since this is related to annotations - type-checking is required.

docs/pages/reusable-code.rst
```diff
@@ -58,6 +58,39 @@ Let's try to create two exact controllers with exact serializers:
...
+    >>> class MyBaseController(Controller[PydanticSerializer]):
+    ...     responses: ClassVar[Sequence[ResponseSpec]] = (
+    ...         ResponseSpec(ErrorModel, status_code=HTTPStatus.NOT_FOUND),
+    ...     )
+    ...
+    ...     def get(self) -> str:
+    ...         return 'reusable'
+
+    >>> class MyController(MyBaseController):
+    ...     responses = (
```
sobolevn: please, add a comment here: that without the parent annotation it would raise an error. and provide the error text, so it would be greppable.

sobolevn (commented):
Great catch! Thank you!

docs/pages/reusable-code.rst
vyhuholl: Done

docs/pages/reusable-code.rst
vyhuholl: Done

sobolevn (approved):

PR #1373, opened 2026-09-01
fix(jwt): raise semantic JWTokenError from encode, convert at HTTP boundary (fixes #1364)

### LINKED ISSUE #1364: `JWToken.encode` raises `InternalServerError`, an HTTP-layer exception

fwh888:
## Summary
Fixes #1364. `JWToken.encode` now raises a token-layer `JWTokenError` instead of the HTTP-layer `InternalServerError`, so non-request callers (management commands, Celery, test factories) get a meaningful exception. The error is converted back to `InternalServerError` in `create_jwt_token` where a request is served.

## Decisions (issue open questions)
- New dedicated `JWTokenError` (not plain ValueError) so encoding failures are distinguishable from input-validation `ValueError`
- Lives in `dmr/security/jwt/` (jwt is an optional extra)
- `validate_issued_claims` ValueError contract unchanged
- Non-serializable payload `TypeError` also converted to `JWTokenError`
- Original pyjwt cause preserved (`from exc`, no more `from None`)

## Tests
- `test_encode_validation` now expects `JWTokenError`
- 156 jwt tests pass; ruff clean

Fixes #1364

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,8 @@ of requirements for an API to count as public.
 - Parsed request data is no longer stored as a local variable
   of the endpoint's frame, because it was shown
   in error reports of any endpoint, #1323
+- Fixed `JWToken.encode` raising a bare `TypeError`
+  when `extras` cannot be serialized to json, #1373
 - JWT auth, refresh, and verify now return `401` instead of `500`
   when the token subject cannot be a value of the user lookup field,
   for example a non-numeric `sub` with the default integer `pk`, #1284
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,13 @@ of requirements for an API to count as public.
 
 ### Breaking changes
 
+- `JWToken.encode` now raises `JWTokenError` (a token-layer semantic error)
+  instead of the HTTP-layer `InternalServerError` when encoding fails.
+  The error is converted back to `InternalServerError` at the HTTP boundary
+  in `_BaseTokenController.create_jwt_token`, so request-serving paths keep
+  their 500 contract while non-request callers (management commands, Celery
+  tasks, test factories) get a meaningful exception. The original `pyjwt`
+  cause is preserved in the traceback (no more `from None`).
 - Renamed `json_dump` to `json_dumps` in `dmr.openapi.dump` and `dmr.internal.json`
   to follow standard string-serialization conventions, #1399
 - Removed `QueryTokenSyncAuth` and `QueryTokenAsyncAuth` auth classes,
--- a/dmr/security/jwt/__init__.py
+++ b/dmr/security/jwt/__init__.py
@@ -23,3 +23,4 @@ from dmr.security.jwt.auth.header import HeaderJWTSyncAuth as HeaderJWTSyncAuth
 from dmr.security.jwt.auth.header import JWTAsyncAuth as JWTAsyncAuth
 from dmr.security.jwt.auth.header import JWTSyncAuth as JWTSyncAuth
 from dmr.security.jwt.token import JWToken as JWToken
+from dmr.security.jwt.token import JWTokenError as JWTokenError
--- a/dmr/security/jwt/__init__.py
+++ b/dmr/security/jwt/__init__.py
@@ -23,4 +23,3 @@ from dmr.security.jwt.auth.header import HeaderJWTSyncAuth as HeaderJWTSyncAuth
 from dmr.security.jwt.auth.header import JWTAsyncAuth as JWTAsyncAuth
 from dmr.security.jwt.auth.header import JWTSyncAuth as JWTSyncAuth
 from dmr.security.jwt.token import JWToken as JWToken
-from dmr.security.jwt.token import JWTokenError as JWTokenError
--- a/dmr/security/jwt/token.py
+++ b/dmr/security/jwt/token.py
@@ -40,7 +40,16 @@ from dmr.exceptions import NotAuthenticatedError
 
 @final
 class JWTokenError(Exception):
-    """Raised when a token cannot be encoded."""
+    """
+    Raised when a token cannot be created, encoded, or decoded.
+
+    This is a semantic error about the token itself: its claims,
+    the signing algorithm, or the key. It is not an HTTP error,
+    because tokens are regularly created outside of any request:
+    in management commands, background tasks, and scripts.
+
+    .. versionadded:: 0.15.0
+    """
 
 
 @dataclass(frozen=True, slots=True)
--- a/dmr/security/jwt/token.py
+++ b/dmr/security/jwt/token.py
@@ -30,12 +30,17 @@
 import datetime as dt
 from collections.abc import Sequence
 from dataclasses import asdict, dataclass, field, fields
-from typing import Any, Self
+from typing import Any, Self, final
 
 import jwt
 from jwt.types import Options
 
-from dmr.exceptions import InternalServerError, NotAuthenticatedError
+from dmr.exceptions import NotAuthenticatedError
+
+
+@final
+class JWTokenError(Exception):
+    """Raised when a token cannot be encoded."""
 
 
 @dataclass(frozen=True, slots=True)
@@ -97,16 +102,16 @@ class JWToken:  # noqa: WPS214
         the checks that we run before signing a token.
 
         Raises:
-            ValueError: If this token cannot be issued right now.
+            JWTokenError: If this token cannot be issued right now.
 
         .. versionadded:: 0.15.0
 
         """
         now = _normalize_datetime(dt.datetime.now(dt.UTC)).timestamp()
         if self.exp.timestamp() < now:
-            raise ValueError('exp value must be a datetime in the future')
+            raise JWTokenError('exp value must be a datetime in the future')
         if self.iat.timestamp() > now:
-            raise ValueError('iat must be a current or past time')
+            raise JWTokenError('iat must be a current or past time')
 
     def encode(
         self,
@@ -127,14 +132,18 @@ class JWToken:  # noqa: WPS214
             An encoded token string.
 
         Raises:
-            ValueError: If this token cannot be issued right now.
-            InternalServerError: If encoding fails.
+            JWTokenError: If the token cannot be issued right now
+                (`exp`/`iat` validation) or encoding fails. pyjwt errors
+                are wrapped and the original exception is preserved
+                as the cause.
 
         .. versionchanged:: 0.15.0
 
             ``exp`` and ``iat`` are validated here
             via :meth:`validate_issued_claims`,
             previously it was done during the instance creation.
+            Encoding failures now raise :class:`JWTokenError`
+            instead of the HTTP-layer ``InternalServerError``.
 
         """
         self.validate_issued_claims()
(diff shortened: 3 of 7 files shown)


dmr/security/jwt/token.py
```diff
@@ -128,7 +140,7 @@ def encode(
 
         Raises:
             ValueError: If this token cannot be issued right now.
```
sobolevn: Let's remove `ValueError` from `validate_issued_claims` and only use `JWTokenError` from now on :)

dmr/security/jwt/views.py
```diff
@@ -92,18 +92,23 @@ def create_jwt_token(  # noqa: WPS211
...
-            aud=audiences or self.jwt_audiences,
-            jti=jwt_id or self.make_jwt_id(),
-            extras={'type': token_type} if token_type else {},
-        ).encode(
-            secret=secret or self.jwt_secret or settings.SECRET_KEY,
-            algorithm=algorithm or self.jwt_algorithm,
-            headers=token_headers,
-        )
+        try:
+            return self.jwt_token_cls(
```
sobolevn: Let's split this into two parts: `self.jwt_token_cls(...)` can't trigger any errors, so it must not be protected.

CHANGELOG.md
```diff
@@ -21,6 +21,16 @@ of requirements for an API to count as public.
...
+- `JWToken.encode` now raises `JWTokenError` (a token-layer semantic error)
+  instead of the HTTP-layer `InternalServerError` when encoding fails.
+  The error is converted back to `InternalServerError` at the HTTP boundary
+  in `_BaseTokenController.create_jwt_token`, so request-serving paths keep
+  their 500 contract while non-request callers (management commands, Celery
+  tasks, test factories) get a meaningful exception. The original `pyjwt`
+  cause is preserved in the traceback (no more `from None`).
+
+### Breaking changes
+
```
```suggestion
```

these lines are not required.

dmr/security/jwt/token.py
```diff
@@ -35,7 +35,19 @@
 import jwt
 from jwt.types import Options
 
-from dmr.exceptions import InternalServerError, NotAuthenticatedError
+from dmr.exceptions import NotAuthenticatedError
+
+
+class JWTokenError(Exception):
```
sobolevn, line 42:
```suggestion
@final
class JWTokenError(Exception):
```

dmr/security/jwt/token.py
```diff
@@ -35,7 +35,19 @@
...
+    """Raised when a token cannot be encoded.
+
+    ``JWToken`` is a plain DTO and ``encode`` is public API that is regularly
+    called outside any request cycle (management commands, Celery tasks, test
+    factories). HTTP-layer exceptions like :class:`InternalServerError` are
+    meaningless there; this semantic error is converted to
+    ``InternalServerError`` at the HTTP boundary (see
+    ``_BaseTokenController.create_jwt_token``).
+    """
```
sobolevn, line 52:
```suggestion

    .. versionadded:: 0.15.0
    """
```

dmr/security/jwt/token.py
```diff
@@ -35,7 +35,19 @@
...
-from dmr.exceptions import InternalServerError, NotAuthenticatedError
+from dmr.exceptions import NotAuthenticatedError
+
+
+class JWTokenError(Exception):
+    """Raised when a token cannot be encoded.
+
+    ``JWToken`` is a plain DTO and ``encode`` is public API that is regularly
+    called outside any request cycle (management commands, Celery tasks, test
+    factories). HTTP-layer exceptions like :class:`InternalServerError` are
```
sobolevn: this reference is not correct, use 
```suggestion
    factories). HTTP-layer exceptions like :class:`dmr.exceptions.InternalServerError` are
```

dmr/security/jwt/token.py
```diff
@@ -35,7 +35,19 @@
 import jwt
 from jwt.types import Options
 
-from dmr.exceptions import InternalServerError, NotAuthenticatedError
+from dmr.exceptions import NotAuthenticatedError
+
+
+class JWTokenError(Exception):
+    """Raised when a token cannot be encoded.
```
sobolevn:
```suggestion
    """
    Raised when a token cannot be encoded.
```

dmr/security/jwt/token.py
sobolevn, line 153: Please, update `.. versionchanged` docs here as well.

sobolevn (commented):
Thank you! Please, copy several tests from https://github.com/wemake-services/django-modern-rest/pull/1369

1. https://github.com/wemake-services/django-modern-rest/pull/1369/changes#diff-c780dfea89f1d7c2cd3845fc53a7b6c391af0c5b7a9d9ea9c2c5f7e650ba16e0
2. `test_encode_keeps_the_original_error` and `test_encode_unserializable_extras`

And changelog entries :)

fwh888:
Thanks for the detailed review @sobolevn! I've addressed all 8 points in commit `e15ac149`:

1. **Remove `ValueError` from `validate_issued_claims`** — now raises `JWTokenError` only, matching `encode`.
2. **Split the try scope in `create_jwt_token`** — `self.jwt_token_cls(...)` is no longer protected; only `.encode()` is.
3. **CHANGELOG** — merged the duplicate `### Breaking changes` headers.
4. **`@final` on `JWTokenError`** — added, with `from typing import final`.
5. **Docstring cleanup** — simplified `JWTokenError` docstring, fixed the `InternalServerError` reference to `dmr.exceptions.InternalServerError`, and removed the redundant `versionadded` blank block.
6. **`.. versionchanged` update** — documents that encoding failures now raise `JWTokenError` with the original error preserved as `__cause__`.

New tests added (including the ones you pointed to in #1369):
- `test_encode_keeps_the_original_error` — asserts the original `PyJWTError` is preserved as the cause.
- `test_encode_unserializable_extras` — asserts a non-serializable payload raises `JWTokenError` with a `TypeError` cause.
- `tests/test_unit/test_security/test_jwt/test_jwt_broken_encoding.py` — views-level test that an unsignable token is converted to a 500 response.

Local verification: `test_jwt_token.py` + `test_jwt_broken_encoding.py` (80 tests) pass, token.py is fully covered, ruff and `ty check` are clean.

fwh888:
Hi @sobolevn 👋 — heads-up on the two remaining CI items after the rebase:

- **pre-commit.ci** ✅ now green (the earlier error was the branch being based on the pre-refactor `main`; after rebasing onto `7e7b478` the `auth.base` import path matches and everything passes, including ruff-check/ruff-format).
- **readthedocs** still reports failure, but this PR touches **zero** files under `docs/` — only `dmr/security/jwt/*`, its tests, and `CHANGELOG.md`. The same readthedocs failure is present on PR #1355 (build 34341393), which suggests it's an upstream build-environment/example issue rather than something introduced here. `dmr.security.jwt` imports fine and `docs/conf.py` autodoc references resolve locally.

The PR is now `mergeable` with zero conflicts. Happy to pick up anything else you'd like adjusted.

dmr/security/jwt/__init__.py
```diff
@@ -23,3 +23,4 @@
 from dmr.security.jwt.auth.header import JWTAsyncAuth as JWTAsyncAuth
 from dmr.security.jwt.auth.header import JWTSyncAuth as JWTSyncAuth
 from dmr.security.jwt.token import JWToken as JWToken
+from dmr.security.jwt.token import JWTokenError as JWTokenError
```
sobolevn:
```suggestion
```

dmr/security/jwt/token.py
```diff
@@ -30,12 +30,17 @@
...
 import jwt
 from jwt.types import Options
 
-from dmr.exceptions import InternalServerError, NotAuthenticatedError
+from dmr.exceptions import NotAuthenticatedError
+
+
+@final
+class JWTokenError(Exception):
+    """Raised when a token cannot be encoded."""
```
sobolevn:
```suggestion
    """
    Raised when a token cannot be created, encoded, or decoded.

    This is a semantic error about the token itself: its claims,
    the signing algorithm, or the key. It is not an HTTP error,
    because tokens are regularly created outside of any request:
    in management commands, background tasks, and scripts.

    .. versionadded:: 0.15.0
    """
```

CHANGELOG.md
sobolevn, line 136:
```suggestion
- Fixed `JWToken.encode` raising a bare `TypeError`
  when `extras` cannot be serialized to json, #1373
- JWT auth, refresh, and verify now return `401` instead of `500`
```

tests/test_unit/test_security/test_jwt/test_jwt_token.py
```diff
@@ -13,8 +13,8 @@
 from faker import Faker
 from typing_extensions import TypedDict, override
 
-from dmr.exceptions import InternalServerError, NotAuthenticatedError
-from dmr.security.jwt import JWToken
+from dmr.exceptions import NotAuthenticatedError
+from dmr.security.jwt import JWToken, JWTokenError
```
sobolevn:
```suggestion
from dmr.security.jwt.token import JWToken, JWTokenError
```

sobolevn (approved):

PR #1374, opened 2026-09-01
Do not require describing our own error responses

### LINKED ISSUE #1370: `InternalServerError` raised by a handler is returned as `422`
### What's wrong?

`InternalServerError` is public documented API, and it is listed in `_default_handled_excs` in `dmr/errors.py`, so it looks like a supported way to say "this request failed on our side". But with default settings a handler that raises it never produces a `500`:

```python
class BrokenController(Controller[PydanticFastSerializer]):
    def get(self) -> str:
        raise InternalServerError('something broke')
```

```
STATUS: 422
{"detail":[{"msg":"Returned status code 500 is not specified in the list of
allowed status codes: [200, 422, 406]","type":"value_error"}]}
```

What happens: the global error handler builds the `500` response, response validation (`dmr/validation/response.py`, `_get_response_schema`) sees a status code that is not declared in `responses`, raises `ResponseSchemaError`, and that error carries `status_code = HTTPStatus.UNPROCESSABLE_ENTITY`. So the client is told that *its request* was unprocessable, while the real error is gone.

`Settings.validate_responses` defaults to `True`, so this is what happens out of the box.

This also hits the views we ship. `ObtainTokensSyncController.create_jwt_token` converts `JWTokenError` into `InternalServerError` (#1369), but the controller only declares `401`, so a server that cannot sign its own tokens answers:

```
422 {"detail":[{"msg":"Returned status code 500 is not specified in the list of
allowed status codes: [401, 200, 400, 422, 406]","type":"value_error"}]}
```

## What already exists

This is not undiscovered territory:

* `Settings.responses` is documented in `docs/pages/configuration.rst` with a `500` `ResponseSpec` as *the* example, so declaring `500` globally is the intended way to make it appear in the schema;
* the docs recommend switching `validate_responses` off in production.

So the mechanism is by design. What feels wrong is the interaction: the framework's own exception, raised by the framework's own views, is turned into a misleading `422` unless the user knows to declare `500` first.

### How it should be?

I see four possible directions:

1. Let the responses produced by `global_error_handler` for our own handled exceptions skip the declared-status check — they are our errors, not user data that needs validating.
2. Add `500` to semantic responses whenever the endpoint can actually raise `InternalServerError`.
3. Keep the behaviour, but make the message say what to do (`declare it in Settings.responses`) instead of only what is wrong.
4. At the very least, make the auth views we ship declare the `500` they can produce.

Reproducing needs nothing but the snippet above, no auth and no jwt.

### Used versions

0.14.0

### OS information

macOS

vyhuholl:
# I have made things!

Errors that we raise and handle ourselves are no longer turned into a `422` when the user has not listed their status code in `responses`. The reproducer from the issue now answers what it should:

```
STATUS: 500 {"detail":[{"msg":"Internal server error"}]}
```

This is direction 1 from the issue: responses built by `global_error_handler` for our own handled exceptions skip the declared-status check.

## How

The whole question is how to tell "our response to our own exception" apart from a response the user produced. The mark is set in exactly one place — the `_default_handled_excs` branch of `global_error_handler` — and nowhere else:

* `dmr/internal/errors.py` — `mark_handled_error()` / `is_handled_error()` over a `__dmr_handled_error__` attribute, the same trick the codebase already uses for `__dmr_auth__` and `__dmr_endpoint__`. It lives in `internal/` so that `dmr/errors.py` and `dmr/validation/response.py` don't have to import protected names from each other.
* `dmr/errors.py` — marks the response it builds for a handled exception.
* `dmr/validation/response.py` — `_is_undescribed_handled_error()` lets such a response through.

Deliberately **not** marked: `Controller.to_error()` itself, and the error handling path as a whole. `to_error` is public API that people call from their own handlers, and those responses have to keep being validated like any other response of their endpoint.

## The skip is narrow on purpose

Validation is skipped only when the status code is **not described**. If `500` is declared — per controller, or globally via `Settings.responses` — everything works exactly as before, body included. So the self-checks we already have for described statuses (`401` from auth, `429` from throttling) are untouched, and this is the smallest behavioural change that fixes the report.

The OpenAPI schema does not change: `500` still shows up only if the user declares it.

## Tests

`tests/test_unit/test_endpoint/test_error_response_validation.py`:

* an undescribed `500` from a sync and from an async endpoint — now a real `500`;
* a described `500` — still a `500`, nothing regressed;
* a `500` described with the *wrong* type (`list[int]`) — comes back as `422`, which proves that describing a status code brings the validation back rather than silently disabling it;
* a custom `handle_error` returning an undescribed `501` — still `422`, which pins down the boundary of this change.

The existing `test_validate_status_code` (a user endpoint returning an undeclared `200`) stays green: user responses are validated exactly as before.

## Docs

A note in `docs/pages/error-handling.rst`, next to the description of the three handler levels: our errors don't have to be described, yours do, and describing ours is still worth it because it puts them in the OpenAPI schema and turns validation back on for them.

## Worth knowing for #1369

That PR converts `JWTokenError` into `InternalServerError` inside `create_jwt_token`, and I noted there that the resulting `500` did not actually reach the client. This is the fix for that: once both land, a server that cannot sign its own tokens answers `500` instead of a misleading `422`, without anyone having to declare anything first.

## AI Policy

## Checklist

## Related issues

Closes #1370

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,9 @@ of requirements for an API to count as public.
 
 ### Bugfixes
 
+- Fixed `InternalServerError` and other errors we handle ourselves
+  being returned as `422` when their status code
+  was not listed in `responses`, #1370
 - Added missing `@sensitive_variables` decorator to all auth views,
   so credentials and tokens are hidden
   in error reporting middlewares and logs, #1323
--- a/dmr/errors.py
+++ b/dmr/errors.py
@@ -26,6 +26,7 @@ from dmr.exceptions import (
     TooManyRequestsError,
     ValidationError,
 )
+from dmr.internal.errors import mark_handled_error
 
 if TYPE_CHECKING:
     from dmr.controller import Controller
@@ -305,11 +306,13 @@ def global_error_handler(
 
     """
     if isinstance(exc, _default_handled_excs):
-        return controller.to_error(
-            controller.format_error(exc),
-            status_code=exc.status_code,
-            headers=getattr(exc, 'headers', None),
-            cookies=getattr(exc, 'cookies', None),
-            renderer=getattr(exc, 'renderer', None),
+        return mark_handled_error(
+            controller.to_error(
+                controller.format_error(exc),
+                status_code=exc.status_code,
+                headers=getattr(exc, 'headers', None),
+                cookies=getattr(exc, 'cookies', None),
+                renderer=getattr(exc, 'renderer', None),
+            ),
         )
     raise  # noqa: PLE0704
--- a/dmr/internal/errors.py
+++ b/dmr/internal/errors.py
@@ -1,0 +1,23 @@
+from typing import TypeVar
+
+from django.http import HttpResponseBase
+
+_ResponseT = TypeVar('_ResponseT', bound=HttpResponseBase)
+
+
+def mark_handled_error(response: _ResponseT) -> _ResponseT:
+    """
+    Mark a response that we have built for an exception that we handle.
+
+    Users describe the responses that *their* endpoints return.
+    Errors that we raise and handle ourselves are not a part of that
+    contract, so response validation lets them through
+    even when their status code is not described.
+    """
+    response.__dmr_handled_error__ = True  # type: ignore[attr-defined]
+    return response
+
+
+def is_handled_error(response: HttpResponseBase) -> bool:
+    """Tells whether this response was built by our own error handling."""
+    return getattr(response, '__dmr_handled_error__', False)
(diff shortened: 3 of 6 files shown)


sobolevn (commented):
I don't think that this approach is correct. How is it different from `raise TooManyRequestsError`? This way we would allow raising errors that are not in the spec.

But, I agree that people might want to skip only some errors codes validation. For example, `500`. 

So, what we need?
1. Document that if you are running your production with `validate_responses=True`, you must explicitly define `500` as the possible error code
2. Provide `exclude_validate_responses` option, so you can disable validation for specific `frozenset` of error codes. Like `500`, default should be empty.

What do you think?

vyhuholl:
I like this idea. I'll work on it soon.

PR #1375, opened 2026-09-01
Document how to write your own auth class, #1366

### LINKED ISSUE #1366: Document how to write your own auth class
# Problem

We document every auth backend we ship (:doc:`http-basic`, :doc:`django-session`, :doc:`jwt`, :doc:`token`), but we never explain how to write an auth class from scratch.

Right now a user who needs a transport we don't ship has to read our source code to find out what to implement.

This came up in #1359: the `WWW-Authenticate` docs started growing a "Writing your own auth" section that only covered `www_authenticate_challenge`. That is one piece of a much bigger picture, and documenting it alone would be misleading. So it was removed, and the whole topic deserves a page of its own.

# What the guide should cover

The full contract of `SyncAuth` / `AsyncAuth`:

1. `__call__` — the auth decision itself:
   - return `self` on success
   - return `None` to fall through to the next auth in the chain
   - raise `NotAuthenticatedError` to stop the chain immediately
   - raise `APIError` to return some other status code
   - why the difference between returning `None` and raising matters for chains, and how #1289 got this wrong for cookie auth
2. `security_schemes` — how the auth appears in the OpenAPI spec
3. `security_requirement` — how endpoints reference that scheme
4. `www_authenticate_challenge` — the `401` challenge, and when it has to be `None` (added in #1334)
5. `provide_response_specs` — extra responses this auth can produce, like the `403` that CSRF-protected cookie auth adds
6. What to set on the request: `request.user`, `request.auser`, and the `__dmr_*` attributes read by `request_auth` and friends

Plus the rules that are easy to miss:

- Instances must be stateless, they are shared across requests and can live in settings
- Sync and async classes are separate, and there is `SyncOrAsyncAuth` for settings that serve both
- `__init__` must work with no required arguments
- `__slots__` is mandatory, we check it in CI
- Credentials must be kept out of error reports, see #1323

# Suggested shape

A new `docs/pages/auth/custom.rst` page, linked from the "Next up" grid in `docs/pages/auth/common.rst`, built around one worked example.

A good example would be an auth transport we deliberately don't ship, so the page is useful rather than a copy of an existing backend.

# Notes

`docs/pages/auth/common.rst` currently sends people to `SyncAuth` / `AsyncAuth` autodoc for this. That is a reference, not a guide, and it does not explain the interactions between the parts.

vyhuholl:
# I have made things!

Adds a guide built around one worked example: authenticating requests that a reverse proxy has already checked and passes down in a header. It is a transport we deliberately don't ship, so the page teaches the contract instead of repeating an existing backend.

Covers `__call__` and why returning `None` differs from raising, both `security_*` properties, `provide_response_specs`, the request attributes a successful auth sets, the sync and async split, and the rules that are easy to miss.

## AI Policy

## Checklist

## Related issues
Closes #1366


docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+  Before writing a new auth class, check whether you only need to change
+  *where* the credentials come from. Most of our classes let you swap
+  the header, the cookie, or the model without any new code.
+
+
+The contract
+------------
+
+Your auth class subclasses :class:`~dmr.security.SyncAuth`
+or :class:`~dmr.security.AsyncAuth` and implements three things:
```
sobolevn:
```suggestion
or :class:`~dmr.security.AsyncAuth` and implements several things:
```

docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+  :header-rows: 1
+  :widths: 30 70
+
+  * - Member
+    - What it does
+  * - :meth:`~dmr.security.SyncAuth.__call__`
+    - Decides whether this request is authenticated.
+  * - :meth:`~dmr.security.SyncAuth.security_schemes`
+    - Describes the auth itself in the OpenAPI spec.
+  * - :meth:`~dmr.security.SyncAuth.security_requirement`
```
sobolevn, line 36: `www_authenticate_challenge` abstract property is missing from this list :)

docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+    - References that description from every endpoint using this auth.
+
+There is a fourth, optional one:
```
sobolevn:
```suggestion
There is one more optional method:
```

docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+:meth:`~dmr.metadata.ResponseSpecProvider.provide_response_specs`
+declares extra responses your auth can produce.
+We cover it :ref:`below <auth-extra-responses>`.
+
+
+A worked example
+----------------
```
sobolevn:
```suggestion
``X-Forwarded-User`` example
----------------------------
```

docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+    - Authentication succeeded, we stop and run the endpoint.
+  * - ``return None``
+    - This auth does not apply, we try the next one in the chain.
+      When it is the last one, the request gets a ``401``.
+  * - raise :exc:`~dmr.exceptions.NotAuthenticatedError`
+    - Authentication failed, we stop the chain right there
+      and return a ``401``.
+  * - raise :exc:`~dmr.response.APIError`
+    - Same, but with a status code of your choice.
+      Use it when the credentials are malformed rather than wrong.
```
sobolevn:
```suggestion
      Use it for unexpected cases.
```

docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+
+Every auth automatically documents the ``401`` it can produce.
+If yours can fail in some other way, say so by overriding
+:meth:`~dmr.metadata.ResponseSpecProvider.provide_response_specs`,
+and the OpenAPI schema will list it:
+
+.. code:: python
+
+  @override
+  def provide_response_specs(
```
sobolevn: this needs to be in example file, the code is too complex to keep here, it is not tested / checked by mypy.

docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+If your auth resolves something else worth keeping, like a token row or
+a session, store it under a name of your own and give your users
+a helper to read it back. That is exactly what
+:func:`~dmr.security.jwt.auth.request_jwt`
+and :func:`~dmr.security.token.request_token` do.
+
+
+.. _auth-extra-responses:
+
+Extra responses
```
sobolevn: this should come after `async` auth, which is more common and more important

docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+else, including the whole OpenAPI description, comes from
+``BaseProxyHeaderAuth``.
+
+When you set auth globally and your project has both kinds of
+endpoints, wrap the pair in :class:`~dmr.security.SyncOrAsyncAuth`:
+
+.. code-block:: python
+  :caption: settings.py
+
+  from dmr.settings import Settings
```
sobolevn: please, convert to a doctest

docs/pages/auth/custom.rst
```diff
@@ -0,0 +1,300 @@
...
+
+One instance serves every request. It can live in ``DMR_SETTINGS`` and
+be shared by every controller in the project, across threads and
+coroutines.
+
+So configuration set in ``__init__`` is fine, but anything per-request
+is not. Not even a lock. Put per-request data on the request, like
+``set_request_attrs`` does above.
+
+``__init__`` must work with no arguments
```
sobolevn: this is not a rule, you can require parameters, this is your auth after all.

docs/pages/auth/custom.rst
vyhuholl: Added.

docs/pages/auth/custom.rst
vyhuholl: Done

docs/pages/auth/custom.rst
vyhuholl: Done

docs/pages/auth/custom.rst
vyhuholl: Removed

docs/pages/auth/custom.rst
vyhuholl: Moved to example file.

sobolevn (approved):
Very useful doc, thanks!

PR #1376, opened 2026-09-02
Validate SSE `id` and `event` fields in the stream validator

### LINKED ISSUE #1329: Make sure that `SSE` validation for events work correctly, including custom types and `validate_events=False`
https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/streaming/sse/metadata.py#L169-L175

Currently we have two problems:
1. Even when `validate_events=False`, we still validate that `id` and `event` fields for `SSEvent` does not contain `NULL` char and does not contain multiline strings. Which is not really cool. Why? Beceause we slow things down in production, when users explicitly ask us not to. We need to move the validation somewhere else. I propose moving this login into the validator or renderer. But, it must respect the setting. Even if some field is not valid in production, it must not be validated if `validate_events=False` 
2. Currently custom SSE event types are not validated the same way. It would be automatically solved, when `1.` is fixed. We would just need more tests for this :) 

vyhuholl:
# I have made things!

1. `SSEvent.__init__` called `check_event_field(id)` / `check_event_field(event)` unconditionally, so the null-byte and line-break checks ran even with `validate_events=False` — exactly what that flag is supposed to turn off.
2. Custom event types (anything respecting the `SSE` protocol rather than subclassing `SSEvent`) were never checked at all.

I moved the check out of the constructor and into the SSE validation pipeline, which is the one place that already knows about `validate_events` (`StreamingValidator.apply_event_pipeline` short-circuits when it is `False`). The renderer was the other candidate from the issue, but it has no access to that setting and cannot be extended per controller — the validator can.

Both problems collapse into one fix: the pipeline receives whatever the user yields, so custom types go through the same check as `SSEvent`.

## AI Policy

## Checklist

## Related issues

Closes #1329


sobolevn (approved):
Nice! Thank you!

PR #1377, opened 2026-09-02
Require jti when the JWT blocklist is used

### LINKED ISSUE #1322: The JWT blocklist is silently bypassed by tokens without a `jti`
There are two problem with `jti` and `blocklist` app:

1. We ignore cases where `jti` is `None` in the payload. `None` can't be found here: `self.blocklist_model().objects.filter(jti=token.jti).exists()`, so this check always passes. Moreover, it does not make sence to use tokens without `jti` and `blocklist` app
2. We don't check that token is created with a valid `jti` when `blocklist` app is used. We must do that, so this won't potentially fail on tokens with `jti=None`:

https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/security/jwt/blocklist/auth.py#L59-L70

So, the logout path raises `IntegrityError` (HTTP 500) rather than a clean error.
We need to add `['jti']` to `self.require_claims` with `JWTokenBlocklistSyncMixin` and `JWTokenBlocklistAsyncMixin`.

vyhuholl:
# I have made things!

Two separate problems, both in `dmr/security/jwt/blocklist/auth.py`:

1. **Silent bypass.** `check_auth` did `blocklist_model().objects.filter(jti=token.jti).exists()`. With `token.jti is None` this compiles to `WHERE jti IS NULL`, and the `jti` column is `unique=True` and non-nullable, so the lookup can never match. The check always passed — a token without `jti` was permanently un-blocklistable while still authenticating fine.
2. **500 on logout.** `blocklist()` did `get_or_create(jti=None, ...)`, which hits the `NOT NULL` constraint and raises `IntegrityError` → HTTP 500 instead of a clean auth error.

### 1. Require `jti` at decode time (the actual security fix)

`_BaseBlocklistMixin.__init__` now appends `'jti'` to `require_claims` of whatever auth class the mixin is mixed into, preserving claims the user passed and not duplicating `jti` if it is already there:

```python
def __init__(self, *args: Any, **kwargs: Any) -> None:
    super().__init__(*args, **kwargs)
    require_claims = list(self.require_claims or ())
    if _JTI_CLAIM not in require_claims:
        require_claims.append(_JTI_CLAIM)
    self.require_claims = require_claims
```

This is exactly what #1322  asked for. It works because `require_claims` flows into `pyjwt`'s `options['require']`, and `pyjwt` rejects a claim that is either absent **or** explicitly `null` (`_validate_required_claims` checks `payload.get(claim) is None`). So `{"jti": null}` is rejected too, and the existing `decode` error path turns `MissingRequiredClaimError` into `NotAuthenticatedError` → 401.

Chose `__init__` over overriding `decode_token` because the option belongs to the auth object's configuration, it is visible to anyone inspecting `auth.require_claims`, and it reuses the already-tested `pyjwt` code path rather than adding a second, parallel validation.

### 2. Defense in depth — a single `jti` accessor

Added `token_jti()` on `_BaseBlocklistMixin`; both `check_auth` and `blocklist` now go through it instead of touching `token.jti` directly:

```python
def token_jti(self, token: JWToken) -> str:
    if token.jti is None:
        raise NotAuthenticatedError
    return token.jti
```

Step 1 alone covers the normal decode path, but not a hand-built `JWToken`, a custom `token_cls`, or code that reassigns `require_claims` after construction. Both call sites need the same guard, so it lives in one method. This is what turns the logout 500 into a 401.

In the async `blocklist()` the `token_jti` call is placed **before** `await self.get_user(token)`, so a bad token fails without a pointless DB query.

## Behaviour change to be aware of

Tokens without `jti` used to authenticate successfully under the blocklist mixins; they now get `401`. That is the point of the fix, but it will break anyone who overrode `make_jwt_id()` to return `None` while using the blocklist. Documented in `docs/pages/auth/jwt.rst` under "Blocklisting tokens" and listed in the changelog under Bugfixes.

## AI Policy

## Checklist

## Related issues

Closes #1322


dmr/security/jwt/blocklist/auth.py
```diff
@@ -34,6 +41,24 @@ async def get_user(self, token: JWToken) -> 'AbstractBaseUser': ...
 
 
 class _BaseBlocklistMixin:
+    # Provided by the auth class we are mixed into:
+    require_claims: 'Sequence[str] | None'
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
sobolevn: this is a potential typing-killer. if we are to use this mixin as the first class like we do it:

```python
class MyJWTSyncAuth(JWTokenBlocklistSyncMixin, JWTSyncAuth):
    """JWTSyncAuth with blocklist mixin."""
```

`MyJWTSyncAuth.__init__` would be resolved to be `(*Any, **Any)`, which is not great.

I propose doing a hack here: maybe we should put this method under `if not TYPE_CHECKING` guard? 

Please, add typesafety tests for this mixin usage.

dmr/security/jwt/blocklist/auth.py
```diff
@@ -42,6 +67,26 @@ def blocklist_model(self) -> type['BlocklistedJWToken']:
...
+        :class:`~dmr.security.jwt.token.JWToken` subclass can still
+        reach us without it, and ``None`` would match no blocklist rows
+        at all, while :meth:`blocklist` would fail on a database
+        constraint. Both cases mean the same thing: this token cannot
+        be used with the blocklist.
+
+        Raises:
+            NotAuthenticatedError: If the token has no ``jti``.
+
+        """
```
sobolevn, line 95:
```suggestion
        .. versionadded:: 0.15.0
        """
```

dmr/security/jwt/blocklist/auth.py
```diff
@@ -34,6 +41,24 @@ async def get_user(self, token: JWToken) -> 'AbstractBaseUser': ...
...
+    require_claims: 'Sequence[str] | None'
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """
+        Force the ``jti`` claim to be required.
+
+        The blocklist stores tokens by their ``jti``,
+        so a token without one can never be blocklisted.
+        Accepting such tokens would mean that the blocklist
+        is silently bypassed, we require the claim instead.
```
sobolevn:
```suggestion
        is silently bypassed, we require the claim instead.
        
        .. versionadded:: 0.15.0
```

dmr/security/jwt/blocklist/auth.py
vyhuholl: Fixed and added typesafety tests.

dmr/security/jwt/blocklist/auth.py
```diff
@@ -33,7 +38,34 @@ async def check_auth(
...
+        # See `typesafety/assert_type/test_security/test_jwt/`
+        # for the tests that guard this behavior.
+        def __init__(self, *args, **kwargs) -> None:
+            """
+            Force the ``jti`` claim to be required.
+
+            The blocklist stores tokens by their ``jti``,
+            so a token without one can never be blocklisted.
+            Accepting such tokens would mean that the blocklist
+            is silently bypassed, we require the claim instead.
```
sobolevn, line 61:
```suggestion
            is silently bypassed, we require the claim instead.

            .. versionadded:: 0.15.0
```

sobolevn (approved):
Nice! 👍

PR #1378, opened 2026-09-02
Fix JWT refresh token authentication

### LINKED ISSUE #1320: Refresh JWT tokens authenticate as access JWT tokens
Currently it is possible to auth with JWT access token and JWT refresh token when `HeaderJWTSyncAuth` / `HeaderJWTAsyncAuth` / `CookieJWT*Auth` are used.

We need to change how `decode_token` method works. It must check:

```python
if token.extras.get('type') != self.expected_token_type:
    raise NotAuthenticatedError
```

And define `expected_token_type` attribute on base JWT auth with `'access'` as the default value.

Ryota-Di:
# I have made things!

## AI Policy

## Checklist

- [] I have updated the documentation for the changes I have made

## Related issues

- Closes #1320 

Summary
- refresh tokens could also be used for authentication
- block authentication if the token type does not match expected_token_type

Testing
- just test
- just lint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,8 @@ of requirements for an API to count as public.
 - Fixed `JWToken.decode` validating `exp` and `iat` twice,
   now `leeway`, `verify_exp`, and `verify_iat` are respected
   and invalid tokens return `401` instead of `500`, #1324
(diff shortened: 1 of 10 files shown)


sobolevn (approved):
Awesome! Thank you!

PR #1381, opened 2026-09-02
Show how to use inline-snapshot and dirty-equals

### LINKED ISSUE #1346: Show how to use `inline-snapshot` and `dirty-equals` in tests
We already have both `inline-snapshot` and `dirty-equals` in our test dependencies.

However, we don't really document how and when they should be used in `django-modern-rest` tests.
Let's add several examples to our testing documentation / contributor guidelines.

The important part is to explain the distinction:
* use `inline-snapshot` when the complete value is deterministic and worth asserting
* use `dirty-equals` for dynamic values where we still want to assert their type / shape / properties

kondratevdev:
# I have made things!

## AI Policy

## Checklist

## Related issues

Closes #1346


sobolevn (commented):
Two comments:
1. I don't think that we need https://django-modern-rest--1381.org.readthedocs.build/en/1381/pages/testing/assertions.html#describe-dynamic-values-with-dirty-equals section, we can just show that they work together. Use one assertion with just the `IsUUID()` and one with the `snapshot()` and dirty equals.
2. Move "Testing" after "OpenAPI" in the left menu

docs/examples/testing/combined_assertion_usage.py
```diff
@@ -0,0 +1,26 @@
...
+        data={'email': 'test@example.com', 'age': 43},
+    )
+
+    response = UserController.as_view()(request)
+
+    assert isinstance(response, HttpResponse)
+    assert response.status_code == HTTPStatus.CREATED
+    assert json.loads(response.content) == snapshot({
+        'uid': IsUUID(4),
+        'email': 'test@example.com',
```
sobolevn: let's also show that `snapshot` can assert the random value. generate random email with faker and assert that it is returned.

docs/pages/testing/assertions.rst
```diff
@@ -0,0 +1,144 @@
...
+  :caption: test_combined_assertion.py
+  :language: python
+  :linenos:
+
+This composition gives the test both properties we want: changing the response
+shape or a deterministic field produces a snapshot diff, while a newly
+generated UUID remains valid without making the test flaky.
+
+
+Choosing an assertion
```
sobolevn: I don't think that we need this section.

sobolevn (approved):

PR #1382, opened 2026-09-02
Add `exclude_validate_responses` option

### LINKED ISSUE #1370: `InternalServerError` raised by a handler is returned as `422`

vyhuholl:
# I have made things!

This replaces my earlier approach in #1374, which let every error we handle ourselves through response validation. The review was right: `InternalServerError` is not special there, and `TooManyRequestsError` would have been allowed past the spec in exactly the same way. Nothing implicit is allowed through anymore — you say which status codes you don't want validated.

The new option `exclude_validate_responses` is a set of status codes that are not validated, even when `validate_responses` is on. Default is empty, so nothing changes for anyone who does not set it.

It is configurable on all three levels, exactly like `exclude_semantic_responses`:

```python
DMR_SETTINGS = {
    Settings.exclude_validate_responses: {HTTPStatus.INTERNAL_SERVER_ERROR},
}
```

```python
    exclude_validate_responses = frozenset((HTTPStatus.INTERNAL_SERVER_ERROR,))
```

```python
@modify(exclude_validate_responses={HTTPStatus.INTERNAL_SERVER_ERROR})
def get(self) -> str: ...
```

I reused `_build_optional_set`, so the merge and reset rules are the same as for `exclude_semantic_responses`: levels are merged, and `None` on any level resets everything back to empty. The check itself lives in `ResponseValidator._should_validate_responses`, which now takes a status code. The skip is complete: neither the declared-status check, nor the body, headers or cookies are validated for an excluded code.

The OpenAPI schema is untouched. Excluding a status code documents nothing — that is what `Settings.responses` is for, and the docs say so, so that the two do not get confused.

## Docs
* a warning under `Settings.validate_responses` that with validation enabled in production you have to describe `500`, or the client gets `422 Returned status code 500 is not specified` instead of your error — with the two ways out: describing it via `Settings.responses`, or excluding it from validation;
* a `Settings.exclude_validate_responses` entry of its own, plus a cross-reference from `docs/pages/error-handling.rst`, because that is where people read about `500` in the first place.

## AI Policy

## Checklist

## Related issues

Closes #1370


dmr/validation/response.py
```diff
@@ -113,8 +123,13 @@ def validate_modification(
         )
         return all_response_data
 
-    def _should_validate_responses(self) -> bool:
-        return self.metadata.validate_responses is True
+    def _should_validate_responses(
+        self,
+        status_code: HTTPStatus | int,
+    ) -> bool:
+        if self.metadata.validate_responses is not True:
```
sobolevn:
```suggestion
        if self.metadata.validate_responses:
```

this is a bug :)
`metadata.validate_responses` can't be `None`, annotation is just wrong :)

docs/pages/configuration.rst
```diff
@@ -197,6 +197,73 @@ Response handling
     to :func:`~dmr.endpoint.modify`
     and :func:`~dmr.endpoint.validate`.
 
+  .. warning::
```
sobolevn: this is not a proper place for such docs, let's remove it from the settings description.

tests/test_unit/test_controllers/test_exclude_validate_responses.py
```diff
@@ -0,0 +1,147 @@
...
+    class _PerSettings(Controller[PydanticSerializer]):
+        exclude_validate_responses = frozenset((HTTPStatus.NOT_FOUND,))
+
+        @modify(exclude_validate_responses=None)
+        def get(self) -> str:
+            raise NotImplementedError
+
+    metadata = _PerSettings.api_endpoints['GET'].metadata
+
+    assert metadata.exclude_validate_responses == frozenset()
```
sobolevn, line 138: please, also test the runtime effect, not just the metadata state.

sobolevn (commented):
Please update docs at `docs/pages/using-controller/validation.rst`

sobolevn:
please, fix `EndpointMetadata.validate_responses` annotation to be `bool` :)

docs/pages/configuration.rst
vyhuholl: Moved to `error-handling.rst`

tests/test_unit/test_controllers/test_exclude_validate_responses.py
vyhuholl: Added new tests.

docs/pages/error-handling.rst
```diff
@@ -146,15 +146,65 @@ The same error handling logic can be represented as a diagram:
...
+--------------------------
+
+Error responses are validated like any other response.
+With :data:`~dmr.settings.Settings.validate_responses` enabled,
+only the status codes that you describe are allowed to be returned.
+
+This matters the most for ``500``: we raise
+:exc:`~dmr.exceptions.InternalServerError` ourselves in several places,
+and your own code can raise it as well.
+An undescribed ``500`` will not reach the client, response validation
```
sobolevn, line 161: I would document this differently: 

- Let's add several tabs for several examples
- First examples documents `500` as the response and we describe why and what the effect is: strict guarantee for users. The right way
- Second one does not document `500`, tests that `500` is never sent on normal ocasions and runs with `validate_responses=False` in production
- Just ignores the `500` via this new feature and that's it

Does it sound reasonable?

vyhuholl: I agree. Rewrote the docs.

CHANGELOG.md
sobolevn, line 1: please, add `validate_responses: bool | None` bug fix to the changelog

sobolevn (approved):
🎉

CHANGELOG.md
vyhuholl: Done

PR #1385, opened 2026-09-03
Allow to use stream data in  `parse_as_post` #1328

### LINKED ISSUE #1328: `BytesIO(request.body)` in `parse_as_post` copy the `request.body` object
https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/internal/django.py#L136-L145

This is not really cool, we would copy the request data twice. In some cases this can be really bad, because files can be rather big.

We need to find a way not to copy the data.

max0nT:
# I have made things!

`request.body` calling force to read all data. it causes risk for repeating data read

## AI Policy

## Checklist

## Related issues

- Closes #1328

### DIFF
--- a/dmr/internal/django.py
+++ b/dmr/internal/django.py
@@ -141,6 +141,21 @@ def parse_as_post(request: HttpRequest) -> None:
     """
     # This code is adapted from Django itself:
     if request.content_type == 'multipart/form-data':
+        # There is some cases where request cannot has `_stream`` attribute
+        # 1. App uses custom request class inherited from HttpRequest, which not
+        # initialize `_stream`.
+        # 2. Once the stream has been read
+        # (for example, by a custom middleware that inspects or logs
+        # the request body), it becomes empty and cannot be read again.
+        # Subsequent attempts to read from _stream will return no data,
+        #  making `parse_as_post` fail to retrieve the request body.
+        # 3.Django`s `Client` and `RequestFactory` mock the network socket using
+        # `FakePayload``. Once read, this mock stream is automatically emptied
+        # and cleared by Django. Using `getattr(request, '_stream', request)`
+        # provides a safe fallback if the stream was already consumed or
+        # omitted in the test environment.
+        # `HttpRequest` objects is chosen as fallback argument since it has
+        # methods to read data from request body too.
         request_data = getattr(request, '_stream', request)
         # This was introduced in Django 6.1:
         multipart_parser_cls = getattr(
--- a/dmr/internal/django.py
+++ b/dmr/internal/django.py
@@ -157,6 +157,7 @@ def parse_as_post(request: HttpRequest) -> None:
         # `HttpRequest` objects is chosen as fallback argument since it has
         # methods to read data from request body too.
         request_data = getattr(request, '_stream', request)
+
         # This was introduced in Django 6.1:
         multipart_parser_cls = getattr(
             request,
--- a/dmr/internal/django.py
+++ b/dmr/internal/django.py
@@ -141,21 +141,16 @@ def parse_as_post(request: HttpRequest) -> None:
     """
     # This code is adapted from Django itself:
     if request.content_type == 'multipart/form-data':
-        # There is some cases where request cannot has `_stream`` attribute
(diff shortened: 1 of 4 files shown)


dmr/internal/django.py
```diff
@@ -142,7 +141,7 @@ def parse_as_post(request: HttpRequest) -> None:
     """
     # This code is adapted from Django itself:
     if request.content_type == 'multipart/form-data':
-        request_data = BytesIO(request.body)
+        request_data = getattr(request, '_stream', request)
```
sobolevn, line 154: question: when `_stream` attribute might be missing on `request` object? why do we use `request` as the fallback?

please, explain this in great detail with the comment in the code.

tests/test_unit/test_internal/test_django/test_parse_as_post.py
```diff
@@ -0,0 +1,40 @@
...
+    boundary = 'test_boundary'
+    request = rf.put(
+        '/whatever/',
+        data=encode_multipart(
+            boundary,
+            {
+                'file': SimpleUploadedFile('large.txt', large_content),
+            },
+        ),
+        content_type=f'multipart/form-data; boundary={boundary}',
```
sobolevn: we use `MULTIPART_CONTENT` for this, see other tests

tests/test_unit/test_internal/test_django/test_parse_as_post.py
```diff
@@ -0,0 +1,40 @@
...
+from django.test.client import encode_multipart
+
+from dmr.internal.django import parse_as_post
+
+
+@override_settings(
+    DATA_UPLOAD_MAX_MEMORY_SIZE=1,
+    FILE_UPLOAD_MAX_MEMORY_SIZE=1,
+)
+def test_parse_as_post_not_copy_data(
```
sobolevn, line 14: sorry, this is not clear for me what this test actually does. please, explain where we check that no copy happens exactly :)

sobolevn (commented):
Thank you!

dmr/internal/django.py
max0nT: hm I found three ways where fallback  might be happened
1. App uses custom request class inherited from `HttpRequest`, which not initialize `._stream`
2. `request._stream` might be used once. So if a custom middleware uses that, `parse_as_post` cannot fetch data from stream
3. Django’s `Client` and `RequestFactory` mock the network socket using `FakePayload`. Once read, this mock stream is automatically emptied and cleared by Django. Using `getattr(request, '_stream', request)` provides a safe fallback if the stream was already consumed or omitted in the test environment.

Could it be enough to keep it as is? If yes, I'll detail it in comments

dmr/internal/django.py
```diff
@@ -142,7 +141,23 @@ def parse_as_post(request: HttpRequest) -> None:
...
-        request_data = BytesIO(request.body)
+        # There is some cases where request cannot has `_stream`` attribute
+        # 1. App uses custom request class inherited from HttpRequest, which not
+        # initialize `_stream`.
+        # 2. Once the stream has been read
+        # (for example, by a custom middleware that inspects or logs
+        # the request body), it becomes empty and cannot be read again.
+        # Subsequent attempts to read from _stream will return no data,
+        #  making `parse_as_post` fail to retrieve the request body.
+        # 3.Django`s `Client` and `RequestFactory` mock the network socket using
```
sobolevn:
```suggestion
        # Subsequent attempts to read from `_stream` will return no data,
        # making `parse_as_post` fail to retrieve the request body.
        # 3. Django`s `Client` and `RequestFactory` mock the network socket using
```

dmr/internal/django.py
sobolevn: can you please also rename `_dmr_parsed_as_post` to be `__dmr_parsed_as_post__`?

sobolevn (commented):
Thank you! Can you please also add a statement about memory consumption. What can we promise for our users? Do we ever copy the files? When? Or can we say that it is "zero-copy always"?

sobolevn (approved):
Great work! Thank you!

PR #1404, opened 2026-09-04
Provide faster JWT decode and encode options with msgspec

### LINKED ISSUE #1390: Provide faster JWT decode and encode options with `msgspec`
When `msgspec` is installed, we can speed up the `pyjwt` encode and decode.

By default it uses the default `json` module for both convertions: 
1. https://github.com/jpadilla/pyjwt/blob/7144e4534c34810f4525dc4578a32addd8212cff/jwt/api_jwt.py#L156-L172
2. https://github.com/jpadilla/pyjwt/blob/7144e4534c34810f4525dc4578a32addd8212cff/jwt/api_jwt.py#L287-L301

We can provide our own `PyJWT` subclass and make the process faster. Since JWT is a hot path, even a smaller boost will be a massive win overall.

This will require a benchmark test :)

aryansinha1908:
# I have made things!

`pyjwt` hardcodes the stdlib `json` module for the token payload, but exposes
`_encode_payload` / `_decode_payload` as documented override points. This adds a
`PyJWT` subclass in `dmr/internal/jwt.py` that reuses the JSON backend
`dmr/internal/json.py` already selects, so `msgspec` is used when installed.

Benchmarks (HS256, 8-claim payload): encode `9.9µs` → `6.6µs`, decode
`12.7µs` → `11.7µs`. HMAC and base64 dominate, so this is ~15-25% end to end,
not the 4-8x the payload JSON alone gets. (Tested locally)

Tokens stay byte-identical on both backends. Registered claims always encode the
same way, but `extras` can hold anything: only json-native values there are
guaranteed to be identical with and without `msgspec`. `datetime` with
microseconds, `timedelta`, `set`/`bytes` and `NaN` differ. These are pinned in
tests and documented; closing them would mean changing `_DMREncoder`, which also
moves parser and renderer output, so I left it out. Happy to do it as a follow-up.

One thing worth flagging: our `PyJWT()` builds its own `PyJWS`, unlike the
`pyjwt` module global, so a globally registered custom algorithm would not be
visible. Mirroring that needs a private name across an unpinned `pyjwt>=2.11`,
so it is documented in the class docstring instead.

## AI Policy

- [x] I have read and agree to the [AI
Policy](https://github.com/wemake-services/django-modern-rest/blob/master/.github/AI_POLICY.md),
removed any "Co-Authored-By" lines attributing coding agents, and manually reviewed the final result

## Checklist

- [x] I have double checked that there are no unrelated changes in this pull request (old patches,
accidental config files, etc)

## Related issues

- Closes #1390


dmr/internal/json.py
```diff
@@ -99,14 +99,16 @@ def wrapper(data: Any) -> str:
 try:
     import msgspec
 except ImportError:  # pragma: no cover
+    json_dump_bytes: Callable[[Any], bytes] = NativeJson.dumps
```
sobolevn:
```suggestion
    json_dumps_bytes: Callable[[Any], bytes] = NativeJson.dumps
```

dmr/internal/json.py
```diff
@@ -99,14 +99,16 @@ def wrapper(data: Any) -> str:
...
     _json_dumps: Callable[[Any], str] = _compact_json_dumps
     json_loads: Callable[['str | Raw'], Any] = NativeJson.loads
 else:
+    json_dump_bytes = msgspec.json.encode
     _json_dumps = _wrap_bytes_dumper(msgspec.json.encode)
```
sobolevn:
```suggestion
    _json_dumps = _wrap_bytes_dumper(json_dump_bytes)
```

dmr/internal/jwt.py
```diff
@@ -0,0 +1,50 @@
+import json
+from typing import Any, cast
+
+import jwt
+from typing_extensions import override
+
+from dmr.internal.json import json_dump_bytes, json_loads
+
+
+class _DMRPyJWT(jwt.PyJWT):
```
sobolevn, line 11:
```suggestion
@final
class _DMRPyJWT(jwt.PyJWT):
```

dmr/internal/jwt.py
```diff
@@ -0,0 +1,50 @@
...
+                f'Invalid payload string: {exc}',
+            ) from exc
+        if not isinstance(payload, dict):
+            raise jwt.exceptions.DecodeError(
+                'Invalid payload string: must be a json object',
+            )
+        return cast('dict[str, Any]', payload)
+
+
+dmr_jwt = _DMRPyJWT()
```
sobolevn:
```suggestion
dmr_jwt: Final = _DMRPyJWT()
```

tests/test_unit/test_security/test_jwt/test_jwt_json_backend.py
```diff
@@ -0,0 +1,192 @@
...
+from dmr.internal.json import NativeJson, json_dump_bytes
+from dmr.internal.jwt import dmr_jwt
+from dmr.security.jwt import JWToken
+
+try:
+    import msgspec
+except ImportError:  # pragma: no cover
+    _has_msgspec = False
+else:
+    del msgspec  # noqa: WPS420
```
sobolevn:
```suggestion
```

this is not needed

tests/test_unit/test_security/test_jwt/test_jwt_json_backend.py
```diff
@@ -0,0 +1,192 @@
...
+def test_backends_agree_on_extended_types(claim: Any) -> None:
+    """Ensures both backends encode these extra types identically."""
+    assert NativeJson.dumps({'v': claim}) == json_dump_bytes({'v': claim})
+
+
+@_msgspec_only
+@pytest.mark.parametrize('claim', _DIVERGING_VALUES)
+def test_backends_diverge_on_extended_types(claim: Any) -> None:
+    """Documents that these types are backend dependent in ``extras``."""
+    assert NativeJson.dumps({'v': claim}) != json_dump_bytes({'v': claim})
```
sobolevn: please, add the check: what will values actually be to show the difference.

tests/test_unit/test_security/test_jwt/test_jwt_json_backend.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+@_msgspec_only
+@pytest.mark.parametrize('claim', [{1, 2}, b'ab'])
+def test_native_rejects_msgspec_only_types(claim: Any) -> None:
+    """Documents that ``set`` and ``bytes`` only work with ``msgspec``."""
+    assert json_dump_bytes({'v': claim})
+
+    with pytest.raises(TypeError):
```
sobolevn:
```suggestion
    with pytest.raises(TypeError, match='...'):
```

provide the error message

sobolevn:
Question: what if we provide an exact type for `msgspec` to decode / encode the value into? It should be even faster. We have the type, it is a supported `@dataclass` object. So, let's try it :)

aryansinha1908:
I did try that as well but it was not faster. I measured it on a 8-claim payload and the results were that the JSON step is like 3-4% of a whole JWT call, so 1.05-1.10x there under 0.5% end to end - which is below codspeed's noise floor. So, not worth the complexity.
There are also some blockers for using the msgspec format for this:
1. Format issue: The JWToken.exp/iat are datetime format, and msgspec encodes those in RFC 3339 strings. RFC 7319 requires NumericDate integers. Typed encoding of the dataclass will produce invalid tokens.
2. Unknown claims will be dropped: Typed decode of certain claims such as nbf and custom into struct will lose/delete those claims. Which will break the contract in the token.py file and some existing tests such as test_verify_nbf_can_be_disabled_independently will be broken.
3. Reverse direction fails too: Decoder(Struct).decode on a NumericDate raises ValidationError: Expected datetime, got int — so JWToken cannot be the decode target as declared regardless.

Structurally, PyJWT's _decode_payload must return a dict because decode_complete then runs _validate_claims (exp/nbf/aud/iss/sub/jti) over it.
Returning a struct means bypassing PyJWT's validation pipeline entirely.

If my reasoning was satisfactory then I will apply the suggested changes and add a new commit. Let me know if I need to add any more changes or to use the already defined type regardless.

sobolevn:
Sounds good, let's keep this as-is then :)
Please, fix the merge conflict.

sobolevn (approved):
Great stuff! Thank you! 🎉

sobolevn:
As the next step, I recommend optimizing this part:

```python
{
    field_name: field_value
    for field_name, field_value in asdict(self).items()
    if field_value is not None
}
```

How? 
1. We can move `dataclasses.fields(JWToken)` call to the module level, like: `_JWTokenFieldNames: Final = [field.name for field in dataclasses.fields(JWToken)]`
2. Getting flat simple field values from pre-computed fields must be much faster than calling `asdict`
3. Any other ideas are welcome!

Would you be interested in sending one more PR? :)

aryansinha1908:
Thanks, @sobolevn 
I am happy to create another PR but I want to clrify somethings before creating the PR, if you don't mind answering:
1. Class-keyed lru_cache vs their module-level constant (the subclass evidence above). Blocking — it decides the shape of the change. 
2. Whether decode gets the same helper, or the PR stays encode-only.
3. Whether the extras nested-dataclass change is acceptable as breaking.
4. Issue number for the branch name.

sobolevn:
1. module level is better, it never changes
2. yes, all places can be optimized. `decode` also uses `fields(cls)`
3. we must not change `extra`, why do you want to change it?
4. you can create a new issue or just link this pr :)

PR #1405, opened 2026-09-05
Don't take lock on Redis cache increment

### LINKED ISSUE #1339: `SyncThrottle` and `AsyncThrottle` must not lock `Redis` backend
Currently we always lock all backends in throttling: https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/throttling/base.py#L250-L270

But, it is not really good. Because `redis` backend does not need locks at all.
I propose adding something like `with self._backend.sync_lock(lock)` and `async with self._backend.async_lock(lock)` APIs, so backend can return `null_context` if no lock is needed. Or create custom more complex locks, if they are needed.

Document what backends are lock-free in the backend class docstring.

eoan-ermine:
# I have made things!

## AI Policy

## Checklist

## Related issues

Closes #1339

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,9 @@ of requirements for an API to count as public.
   and `JWToken.decode` about 1.15x faster.
   Note that only json-native values in `JWToken.extras` are guaranteed
   to be encoded identically with and without `msgspec`, #1390
+- Added `BaseThrottleSyncBackend.lock` and `BaseThrottleAsyncBackend.lock`
+  to control the in-process lock for `incr`,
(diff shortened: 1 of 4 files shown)


eoan-ermine:
@sobolevn  My comments may be bad, cause I know only broken English.

Maybe, `lock` should be named `incr_lock`, cause it's only for increment?

Also, I made `lock` return the passed argument default, not `NotImplementedError`, cause I think it's reasonable default behavior. If you think, it'd be better to return `NotImplementedError` as in other methods, then no objections, i'll fix it.

Also I don't know what to test. If you know, I'll be glad for a tip.

sobolevn (approved):
Looks great! Thank you!

PR #1406, opened 2026-09-05
[gh-1332] do not expose reason in ensure_csrf check

### LINKED ISSUE #1332: Do not expose `reason` in `ensure_csrf` check
Currently our responses show why CSRF check fails: 
1. https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/internal/csrf.py#L12
2. https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/internal/csrf.py#L35-L44

It is even tested in some places as:

```python
    assert json.loads(response.content) == snapshot({
        'detail': [
            {
                'msg': 'CSRF Failed: CSRF cookie not set.',
            },
        ],
    })
```

Which is not correct at all! We must only show this during the debug builds, not in production. Just like Django does: 

1. https://github.com/django/django/blob/73cc09f14f13fedddc14d6ba5b287cb33c24e4a4/django/views/templates/csrf_403.html#L38-L46
2. https://github.com/django/django/blob/73cc09f14f13fedddc14d6ba5b287cb33c24e4a4/django/views/csrf.py#L22-L81

AbduazizZiyodov:
# What's been done

On non-debug mode, it will leave `reason` as "Forbidden.".

As I see, middleware overrides `_reject` - so that view will return reason(string) only. If csrf failure succeeds, it will return `None`(in django's implementation).

I've renamed the `_get_csrf_failure_reason` into `_check_csrf_failure` so that it will explicitly says "csrf failed", its more readable too.

No new tests, rather I adapted existing ones - to make them run on debug/non-debug mode, and check csrf failure body accordingly.

To avoid redundancy(same multiple pytest parametrize decorator), I've created new `conftest.py` not sure whether this the best approach.

Saw this one:

https://github.com/wemake-services/django-modern-rest/blob/58ae46f9d1de6e2cf174479aeb1441c36b57bd77/tests/test_integration/conftest.py#L34-L39

I don't think this fully applicable.

One small unrelated change on `test_cookie_auth_try_next_async`, I think tests should not directly use `settings`.

## AI Policy

- [x] I have read and agree to the [AI Policy](https://github.com/wemake-services/django-modern-rest/blob/master/.github/AI_POLICY.md), removed any "Co-Authored-By" lines attributing coding agents, and manually reviewed the final result - **not used any**

## Checklist

- [x] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc) - **explained above**
- [x] I have created at least one test case for the changes I have made - adapted affected/existing ones**

## Related issues

Closes #1332

### DIFF
--- a/dmr/internal/csrf.py
+++ b/dmr/internal/csrf.py
@@ -11,7 +11,7 @@ if TYPE_CHECKING:
     from dmr.serializer import BaseSerializer
 
 _CSRF_FAILED_MSG: Final = _('CSRF Failed: {reason}')
-_NON_DEBUG_CSRF_FAILED_REASON: Final[str] = 'Forbidden.'
+_NON_DEBUG_CSRF_FAILED_REASON: Final = 'Forbidden.'
 
 
 @final
--- a/dmr/internal/csrf.py
+++ b/dmr/internal/csrf.py
@@ -1,6 +1,7 @@
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Final, final
 
+from django.conf import settings
 from django.http import HttpRequest
 from django.middleware.csrf import CsrfViewMiddleware
 from django.utils.translation import gettext_lazy as _
@@ -10,6 +11,7 @@ if TYPE_CHECKING:
     from dmr.serializer import BaseSerializer
 
 _CSRF_FAILED_MSG: Final = _('CSRF Failed: {reason}')
+_NON_DEBUG_CSRF_FAILED_REASON: Final[str] = 'Forbidden.'
 
 
 @final
@@ -22,22 +24,32 @@ class _EnsureCsrfToken(CsrfViewMiddleware):
 
     def _reject(self, request: HttpRequest, reason: str) -> str:
         # Return the failure reason instead of an ``HttpResponse``.
-        return reason
+        # Expose detailed csrf failure reason on DEBUG mode.
+        # Otherwise, provide default placeholder reason.
+
+        if settings.DEBUG:
+            return reason
+
+        return _NON_DEBUG_CSRF_FAILED_REASON
 
 
-def _get_csrf_failure_reason(request: HttpRequest) -> str | None:
+def _check_csrf_failure(request: HttpRequest) -> tuple[bool, str | None]:
     """Perform CSRF validation using ``_EnsureCsrfToken``."""
     check = _EnsureCsrfToken(lambda _: None)  # type: ignore[arg-type]
     check.process_request(request)
-    return check.process_view(request, None, (), {})  # type: ignore[arg-type, return-value]
+    reason = check.process_view(request, None, (), {})  # type: ignore[arg-type]
+    is_failed = reason is not None
+
+    return is_failed, reason  # type: ignore[return-value]
 
 
 def ensure_csrf(controller: 'Controller[BaseSerializer]') -> None:
     """Raise ``APIError`` (403) if the CSRF check fails."""
     from dmr.response import APIError  # noqa: PLC0415
 
-    reason = _get_csrf_failure_reason(controller.request)
-    if reason:
+    is_failed, reason = _check_csrf_failure(controller.request)
+
+    if is_failed:
         raise APIError(
             controller.format_error(_CSRF_FAILED_MSG.format(reason=reason)),
             status_code=HTTPStatus.FORBIDDEN,
--- a/dmr/internal/csrf.py
+++ b/dmr/internal/csrf.py
@@ -11,7 +11,6 @@ if TYPE_CHECKING:
     from dmr.serializer import BaseSerializer
 
 _CSRF_FAILED_MSG: Final = _('CSRF Failed: {reason}')
-_NON_DEBUG_CSRF_FAILED_REASON: Final = 'Forbidden.'
 
 
 @final
@@ -28,29 +27,26 @@ class _EnsureCsrfToken(CsrfViewMiddleware):
         # Otherwise, provide default placeholder reason.
 
         if settings.DEBUG:
-            return reason
+            return _CSRF_FAILED_MSG.format(reason=reason)  # type: ignore[no-any-return]
 
-        return _NON_DEBUG_CSRF_FAILED_REASON
+        return 'CSRF Failed.'
 
 
-def _check_csrf_failure(request: HttpRequest) -> tuple[bool, str | None]:
+def _check_csrf_failure(request: HttpRequest) -> str | None:
     """Perform CSRF validation using ``_EnsureCsrfToken``."""
     check = _EnsureCsrfToken(lambda _: None)  # type: ignore[arg-type]
     check.process_request(request)
-    reason = check.process_view(request, None, (), {})  # type: ignore[arg-type]
-    is_failed = reason is not None
-
-    return is_failed, reason  # type: ignore[return-value]
+    return check.process_view(request, None, (), {})  # type: ignore[arg-type, return-value]
 
 
 def ensure_csrf(controller: 'Controller[BaseSerializer]') -> None:
     """Raise ``APIError`` (403) if the CSRF check fails."""
     from dmr.response import APIError  # noqa: PLC0415
 
-    is_failed, reason = _check_csrf_failure(controller.request)
+    reason = _check_csrf_failure(controller.request)
 
-    if is_failed:
+    if reason is not None:
         raise APIError(
-            controller.format_error(_CSRF_FAILED_MSG.format(reason=reason)),
+            controller.format_error(reason),
             status_code=HTTPStatus.FORBIDDEN,
         )
--- a/docs/pages/integrations.rst
+++ b/docs/pages/integrations.rst
@@ -45,7 +45,8 @@ By default we exempt all controllers from CSRF checks, unless:
    without CSRF is not secure
 
 .. note::
-   ``note`` Detailed CSRF failure reason on response content will be exposed only in debug mode.
+   Detailed CSRF failure reason on response content will be exposed only in debug mode
+   for security reasons.
 
 .. _bring-your-own-di:
 
--- a/docs/pages/integrations.rst
+++ b/docs/pages/integrations.rst
@@ -44,6 +44,8 @@ By default we exempt all controllers from CSRF checks, unless:
    will require CSRF as well. Because using Django sessions
    without CSRF is not secure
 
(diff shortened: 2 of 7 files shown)


dmr/internal/csrf.py
```diff
@@ -10,6 +11,7 @@
     from dmr.serializer import BaseSerializer
 
 _CSRF_FAILED_MSG: Final = _('CSRF Failed: {reason}')
+_NON_DEBUG_CSRF_FAILED_REASON: Final[str] = 'Forbidden.'
```
sobolevn:
```suggestion
_NON_DEBUG_CSRF_FAILED_REASON: Final = 'Forbidden.'
```

docs/pages/integrations.rst
```diff
@@ -44,6 +44,8 @@ By default we exempt all controllers from CSRF checks, unless:
    will require CSRF as well. Because using Django sessions
    without CSRF is not secure
 
+.. note::
+   ``note`` Detailed CSRF failure reason on response content will be exposed only in debug mode.
```
sobolevn:
```suggestion
   Detailed CSRF failure reason on response content will be exposed only in debug mode
   for security reasons.
```

tests/test_unit/test_plugins/test_pydantic/test_error_model_customization.py
```diff
@@ -139,7 +146,7 @@ def test_error_message_controller_customization(
     assert isinstance(response, HttpResponse)
     assert response.status_code == HTTPStatus.FORBIDDEN, response.content
     assert json.loads(response.content) == snapshot({
-        'error': [{'message': 'CSRF Failed: CSRF cookie not set.'}],
+        'error': [{'message': f'CSRF Failed: {expected_csrf_railure_reason}'}],
```
sobolevn:
```suggestion
        'error': [{'message': 'CSRF Failed'}],
```

let's use this value for non-`DEBUG` build, please split the payload check

tests/test_unit/test_security/test_token/test_token_auth/test_cookie_auth.py
```diff
@@ -20,6 +20,7 @@
 )
 from dmr.security.token.app.models import Token
 from dmr.test import DMRAsyncRequestFactory, DMRRequestFactory
+from tests.test_unit.conftest import with_debug_mode_changed_csrf_failure
```
sobolevn: please, do not import from tests. if you need to do seomthing, use a fixture.

you can inject the assertion to call as well. this assertion can know the expected outcome based on the `DEBUG` flag.

dmr/internal/csrf.py
```diff
@@ -22,22 +24,32 @@ class _EnsureCsrfToken(CsrfViewMiddleware):
...
+    return is_failed, reason  # type: ignore[return-value]
 
 
 def ensure_csrf(controller: 'Controller[BaseSerializer]') -> None:
     """Raise ``APIError`` (403) if the CSRF check fails."""
     from dmr.response import APIError  # noqa: PLC0415
 
-    reason = _get_csrf_failure_reason(controller.request)
-    if reason:
+    is_failed, reason = _check_csrf_failure(controller.request)
```
sobolevn: I don't think that we need an extra `is_failed` value. it is just the same as `reason is not None`. let's just use it.

AbduazizZiyodov:
@sobolevn 

I'm trying to understand why there are two kinds of response format's appearing (for same APIError & message):

```python
# first
{
        'error': [
                {'message': 'CSRF Failed ...'}
        ]
}
# second
{
        'detail': [
                {'msg': 'CSRF Failed ...'}
        ]
}
```
I thought, its introduced by my changes but it is not. 

Do you know about this ?

sobolevn:
Error message seems like a bug: it should be `{'details': [{'msg': 'what happened ...'}]}`

AbduazizZiyodov:
> Error message seems like a bug: it should be `{'details': [{'msg': 'what happened ...'}]}`

Ok confirmed, then lemme try to find it - looks like it was missed because of `snapshot(...)` during test writing

AbduazizZiyodov:
@sobolevn 

Behaviour was correct, turns out I've missed `_CustomErrorModel` ... above for `test_error_model_customization.py` (should've known from test file name :smile: )

In `conftest.py` defined `assert_csrf_failure_message`, returns assertion functions by looking into debug mode.

I also had to adapt it for custom error model.

For typing part, I couldn't came up defining custom `_CsrfFailureAssertion` protocol in three test files.

Let me know, if you have better design.

tests/test_unit/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+            'detail': [{'msg': 'CSRF Failed: CSRF cookie not set.'}],
+        })
+
+
+def _assert_on_non_debug(
+    response: HttpResponse,
+    *,
+    uses_custom_error_model: bool = False,
+) -> None:
+    if uses_custom_error_model:
```
sobolevn: I propose just not using this assert for the custom models, just inline the `if settings.DEBUG` with two different asserts in the custom error model tests. Keep this simple.

tests/test_unit/conftest.py
```diff
@@ -0,0 +1,54 @@
+import json
+from collections.abc import Callable
+
+import pytest
+from django.conf import LazySettings
+from django.http import HttpResponse
+from inline_snapshot import snapshot
+
+
+def _assert_on_debug(
```
sobolevn: we don't need these two functions, just add an inner function to `assert_csrf_failure_message`

sobolevn (commented):
Almost ready :)

tests/test_unit/conftest.py
```diff
@@ -0,0 +1,32 @@
+import json
+from collections.abc import Callable
+
+import pytest
+from django.conf import LazySettings
+from django.http import HttpResponse
+from inline_snapshot import snapshot
+
+
+@pytest.fixture(
```
sobolevn, line 13: Please, move the protocol here, make it public and import it in files that need it under `if TYPE_CHECKING:` flag and use it as a string :)

AbduazizZiyodov: Hmm, which protocol ? haven't used `typing.Protocol` on latest commit(previously, yes).

Or you mean this one - `Callable[[HttpResponse], None]` ? Defining it as type alias ...

sobolevn: Yes, sorry :) Type alias :)

tests/test_unit/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+from collections.abc import Callable
+from typing import TYPE_CHECKING, TypeAlias
+
+import pytest
+from django.conf import LazySettings
+from django.http import HttpResponse
+from inline_snapshot import snapshot
+
+if TYPE_CHECKING:
+    CsrfFailureAssertion: TypeAlias = Callable[[HttpResponse], None]
```
sobolevn:
```suggestion
CsrfFailureAssertion: TypeAlias = Callable[[HttpResponse], None]
```

tests/test_unit/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@pytest.fixture(
+    params=[(True), (False)],
+    ids=['debug_on', 'debug_off'],
+)
+def assert_csrf_failure_message(
+    settings: LazySettings,
+    request: pytest.FixtureRequest,
+) -> 'CsrfFailureAssertion':
```
sobolevn:
```suggestion
) -> CsrfFailureAssertion:
```

tests/test_unit/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+from django.conf import LazySettings
+from django.http import HttpResponse
+from inline_snapshot import snapshot
+
+if TYPE_CHECKING:
+    CsrfFailureAssertion: TypeAlias = Callable[[HttpResponse], None]
+
+
+@pytest.fixture(
+    params=[(True), (False)],
```
sobolevn:
```suggestion
    params=[True, False],
```

sobolevn (approved):
Thanks!

sobolevn (changes requested):
Please, address my last comments and add a changelog entry :)
Almost ready!

tests/test_unit/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+def assert_csrf_failure_message(
+    settings: LazySettings,
+    request: pytest.FixtureRequest,
+) -> CsrfFailureAssertion:
+    """Assert CSRF failure message according to debug mode setting."""
+    is_debug_mode_active = request.param
+    settings.DEBUG = is_debug_mode_active
+
+    def _assert(response: HttpResponse) -> None:
+        if is_debug_mode_active:
```
sobolevn:
```suggestion
def assert_csrf_failure_message(
    settings: LazySettings,
) -> CsrfFailureAssertion:
    """Assert CSRF failure message according to debug mode setting."""

    def factory(response: HttpResponse) -> None:
        if settings.DEBUG:
```

AbduazizZiyodov:
> Please, address my last comments and add a changelog entry :) Almost ready!

On changelog, in which order it is kept ? 

I see changes should be made on "0.15.0 WIP", shall I append into list or prepend, or anything like alphabetical order (like cpython)

sobolevn:
just add it to the bottom of "bugfixes" of "0.15.0 WIP" :)

AbduazizZiyodov:
note: re-testing locally again, since we accidentaly applied change that remove `@pytest.fixture` on conftest

AbduazizZiyodov:
@sobolevn 

image

Unit tests are fine but failing on coverage. In my commits, I had parametrized debugging modes and set them before returning that inner function(i.e. factory).

Now those tests don't have the ability to switch between the modes, shall I keep some of my previous changes.

sobolevn:
Yes, you were right, this param was required :) 
I messed it up with integration tests where we do this automatically.
Please, revert my last suggestion

AbduazizZiyodov:
> Yes, you were right, this param was required :) I messed it up with integration tests where we do this automatically. Please, revert my last suggestion

I did [`6ef8a67` (this PR)](https://github.com/wemake-services/django-modern-rest/pull/1406/changes/6ef8a6739deb167ce849010f6a5d260ce29df5c7).

PR #1408, opened 2026-09-06
Build the JWToken payload without asdict

aryansinha1908:
# I have made things!

JWToken.encode built its payload with dataclasses.asdict, which deep-copies the whole instance recursively, and JWToken.decode recomputed dataclasses.fields on every call. Both run on every authenticated
request.

Field names are now precomputed at module level, and encode reads the values flatly with getattr.

Measured on an 8-claim token, HS256, msgspec installed:

┌────────┬────┬────┐
│                               │ before  │    after   │
├────────┼────┼────┤
│ JWToken.encode│ 29.9 us  │ 14.6 us │
├────────┼────┼────┤
│ JWToken.decode│ 17.9 us  │ 16.7 us │
└────────┴────┴────┘

The decode side is a small gain: fields(cls) was only ~0.8 us of it. Most of the win is in encode, where asdict was ~40% of the call.

Tokens themselves do not change, and two details are what keep it that way:

- The names are stored as a tuple, not a set. The claim order in a token follows the field declaration order, and set iteration order depends on the process hash seed, so a set would make the same token
  encode differently between runs. There is a test for the order.
- extras still goes through asdict. It is the only thing that recursively converts nested dataclasses, and dropping it is not a no-op: the pure Python json fallback then raises TypeError while msgspec
  encodes the dataclass on its own, so the two backends would stop agreeing. This is what you meant by "we must not change extra", and it is why encode still pays for a small wrapper instead of a plain
  flat build. Registered claims are flat, so they skip it.

One thing I would like you to confirm: precomputing the names at module level means a subclass of JWToken cannot add new dataclass fields any more, because decode would treat them as unknown claims and
move them into extras. Every subclass in the repo and the docs already uses extras for this, so nothing breaks today, but it does turn "use extras, not new fields" into a rule rather than a convention.
Happy to key the lookup on the class instead if you would rather keep subclassing open.

## AI Policy

## Checklist

## Related issues

Format is:
- Refs #1404 
- Refs #1390


dmr/security/jwt/token.py
```diff
@@ -371,6 +386,30 @@ def _decode_datetime_claim(
             raise NotAuthenticatedError from None
 
 
+@final
+@dataclass(frozen=True, slots=True)
+class _ExtraClaims:
```
sobolevn: I would advocate for the extra simplicity here: we don't document `dataclass` support in `extra`, nor we have any real examples where this is actually the case.

So, let's just treat `extras` as `dict[str, PrimitiveJsonType]`.

No need to support nested dataclasses :)

CHANGELOG.md
```diff
@@ -116,6 +116,13 @@ of requirements for an API to count as public.
   and `JWToken.decode` about 1.15x faster.
   Note that only json-native values in `JWToken.extras` are guaranteed
   to be encoded identically with and without `msgspec`, #1390
+- `JWToken.encode` no longer builds its payload with `dataclasses.asdict`,
```
sobolevn: Please update the changelog with the breaking change: nested dataclasses in `extras` are no longer supported.

sobolevn (commented):
Please, test `JWToken` subclasses as well. They must work with the old way :)

sobolevn:
Can you please also fix the perf suite? Increase the number of times we decode / encode objects. You can do that in a loop 1000 times.

aryansinha1908:
About subclassing JWToken — right now a subclass can only override methods, not add new fields, because the field names are read once at import. Adding a field would make decode treat it as an unknown claim and push it into extras. Nothing in the repo does this today, so I documented it as a rule.
Is that the behaviour you want, or should subclasses be able to add fields?

sobolevn:
No, fields should pushed to extras

aryansinha1908:
One more on subclass fields. Decode pushes them to extras as you said. But encode doesn't write them at all, so a value set on a subclass field is silently lost: set role='staff', encode, decode, and you get the default back with no error. On master this round-tripped, since asdict picked the field up. Three options as I see it — leave it and document it, have encode fold subclass fields into extras so it's symmetric with decode, or raise on encode so it fails loudly instead of silently.
Which do you prefer?

Sorry for so many questions.

sobolevn:
we should keep the current behavior everywhere, just optimize it

CHANGELOG.md
sobolevn, line 1: please, revert all changes except a single new changelog entry

tests/test_unit/test_security/test_jwt/test_jwt_token.py
```diff
@@ -135,6 +135,62 @@ def test_extra_fields(faker: Faker) -> None:
...
+        role: str = 'default'
+
+    token_secret = secrets.token_hex()
+    token = _RoleToken(
+        sub=secrets.token_hex(),
+        exp=dt.datetime.now(dt.UTC) + dt.timedelta(seconds=10),
+        role='staff',
+    )
+
+    decoded = _RoleToken.decode(
```
sobolevn, line 156: please, also test `JWToken.decode()` of encoded `token`
it must also work correctly.

dmr/security/jwt/token.py
```diff
@@ -371,6 +403,30 @@ def _decode_datetime_claim(
...
+    Wraps ``extras`` so we can reuse :func:`~dataclasses.asdict` on it alone.
+
+    ``asdict`` only accepts dataclass instances, but it is the single place
+    that knows how to recursively convert nested dataclasses. We need that
+    conversion for ``extras``, and we don't need it for anything else.
+    """
+
+    extras: dict[str, Any]
+
+
```
sobolevn:
```suggestion
```

it is not needed anymore

dmr/security/jwt/token.py
```diff
@@ -309,6 +308,39 @@ def decode(  # noqa: WPS211
...
+        for field_name in _JWTOKEN_FIELD_NAMES:
+            field_value = getattr(self, field_name)
+            if field_value is not None:
+                payload[field_name] = field_value
+
+        extra_claims = payload.get(_EXTRAS_FIELD)
+        if extra_claims:
+            # `extras` can hold nested dataclasses, which only `asdict`
+            # converts. Registered claims are flat, so they skip it.
+            converted = asdict(_ExtraClaims(extra_claims))
```
sobolevn: As I previously said, we can drop support of dataclasses in `extras`, this was never documented / advertised.
We should better document that `extras` is for simple json-like values only. And add "Breaking changes" entry. And prefer fast code over overly-complex one for no reason.

dmr/security/jwt/token.py
```diff
@@ -371,6 +403,30 @@ def _decode_datetime_claim(
             raise NotAuthenticatedError from None
 
 
+@final
+@dataclass(frozen=True, slots=True)
+class _ExtraClaims:
```
sobolevn: we can remove this class after my requested change :)

aryansinha1908:
Sorry that I wasn't able to add the fixes by myself. I was actually working on them just now. And thank you for providing me with an opportunity to make 2 PRs.

sobolevn:
Thank you for your help and amazing work! I want to release 0.15.0 today :)
Please, feel free to propose more optimizations / ideas.

PR #1418, opened 2026-09-09
Use corrected namespace for postponed annotations

### LINKED ISSUE #1417: `AnnotationsContext` uses the wrapper namespace for postponed endpoint annotations
### What's wrong?

`AnnotationsContext` fails to resolve postponed annotations when an endpoint is decorated with `@endpoint_decorator`.

The decorated function contains annotations copied from the original endpoint, but its `__globals__` belong to `dmr.decorators`. As a result, names imported in the module containing the original endpoint are unavailable during annotation resolution.

Example:
```python
from __future__ import annotations

import inspect
from typing import TypedDict

from django.views.decorators.debug import sensitive_post_parameters

from dmr import Body
from dmr.decorators import endpoint_decorator
from dmr.types import AnnotationsContext


class Payload(TypedDict):
    token: str


class ExampleController:
    @endpoint_decorator(sensitive_post_parameters())
    def post(
        self,
        parsed_body: Body[Payload],
    ) -> dict[str, str]:
        return {}


wrapped = ExampleController.post
original = inspect.unwrap(wrapped)

print(wrapped.__annotations__)
print(wrapped.__globals__.get('Body'))
print(wrapped.__globals__.get('Payload'))

print(1)

print(original.__globals__.get('Body'))
print(original.__globals__.get('Payload'))

annotations = AnnotationsContext()(ExampleController.post)

assert annotations['parsed_body'] == Body[Payload]
assert annotations['return'] == dict[str, str]
```

Output
```bash
1
{'parsed_body': 'Body[Payload]', 'return': 'dict[str, str]'}
None
None
1
{'parsed_body': 'Body[Payload]', 'return': 'dict[str, str]'}
typing.Annotated[~_BodyT, <dmr.components.BodyComponent object at 0x10c89f340>]
<class '__main__.Payload'>
```

The annotation resolution then fails:

```bash
Traceback (most recent call last):
  ...
  File "<string>", line 1, in <module>
NameError: name 'Body' is not defined

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  ...
dmr.exceptions.UnsolvableAnnotationsError:
Annotations of <function ExampleController.post ...> cannot be solved
```

### How it should be?

`AnnotationsContext` should resolve annotations of an endpoint decorated with `@endpoint_decorator`:

```python
{
    'parsed_body': Body[Payload],
    'return': dict[str, str],
}
```

### Used versions

0.14.0

### OS information

No matter

kondratevdev:
# I have made things!

## AI Policy

## Checklist

## Related issues

Closes #1417

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,6 +173,8 @@ of requirements for an API to count as public.
 - JWT authentication now rejects refresh tokens when access tokens are expected,
   #1320
 - Fixed a bug when request data might be copied in `parse_as_post` #1328
+- Fixed postponed annotation resolution for wrapped endpoint functions
+  whose decorators are defined in another module, #1417
 
 ### Misc
 
--- a/dmr/types.py
+++ b/dmr/types.py
@@ -1,3 +1,4 @@
+import inspect
 import sys
 from collections.abc import Callable, Iterator, Mapping
 from typing import (  # noqa: WPS235
@@ -239,7 +240,9 @@ class AnnotationsContext:
         self,
         endpoint_func: Callable[..., Any],
     ) -> dict[str, Any]:
-        return self._globalns or endpoint_func.__globals__
+        if self._globalns is not None:
+            return self._globalns
+        return inspect.unwrap(endpoint_func).__globals__  # type: ignore[no-any-return]
 
 
 class TypeVarInference:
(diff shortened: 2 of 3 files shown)


sobolevn (commented):
Please, add a changelog entry :)

kondratevdev:
> Please, add a changelog entry :)

Done!

sobolevn (approved):

Issue #1341, opened 2026-08-30
[meta] Add new `HttpSpec` checks


sobolevn:
Currently `HttpSpec` only has two checks: https://github.com/wemake-services/django-modern-rest/blob/a2d44b103a3ce8655db9afdbf456f11eb321fdaa/dmr/settings.py#L75-L101

But, we need to add new checks. Things that we can easily do:

 - [x] Check that hop-by-hop headers are not defined anywhere in our schema as response headers: `Connection`, `Keep-Alive`, `Proxy-Authenticate`, `Proxy-Authorization`, `TE`, `Trailer`, `Transfer-Encoding`, `Upgrade`. And server-owned ones: `Content-Length`, `Date`, `Server`. But `HeaderSpec(skip_validation=True)` must still be allowed, like we do in SSE. We can call it `header_name_server_managed`
 - [x] Validate `HeaderSpec` and `NewHeader` names to follow the spec: no `X Custom Name` or `@@@` header names are allowed. We can call it `header_name_syntax`. I am not sure that we can do the same for `parsed_headers` model. Let's discuss this separately
 - [x] Validate `CookieSpec` and `NewCookie` names. Let's start with names as well. And call it `cookie_name_syntax`
 - [ ] Validate cookie semantics. For example: `samesite='none'` requires `secure=True`, `max_age` must not be negative, `__Secure-` prefixed name requires `secure=True`, `__Host-` prefixed name requires `secure=True`, `path='/'` and
  `domain=None`. We can call this `cookie_semantics`

Please, solve each item in its own task. I will create new items if there are something else we can validate.
New ideas are welcome! 

(please, do not take this issue before the 1st of September)

MasoNord:
Hi! I'd like to work on this issue. Is that okay?

sobolevn:
Sure, it would be better to complete one bullet point at a time :) Thanks a lot!

MasoNord:
Thank you! Just for clarification, should I create a separate PR for each bullet point

sobolevn:
yes :)

MasoNord:
Thank you for your reviews! Can I work on the other bullets as well?

sobolevn:
Yes, please!

PR #1420, opened 2026-09-09
Cookie name syntax

MasoNord:
# I have made things!
Added new HTTP spec validation rule for cookie names.

Now, if `cookie_name_syntax` not included in `no_validate_http_spec` any incorrect cookie name will raise `EndpointMetadataError`

## AI Policy

## Checklist

## Related issues
- Refs #1341


sobolevn (approved):
Thank you! Can you please create the next PR that will refactor a common code like

```python
def _check_modification_cookie_syntax(
        self,
    ) -> None:

        modification = self.metadata.modification

        if not modification or not modification.cookies:
            return

        invalid_cookie = self._get_invalid_cookie(
            modification.cookies.keys(),
        )

        if invalid_cookie:
            raise EndpointMetadataError(
                f'Cookie name {invalid_cookie!r} is not following http spec.',
            )
```

into some abtraction / helper?
Otherwise, we would have lots of boilerplate in the core :)

MasoNord:
Sure, I'll make it!

PR #1427, opened 2026-09-10
Add reusable controllers to issue JWT tokens as cookies

### LINKED ISSUE #1290: Reusable controllers to issue JWT tokens as cookies

vyhuholl:
# I have made things!

Six controllers that issue jwt tokens as cookies: `CookieObtainTokensSyncController`, `CookieRefreshTokensSyncController`, `CookieLogoutSyncController`, and the `Async` version of each. This also removes the `.. todo::` from the jwt docs.

`@validate.lazy` from #1409 unblocked this. Cookie names and flags are class-level settings, and `validate_spec` builds the response spec from them for every concrete controller, so a subclass can rename a cookie without redefining `post`. One `CookieSpec` feeds both the OpenAPI schema and the cookie that is actually set, through the new `NewCookie.from_spec` — a response cookie that does not match its spec is a validation error, and I did not want two places to keep in sync.

What I decided on the open questions from the issue:

- Settings are `ClassVar`s next to the existing `jwt_*` ones. `jwt_refresh_cookie_path` has no default: without it the refresh token goes to every endpoint, so a controller that leaves it unset raises `EndpointMetadataError` when its endpoint is built.
- No tokens in the response body. The default response is `204` with no content. When you do want a body, pass its type as the last type argument and change `response_status_code`.
- Logout is transport-only, with a `revoke_tokens()` hook. Blocklisting needs the app installed and auth on the endpoint, so it is a docs example instead of the default.
- Refresh and logout act on cookies alone, so both run the same CSRF check as `CookieJWTSyncAuth` and document the `403`. Obtain does not, because it needs credentials in the body, but it calls `rotate_token()` the way `django.contrib.auth.login` does, so the frontend has a token for everything after it.

Two fixes fell out of writing this, and the feature does not work without them:

- `CookieSpec(max_age=0)` never matched the cookie it describes, `0` was read as "not set". That is how a cookie is dropped, so logout cookies could not be described at all.
- `PydanticFastSerializer` could not validate an empty response body, so any `@validate` endpoint returning `204` failed with it. `MetaMixin` on `OPTIONS` was already broken this way.

Config changes, so they are not a surprise in the diff: `setup.cfg` gets `WPS201, WPS203` for `jwt/views.py` (the module imports a lot now) and `WPS204` for the test app views, `docs/conf.py` gets a `nitpick_ignore` entry for the new type var.

The required `jwt_refresh_cookie_path` is the call I am least sure about, it is an extra line for everyone who uses these controllers. Say the word and I will default it to `'/'` with a warning in the docs instead.

## AI Policy

## Checklist

## Related issues

- Closes #1290


django_test_app/server/apps/jwt_auth/views.py
```diff
@@ -253,3 +260,95 @@ async def get(self) -> _UserOutput:
             self.request.user,
             from_attributes=True,
         )
+
+
+_REFRESH_COOKIE_PATH: Final = '/api/jwt-auth/jwt-cookie-refresh-sync/'
```
sobolevn: this should use `reverse_lazy` or `reverse` with the URL name, not `str`

dmr/security/jwt/views/cookie.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
     async def convert_verify_payload(self, payload: _VerifyTokenT) -> str:
         """Extract the access token string from the request payload."""
         raise NotImplementedError
+
+
+class _BaseCookieTokensController(  # noqa: WPS214
```
sobolevn, line 83: I propose spliting `views.py` into `cookie.py` and `body.py` to indicate where we expect the parameters to be passed.

dmr/security/jwt/views/cookie.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
...
+
+    .. versionadded:: 0.15.0
+    """
+
+    jwt_access_cookie: ClassVar[str] = DEFAULT_ACCESS_COOKIE
+    jwt_refresh_cookie: ClassVar[str] = DEFAULT_REFRESH_COOKIE
+    jwt_access_cookie_path: ClassVar[str] = '/'
+    jwt_refresh_cookie_path: ClassVar[str | None] = None
+    jwt_cookie_domain: ClassVar[str | None] = None
+    jwt_cookie_secure: ClassVar[bool] = True
```
sobolevn, line 124: Question: should we use `not settings.DEBUG` as the default for `secure`? This will require users to use `https` for localhost requests during development, won't it?

dmr/security/jwt/views.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
...
+        """
+        Build the response body that is sent next to the cookies.
+
+        Empty by default: the tokens are already in the cookies,
+        and sending them in the body as well would hand them
+        to any script on the page, undoing ``httponly``.
+
+        Change the response status code from ``204`` when you return a body.
+        """
+        return cast(_CookieResponseT, None)
```
sobolevn: Please, prefer `type: ignore` over casts. `cast` has runtime penatlies.

dmr/security/jwt/views.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
...
```
sobolevn: same here

dmr/security/jwt/views.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
...
+        )
+
+    @sensitive_variables()
+    @endpoint_decorator(sensitive_post_parameters())
+    @validate.lazy(validate_spec)
+    def post(self, parsed_body: Body[_ObtainTokensT]) -> HttpResponse:
+        """By default cookies are acquired on post."""
+        return self.login(parsed_body)
+
+    @sensitive_variables()
```
sobolevn:
```suggestion
```

I don't think that sync call stacks need to duplicate this

dmr/security/jwt/views.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
...
+            *cls.csrf_response_specs(),
+        )
+
+    @sensitive_variables()
+    @validate.lazy(validate_spec)
+    def post(self) -> HttpResponse:
+        """Rotate both cookies on post."""
+        return self.refresh()
+
+    @sensitive_variables()
```
sobolevn:
```suggestion
```

dmr/security/jwt/views.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
...
+        from django.contrib.auth import get_user_model  # noqa: PLC0415
+
+        try:
+            return get_user_model().objects.get(**{
+                self.jwt_user_id_field: token.sub,
+            })
+        except USER_LOOKUP_ERRORS:
+            raise NotAuthenticatedError from None
+
+    def check_auth(self, user: Any) -> None:
```
sobolevn: this is not correct. `user` must be `AbstractBaseUser`, it should also accept `token` as the second parameter.

we have the same mistake in other views :( 

why is this a problem? because `JWTokenBlocklistSyncMixin` and `JWTokenBlocklistAsyncMixin` objects won't work with the view. but, we need to support this usecase.

currently, we missed this whole feature :( 
please, test that these two mixins can blocklist token refreshing. when the access token is blocklisted.

dmr/security/jwt/views/cookie.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
...
+            ResponseSpec(
+                _COOKIE_RESPONSE_TYPE,
+                status_code=cls.response_status_code,
+                headers=_NO_STORE_SPEC,
+                cookies=cls.discarded_cookies_spec(),
+            ),
+            *cls.csrf_response_specs(),
+        )
+
+    @validate.lazy(validate_spec)
```
sobolevn, line 750: I think that it still worth adding `@sensitive_variables` here

dmr/security/jwt/views/cookie.py
```diff
@@ -698,3 +737,730 @@ async def check_auth(self, user: Any) -> None:
...
+        await self.check_auth(user)
+        await self.set_request_attrs(self.request, user)
+        return self.to_response(
+            await self.make_api_response(),
+            status_code=self.response_status_code,
+            headers=_NO_STORE_VALUES,
+            cookies=self.issue_cookies(),
+        )
+
+    async def get_user(self, token: JWToken) -> AbstractBaseUser:
```
sobolevn, line 694: let's annotate all async function that actually use `token` as sensitive

docs/pages/auth/jwt.rst
```diff
@@ -296,37 +296,150 @@ On any validation failure it returns ``401 Unauthorized``.
...
+  when your frontend really is on another site.
+
+There is no response body by default: the tokens are already in the cookies,
+and sending them in the body as well would hand them
+to any script on the page. When you do need a body,
+pass its type as the last type argument and change the status code:
+
+.. code:: python
+
+  class ObtainCookiesController(
```
sobolevn: please, never do this in the docs. code can either be:
- really simple, like `user.logout()`
- a doctest for something that is more than 2 lines
- a separate file for something that is more than 5-8 lines. of if the typing of the example is important for us

otherwise we would have broken code in the docs.

docs/pages/using-controller/headers-and-cookies.rst
```diff
@@ -96,6 +96,21 @@ it is required to describe the set cookies with
...
+.. tip::
+
+  A response cookie must match its spec exactly, so keeping the flags
+  in two places is a source of bugs. Build the cookie from its spec with
+  :meth:`NewCookie.from_spec <dmr.cookies.NewCookie.from_spec>`
+  when only the value is known in runtime:
+
+  .. code:: python
+
+    >>> from dmr import CookieSpec, NewCookie
```
sobolevn, line 108: This is a good example of a doctest 👍

sobolevn (commented):
Wow, this is very cool! Thanks a lot for the help.

Additional notes:
- Please, test at least one controller with `validate_spec` redefined

sobolevn:
I am fixing the CI right now :)

sobolevn (approved):
Great stuff! Now we can release 0.15.0 🎉

PR #1428, opened 2026-09-10
Cursor pagination

s3rius:
Signed-off-by: Pavel Kirilin <s3riussan@gmail.com>

# I have made things!

## AI Policy

## Checklist


s3rius:
@sobolevn, I have a question. 

if `PaginatedCursor` is a response model, we can assume that `CursorPaginator` was used to generate the next page. So we need to specify all possible error statuses in the OpenAPI schema. 

But I couldn't find a way to modify the response schema from a returned DTO. Is there a way to do so? 

Once this issue is resolved, I can remove `positive_data_acceptance` from schema thesis.

django_test_app/server/apps/model_cursor/views.py
```diff
@@ -0,0 +1,174 @@
...
+            ResponseSpec(
+                Controller.error_model,
+                status_code=HTTPStatus.BAD_REQUEST,
+            ),
+        ],
+    )
+    def get(
+        self,
+        parsed_query: Query[CursorQuery],
+    ) -> PaginatedCursor[EntryModel]:
```
sobolevn: I think that the main design problem here is that we try to mix two things in this API.

Look at how Django's paginator works in our examples:
- It returns some internal paginator types
- We wrap them in our own dataclasses that we controll

Here we can do the same thing. `.next_page` should return some general `Page` class that can't and must not be serialized to the response at all.

And we must serialize it manually to `PaginatedCursor` type.

This way we can controll the return type. We can create our own model and change the representation of the pagination.

sobolevn (commented):
Thank you! This is not a full review, just a major comment about the design :) 

We should try to keep `Paginator` as the source of truth. So, people can switch easily. Designs must be very similar :)

This would also need to be tested with `mysql` / `postgres` / `sqlite` databases, see https://github.com/wemake-services/django-test-migrations and how we test `redis`. Basically, you can just add different databases to different python versions. So, we won't have many matrix jobs and still test all the possible dbs.

dmr/pagination/cursor.py
sobolevn, line 1: I propose making `pagination` a package and using `pagination/cursor.py` and `pagination/base.py` as names with re-exports of public parts to `pagination/__init__.py`

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,287 @@
...
+
+from dmr.exceptions import InvalidCursorError
+
+_ModelT = TypeVar('_ModelT')
+_DjangoModel = TypeVar('_DjangoModel', bound=models.Model)
+
+
+@dataclasses.dataclass(slots=True, frozen=True, kw_only=True)
+class Paginated(Generic[_ModelT]):
+    """Helper for returning cursor paginated response."""
```
sobolevn: Please, document attributes with `Attributes:` sphinx notation

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,287 @@
...
+    ) -> None:
+        """Ensure each decoded cursor value fits its model field type."""
+        for index, field_name in enumerate(self._order_by_field_names):
+            field = cast(
+                'models.Field[Any, Any]',
+                self.queryset.model._meta.get_field(field_name),  # noqa: SLF001
+            )
+            try:
+                field.get_prep_value(decoded_cursor[index])
+            except (
```
sobolevn: I propose using `except Exception` here, because we can deal with custom fields that raise custom exceptions. We don't really care about them.

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,287 @@
...
+            )
+        if per_page < 1:
+            raise ValueError('`per_page` should be a positive integer!')
+
+        self.order_by = order_by
+        self.queryset = queryset
+        self.per_page = per_page
+        self.cursor_encoder = cursor_encoder or _default_cursor_encoder
+        self.cursor_decoder = cursor_decoder or _default_cursor_decoder
+        self.separator = separator
```
sobolevn: I think that these three attributes can be protected, can't they?

dmr/pagination/cursor.py
s3rius: Yeah, that make sense.

dmr/pagination/cursor.py
s3rius: Fair.

dmr/pagination/cursor.py
s3rius: Yes, they can. I don't think that they are going to be used by people directly. It's mostly for encoding only.

dmr/exceptions.py
```diff
@@ -83,6 +83,26 @@ class RequestSerializationError(Exception):
     status_code: ClassVar[HTTPStatus] = HTTPStatus.BAD_REQUEST
 
 
+@final
+class InvalidCursorError(Exception):
```
sobolevn: this should live near the cursor, here we only store the major exceptions, subdomains can have their own ones.

tests/test_integration/test_cursor_pagination/test_cursor_pagination.py
```diff
@@ -0,0 +1,165 @@
...
+        field = order.removeprefix('-')
+        ordered = sorted(
+            ordered,
+            key=operator.attrgetter(field),
+            reverse=descending,
+        )
+    return [entry.rank for entry in ordered]
+
+
+def walk_pagination(
```
sobolevn: all functions in tests are protected by default

tests/test_integration/test_cursor_pagination/test_cursor_pagination.py
```diff
@@ -0,0 +1,165 @@
...
+    assert walked == _expected_ranks(entries, order_by)
+    for page in pages:
+        assert page['per_page'] == limit
+        assert len(page['page']) <= limit
+    for page in pages[:-1]:
+        assert len(page['page']) == limit
+    assert pages[-1]['has_next'] is False
+
+
+_ORDER_BY_PARAMS = (
```
sobolevn:
```suggestion
_ORDER_BY_PARAMS: Final = (
```

tests/test_integration/test_cursor_pagination/test_cursor_pagination.py
```diff
@@ -0,0 +1,165 @@
...
+    pytest.param(['rank', 'name'], id='rank+name'),
+    pytest.param(['rank'], id='rank'),
+    pytest.param(['-rank'], id='-rank'),
+    pytest.param(['-rank', 'name'], id='-rank+name'),
+    pytest.param(['name'], id='name'),
+    pytest.param(['name', 'rank'], id='name+rank'),
+    pytest.param(['created_at'], id='created_at'),
+)
+_LIMIT_PARAMS: tuple[int, ...] = (1, 2, 3, 5, 10)
```
sobolevn:
```suggestion
_LIMIT_PARAMS: Final = (1, 2, 3, 5, 10)
```

tests/test_integration/test_openapi/test_schema.py
```diff
@@ -64,7 +64,30 @@ def _modify_integration_settings(settings: LazySettings) -> None:
 @pytest.fixture
 def api_schema(transactional_db: None, admin_user: User) -> OpenApiSchema:
     """Load OpenAPI schema as a pytest fixture."""
-    return st.openapi.from_wsgi(reverse('openapi_json'), application)
+    # The cursor-paginated endpoints expose a free-form ``cursor`` string
```
sobolevn: can we move this to the config file? and only apply to paginated urls?

or! even better: can we write a provider for the cursor strings? for example with a custom `format` on a string field.

tests/test_unit/test_cursor_pagination/test_cursor_paginator.py
```diff
@@ -0,0 +1,128 @@
+"""Unit tests for the cursor paginator."""
```
sobolevn:
```suggestion
```

we don't use module-level docstrings

tests/test_unit/test_cursor_pagination/test_cursor_paginator.py
```diff
@@ -0,0 +1,128 @@
...
+
+import pytest
+
+from dmr.cursor_pagination import CursorPaginator
+from dmr.exceptions import InvalidCursorError
+from server.apps.model_cursor.models import (  # type: ignore[import-not-found]
+    Entry,
+)
+
+_SEPARATOR = '|-|'
```
sobolevn:
```suggestion
_SEPARATOR: Final = '|-|'
```

sobolevn: Or you can import protected names in tests

sobolevn (commented):
Are you sure that we can't get ordering from the queryset itself? 
Like `self.queryset.query.order_by`? This way writting

```python
paginator = CursorPaginator(
    Entry.objects.all().order_by('created_at', 'rank'),
    per_page=10,
)
```

would be type-safe, instead of:

```python
paginator = CursorPaginator(
    Entry.objects.all(),
    per_page=10,
    order_by=['missing_field'],
)
```

which is not.
because `django-stubs` can validate field names in `.order_by` calls.

tests/test_unit/test_cursor_pagination/test_cursor_paginator.py
s3rius: Yes. Rewritten the `_encode` function. Now it just uses paginator's protected attributes.

django_test_app/server/apps/model_cursor/views.py
s3rius: Refactored a bit.

dmr/pagination/cursor.py
s3rius: Just did it.

dmr/pagination/cursor.py
s3rius: Done.

tests/test_integration/test_cursor_pagination/test_cursor_pagination.py
s3rius: Fixed all functions which are not tests.

tests/test_integration/test_openapi/test_schema.py
s3rius: I'll move it to config. 

I'm not really sure how we can define it schematically. 

Since, cursor can have its own decode and encode function, let's say for encrypting cursors if you don't want users to be able to access inner attributes of objects.

dmr/exceptions.py
s3rius: Ah, okay.

django_test_app/server/apps/model_cursor/views.py
```diff
@@ -0,0 +1,101 @@
...
+    """List entries using cursor-based pagination."""
+
+    @modify(
+        extra_responses=[
+            ResponseSpec(
+                Controller.error_model,
+                status_code=HTTPStatus.BAD_REQUEST,
+            ),
+        ],
+    )
```
sobolevn:
```suggestion
```

this should not be needed, we inject it automatically for any component parsers, see `dmr/components.py:205`

django_test_app/server/apps/model_cursor/views.py
```diff
@@ -0,0 +1,101 @@
...
+    """List entries using cursor-based pagination (async)."""
+
+    @modify(
+        extra_responses=[
+            ResponseSpec(
+                Controller.error_model,
+                status_code=HTTPStatus.BAD_REQUEST,
+            ),
+        ],
+    )
```
sobolevn:
```suggestion
```

same here

dmr/pagination/__init__.py
```diff
@@ -0,0 +1,6 @@
+from dmr.pagination.base import Page as Page
+from dmr.pagination.base import Paginated as Paginated
+from dmr.pagination.cursor import CursorPage as CursorPage
```
sobolevn:
```suggestion
```

we try to keep the re-exported API as minimal as possible

dmr/pagination/__init__.py
```diff
@@ -0,0 +1,6 @@
...
+from dmr.pagination.cursor import CursorPaginated as CursorPaginated
+from dmr.pagination.cursor import CursorPaginator as CursorPaginator
+from dmr.pagination.cursor import InvalidCursorError as InvalidCursorError
```
sobolevn:
```suggestion
```
if user will want to use all the feature, they would just use `dmr.pagination.cursor` and import all the things.

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,299 @@
...
+@final
+class InvalidCursorError(Exception):
+    """
+    Raised when a pagination cursor cannot be decoded or validated.
+
+    The cursor is a client-provided value,
+    so a malformed one is a client error.
+    """
+
+    default_message: ClassVar[str | Promise] = _('Invalid cursor')
```
sobolevn, line 26: please, update all the translations :) 
see `just makemessages`

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,299 @@
...
+@dataclasses.dataclass(slots=True, frozen=True, kw_only=True)
+class CursorPaginated(Generic[_ModelT]):
+    """
+    Helper for returning cursor paginated response.
+
+    Attributes:
+        next_cursor: Encoded cursor to request the next page.
+            ``None`` when there are no more pages.
+        per_page: Number of objects per one page.
+        has_next: Whether a next page exists.
```
sobolevn: question: why do we need `has_next` if `next_cursor` will be `None` where there are no more pages?

this allows creating invalid state invariant: `CursorPaginated(next_cursor=None, has_next=True)`

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,299 @@
...
+        separator: str = '|-|',
+    ) -> None:
+        """
+        Create a paginator over ``queryset``.
+
+        ``order_by`` is required: it defines the cursor ordering,
+        and the primary key fields are appended automatically
+        as tiebreakers to keep it stable.
+        """
+        if per_page < 1:
```
sobolevn:
```suggestion
        if per_page <= 0:
```

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,299 @@
...
+    def encode_cursor(self, cursor_attrs: dict[str, Any]) -> str:
+        """Build an encoded cursor from the ordered fields of *obj*."""
+        return self._cursor_encoder(
+            self._separator.join([
+                str(cursor_attrs[attr_name])
+                for attr_name in self._order_by_field_names
+            ]),
+        )
+
+    def apply_cursor_filter(  # noqa: WPS210
```
sobolevn:
```suggestion
    def apply_cursor_filter(
```

let's split this method into several ones, so we can remove the complexity rule ignore :)

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,299 @@
...
+            pk.name
+            for pk in self._queryset.model._meta.pk_fields  # noqa: SLF001
+            if pk.name not in field_names
+        ]
+
+        self._order_by_fields = [*orderby_fields, *pks]
+        self._order_by_field_names = [
+            field.removeprefix('-') for field in self._order_by_fields
+        ]
+        self._queryset = self._queryset.order_by(*self._order_by_fields)
```
sobolevn, line 266: I am not sure that this line is required. the queryset is already ordered. moreover, it can be already evaluated. so, re-ordering it will trigger one more db query 🤔

sobolevn: oh, I see. you also order it by all the pks. in this case we must assert that the queryset is still lazy and unevaluated. and document this.

dmr/pagination/cursor.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        Return the page that starts after ``cursor``.
+
+        If ``encoder`` is given, ``items`` will hold its results
+        instead of the raw model instances.
+        """
+        if cursor:
+            self.apply_cursor_filter(cursor)
+
+        objects_list = list(self._queryset[: self.per_page + 1])
```
sobolevn: please, add a comment here that we only load one extra item to see if there are next objects :)

dmr/pagination/__init__.py
sobolevn, line 3:
```suggestion
```

dmr/errors.py
```diff
@@ -26,6 +26,7 @@
     TooManyRequestsError,
     ValidationError,
 )
+from dmr.pagination import InvalidCursorError
```
sobolevn: I vote strongly against using this error in the default error handling. The proper way is to create a `mappers.py` file (we already have an example).

since this error can't happen in the framework code itself, only in user code that uses the `CursorPaginator`, it is a user-facing, logical error. there might be different ways to handle it.

options:
- handle it in business logic with `mappers.py`
- handle it in `handle_error` method / endpoint function

please, adjust tests :)

docs/examples/integrations/cursor_pagination.py
```diff
@@ -0,0 +1,48 @@
...
+    cursor: NotRequired[str]
+    page_size: NotRequired[int]
+
+
+class _Entry(pydantic.BaseModel):
+    rank: int
+    name: str
+
+
+class EntryController(Controller[PydanticSerializer]):
```
sobolevn:
```suggestion
class EntryController(Controller[PydanticFastSerializer]):
```

docs/examples/integrations/cursor_pagination.py
```diff
@@ -0,0 +1,48 @@
...
+    Entry,
+)
+
+
+class _PageQuery(TypedDict):
+    cursor: NotRequired[str]
+    page_size: NotRequired[int]
+
+
+class _Entry(pydantic.BaseModel):
```
sobolevn:
```suggestion
class _EntrySchema(pydantic.BaseModel):
```

docs/examples/integrations/cursor_pagination.py
sobolevn, line 1: I would say that the better way would be:
1. split `django_test_app` test `views` into sync and async files
2. add `# run` comments there
3. include them directly, like we do with `model_fk` app

This way there would be less duplication, no `# type: ignore` comments, single source of truth

docs/pages/pagination.rst
```diff
@@ -0,0 +1,235 @@
+.. _pagination:
+
```
sobolevn:
```suggestion
```

unused ref

docs/pages/pagination.rst
```diff
@@ -0,0 +1,235 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can pass your own ``cursor_encoder`` and ``cursor_decoder``
+functions, as well as a custom ``separator``:
+
+.. code-block:: python
+  :caption: views.py
+  :linenos:
+
+  import base64
```
sobolevn: please, only use doctests in docs

docs/pages/pagination.rst
```diff
@@ -0,0 +1,235 @@
...
+  )
+
+.. note::
+
+  The default ``base64`` encoding may contain ``+`` and ``/`` characters.
+  They are valid in query strings only when percent-encoded by the client.
+  ``base64.urlsafe_b64encode`` avoids this and is usually a better default
+  for cursors sent over HTTP.
+
+Errors
```
sobolevn, line 204: This section would need to be changed to feature the new error handling mechanism

docs/pages/pagination.rst
```diff
@@ -0,0 +1,235 @@
...
+.. autoclass:: dmr.pagination.CursorPage
+  :members:
+
+.. autoclass:: dmr.pagination.CursorPaginated
+  :members:
+
+.. autoclass:: dmr.pagination.InvalidCursorError
+  :members:
+
+Limit offset
```
sobolevn: This should be the first to follow the docs logic.

tests/test_integration/test_cursor_pagination/test_cursor_pagination.py
```diff
@@ -0,0 +1,163 @@
...
+
+import pytest
+from django.urls import reverse
+
+from dmr.test import DMRAsyncClient, DMRClient
+from server.apps.model_cursor.models import (  # type: ignore[import-not-found]
+    Entry,
+)
+
+_SYNC_VIEW = 'api:model_cursor:entries'
```
sobolevn:
```suggestion
_SYNC_VIEW: Final = 'api:model_cursor:entries'
```

tests/test_integration/test_cursor_pagination/test_cursor_pagination.py
```diff
@@ -0,0 +1,163 @@
...
+_ASYNC_VIEW = 'api:model_cursor:entries_async'
```
sobolevn:
```suggestion
_ASYNC_VIEW: Final = 'api:model_cursor:entries_async'
```

tests/test_unit/test_cursor_pagination/test_cursor_paginator.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+
+from dmr.pagination import CursorPaginator, InvalidCursorError
+from server.apps.model_cursor.models import (  # type: ignore[import-not-found]
+    Entry,
+)
+
+
+def _encode(paginator: 'CursorPaginator[Entry]', *to_encode: str) -> str:
```
sobolevn:
```suggestion
def _encode(paginator: CursorPaginator[Entry], *to_encode: str) -> str:
```

tests/test_integration/test_cursor_pagination/test_cursor_paginator.py
```diff
@@ -0,0 +1,75 @@
...
+        InvalidCursorError,
+        match='Cursor has 2 fields, but 3 were expected',
+    ):
+        paginator.page(_encode(paginator, '1', 'c'))
+
+
+def test_empty_result_page() -> None:
+    """An empty queryset yields an empty page without a next cursor."""
+    paginator = CursorPaginator(
+        Entry.objects.none().order_by('rank'),
```
sobolevn, line 58: If you are using `Entry` and database access, this is not a unit test

sobolevn (commented):
Wow! This is an amazing work! Thank you! 

I have several comments :)

tests/test_integration/test_cursor_pagination/test_cursor_paginator.py
s3rius: It uses `objects.none()`, which effectively doesn't do any database calls. And the API itself requires a queryset.

dmr/pagination/cursor.py
s3rius: Oh, damn, true. This one a bit easier to read.

dmr/pagination/cursor.py
s3rius: There is a case, actually. 

Imagine you're paginating over a list while entities are being added frequently. You iterate through all the objects and reach the end of the list.

Here you may have a last page where `0 < len(items) < limit`. 

In this situation, we could give you a cursor that will get new data when it arrives, however, at the time of fetching we can say for sure that `has_next: false`.

On the other hand, we can safely remove `has_next`, because generally on frontend they can calculate this themselves.

dmr/pagination/__init__.py
s3rius: Okay, fair.

django_test_app/server/apps/model_cursor/views.py
```diff
@@ -0,0 +1,84 @@
...
+
+import pydantic
+from annotated_types import Gt, Len, Lt
+
+from dmr import Controller, Query
+from dmr.pagination import CursorPaginated, CursorPaginator
+from dmr.plugins.pydantic.serializer import PydanticSerializer
+from server.apps.model_cursor.models import Entry
+
+ALLOWED_ORDER_FIELDS: TypeAlias = Literal[
```
sobolevn:
```suggestion
_AllowedOrdering: TypeAlias = Literal[
```

type aliases are written in PascalCase

tests/test_integration/test_cursor_pagination/test_cursor_paginator.py
sobolevn: Yes, but I would still suggest moving this to integration tests, not to cause confusion in the future.

dmr/pagination/cursor.py
s3rius: Kk.

dmr/pagination/cursor.py
s3rius: Yes. This implementation of pagination allows you to order by non-unique fields and still get correct pagination without rowskipping.

In order for this thing to work, I have to make sure that PK is included in the order_by to make sorting consistent. 

Where do you want me to put this note?

django_test_app/server/apps/model_cursor/views.py
s3rius: Really? Didn't know that. Thanks!

dmr/pagination/cursor.py
sobolevn: into the main docs :)

docs/pages/pagination.rst
s3rius: It is actually used. Old page about integrations, where pagination previously were described, leads here using this label.

docs/pages/pagination.rst
s3rius: I'm not sure if it's possible here. What shall we do with queryset restriction? Like, in this docstring I use queryset, there's no way I can get it in this docstring. Or can I?

s3rius: I can use same `Entity` model I was using for views example. Is it okay?

tests/test_integration/test_cursor_pagination/test_cursor_paginator.py
s3rius: Okay.

dmr/errors.py
s3rius: Just removed it.

dmr/pagination/cursor.py
s3rius: We won't with new approach.

dmr/pagination/cursor.py
s3rius: We decided to remove `has_next` due to its redundancy.

dmr/pagination/cursor.py
s3rius: Added an explanation section. Since it's a bit complex thing.

docs/pages/pagination.rst
s3rius: Okay, I rewrote this snippet to be a full-fledged standalone example with a controller. Now you can even see encrypted cursor.

docs/pages/pagination.rst
s3rius: Can you tell what exactly needs to be updated according to new error handling mechanism? Or is it just a note for upcoming changes?

django_test_app/server/apps/model_cursor/views.py
```diff
@@ -0,0 +1,37 @@
+from typing import final
+
+from django_test_app.server.apps.model_cursor.schema import (  # type: ignore[import-untyped]
```
sobolevn:
```suggestion
from server.apps.model_cursor.schema import (
```

this will solve the typing issue :)

s3rius: Yo, for real. Thanks.

s3rius: image

docs/pages/pagination.rst
```diff
@@ -0,0 +1,286 @@
...
+  The paginator requires a stable ordering.
+  Set ``.order_by()`` on your queryset
+  before creating the paginator.
+  The primary key fields are appended automatically as tiebreakers,
+  so your ``order_by`` fields do not have to be unique.
+
+Model
+~~~~~
+
+.. literalinclude:: ../../django_test_app/server/apps/model_cursor/models.py
```
sobolevn, line 61:
```suggestion
It works with any model:

.. literalinclude:: ../../django_test_app/server/apps/model_cursor/models.py
```

docs/pages/pagination.rst
```diff
@@ -0,0 +1,286 @@
...
+   :class:`~dmr.pagination.CursorPaginated`
+
+Client protocol
+~~~~~~~~~~~~~~~
+
+1. The first request is made without the ``cursor`` parameter
+2. The response contains ``next_cursor``
+3. To fetch the next page,
+   pass ``next_cursor`` back as the ``cursor`` parameter
+4. Repeat until ``next_cursor`` becomes ``null``
```
sobolevn:
```suggestion
4. Repeat until ``next_cursor`` becomes ``None``
```

docs/pages/pagination.rst
```diff
@@ -0,0 +1,286 @@
...
+  :caption: models.py
+  :language: python
+  :linenos:
+
+Views
+~~~~~
+
+Here is an example of a list endpoint with cursor pagination:
+
+.. literalinclude:: /examples/integrations/cursor_pagination.py
```
sobolevn: Please, use `.. tabs` and include both `async_views.py` and `views.py` from the demo app. They already have `# run` comments.

docs/pages/pagination.rst
```diff
@@ -0,0 +1,286 @@
...
+or when it contains a different number of fields than expected,
+the paginator raises :exc:`~dmr.pagination.cursor.InvalidCursorError`.
+
+Unlike other framework errors,
+it is not handled by the default
+:func:`~dmr.errors.global_error_handler` on purpose.
+You should catch it in your view
+and convert it to an error response yourself:
+
+.. code-block:: python
```
sobolevn: We have a hard rule, that docs can't contain python code that is not executed via doctests or `.. literalinclude`

I think that this one must be in a separate example file.
Run it and show the openapi schema.

docs/pages/pagination.rst
```diff
@@ -0,0 +1,286 @@
...
+in :meth:`~dmr.controller.Controller.handle_error`
+or in a custom global error handler,
+see :doc:`error-handling` for details.
+
+If your endpoint can return this error,
+remember to document the ``400`` status code
+in its :class:`~dmr.metadata.ResponseSpec` definitions.
+
+Other libraries
+~~~~~~~~~~~~~~~
```
sobolevn, line 229:
```suggestion


Other libraries
---------------
```

setup.cfg
```diff
@@ -54,6 +55,8 @@ per-file-ignores =
   dmr/throttling/*.py: WPS226
   # It is fine to have many exceptions:
   dmr/exceptions.py: WPS202
+  # The error hub imports every error type:
+  dmr/errors.py: WPS235
```
```suggestion
```

it is not touched anymore, so it can be removed.

schemathesis.toml
```diff
@@ -18,3 +18,16 @@ extract_selector = '/access_token'
 [output.sanitization]
 # We don't have any real data to hide, but it is easier to debug:
 enabled = false
+
+# Cursor-paginated endpoints answer `400`/`422` when query validation rejects
```
sobolevn: Please, replace with proper error handling.

setup.cfg
s3rius: Facts.

docs/pages/pagination.rst
s3rius: Kk. Removed all code blocks with python code.

schemathesis.toml
s3rius: I made it a bit differently. Instead, I decided to actually cause error and show it in both sync and async views. 

Since they have error handler, it'd show what happens with invalid cursors. In that case we just need to mention that this error exists and we're good.

docs/examples/integrations/cursor_pagination.py
sobolevn, line 1: this file looks unused, please remove it

docs/pages/pagination.rst
sobolevn: nope :) it uses `doc` which links by the filename, not `ref` which needs this label.

docs/pages/pagination.rst
```diff
@@ -0,0 +1,261 @@
...
+
+.. literalinclude:: ../../django_test_app/server/apps/model_cursor/models.py
+  :caption: models.py
+  :language: python
+  :linenos:
+
+Views
+~~~~~
+
+Here is an example of a list endpoint with cursor pagination:
```
sobolevn, line 69:
```suggestion
Here is an example of a list endpoint with cursor pagination:

.. important::

   We can only paginate stably ordered querysets.
```

sobolevn (commented):
final comments 🎉

docs/examples/integrations/cursor_pagination.py
s3rius: Yup! I forgor.

docs/pages/pagination.rst
s3rius: That's not quite true. If it's not stably ordered, we make it stable by adding primary keys to the queryset.

sobolevn (approved):
🎉 🎉 🎉 🎉 🎉 

this is huge! thank you!

PR #1442, opened 2026-09-12
Add tests for custom path converter OpenAPI schemas

### LINKED ISSUE #1440: Improve path customization tests, `__dmr_converter_schema__` is never tested
Currently we claim to support `__dmr_converter_schema__` attribute to customize how Django path converters generate schema, but we never actually test it.

See https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/dmr/openapi/generators/component_parsers.py#L117-L123

We need at least two tests:
- Valid customization
- Invalid one that will fail the final schema validation

adityabagla7:
## AI Policy

- [x] I have read and agree to the AI Policy, removed any "Co-Authored-By" lines attributing coding agents, and manually reviewed the final result

## Checklist

## Related issues

- Closes #1440

`__dmr_converter_schema__` is implemented and documented, but nothing called it. This adds the two cases from the issue:

- a custom converter that declares `int` shows up as an integer path parameter
- a custom converter whose declared schema is invalid fails OpenAPI spec validation

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tests/test_unit/test_openapi/test_converter_schema.py
+++ b/tests/test_unit/test_openapi/test_converter_schema.py
@@ -1,0 +1,103 @@
+from typing import Final
+
+import pydantic
+import pytest
+from django.urls import path, register_converter
+
+from dmr import Controller
+from dmr.openapi import build_schema
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.routing import Router
+
+OpenAPIValidationError = pytest.importorskip(
+    'openapi_spec_validator.validation.exceptions',
+).OpenAPIValidationError
+
+_YEAR_CONVERTER: Final = 'dmr_year'
+_INVALID_CONVERTER: Final = 'dmr_invalid'
+
+
+class _YearConverter:
+    """Maps a 4-digit year and declares an integer OpenAPI schema."""
+
+    regex = '[0-9]{4}'
+    __dmr_converter_schema__ = int
+
+    def to_python(self, value: str) -> int:  # noqa: WPS110
+        """Parse the captured path segment."""
+        return int(value)
+
+    def to_url(self, value: int) -> str:  # noqa: WPS110
+        """Render the year back into a URL segment."""
+        return str(value)
+
+
+class _InvalidSchemaType(pydantic.BaseModel):
+    """Type whose generated schema fails OpenAPI spec validation."""
+
+    model_config = pydantic.ConfigDict(
+        json_schema_extra={'minItems': -1},
+    )
+
+
+class _InvalidConverter:
+    """Custom converter whose declared schema fails spec validation."""
+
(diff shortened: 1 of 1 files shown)


tests/test_unit/test_openapi/test_converter_schema.py
```diff
@@ -0,0 +1,103 @@
...
+                    f'articles/<{_YEAR_CONVERTER}:year>/',
+                    _ArticleController.as_view(),
+                ),
+            ],
+        ),
+    ).convert()
+
+    operation = schema['paths']['/api/articles/{year}/']['get']
+    year_schema = None
+    for parameter in operation['parameters']:
```
sobolevn: use `snapshot()` to check all the parameters at once :)

tests/test_unit/test_openapi/test_converter_schema.py
```diff
@@ -0,0 +1,103 @@
...
+            break
+    assert year_schema is not None
+    assert year_schema['type'] == 'integer'
+
+
+def test_converter_schema_invalid() -> None:
+    """Ensure an invalid ``__dmr_converter_schema__`` fails spec validation."""
+    register_converter(_InvalidConverter, _INVALID_CONVERTER)
+
+    with pytest.raises(OpenAPIValidationError):
```
sobolevn: always use `match=` to assert the error message

tests/test_unit/test_openapi/test_converter_schema.py
```diff
@@ -0,0 +1,103 @@
...
+import pydantic
+import pytest
+from django.urls import path, register_converter
+
+from dmr import Controller
+from dmr.openapi import build_schema
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.routing import Router
+
+OpenAPIValidationError = pytest.importorskip(
```
sobolevn: You should not skip the whole module, you can only skip a single test

tests/test_unit/test_openapi/test_converter_schema.py
adityabagla7: Done — now snapshot()s the full parameter list so every field is checked at once.

tests/test_unit/test_openapi/test_converter_schema.py
adityabagla7: Done — the invalid-schema test now uses match='minItems'.

tests/test_unit/test_openapi/test_converter_schema.py
adityabagla7: Done — dropped the module-level importorskip. Only the invalid-schema test skips when openapi-spec-validator is missing.

PR #1443, opened 2026-09-12
test: cover __dmr_converter_schema__ path converter customization

### LINKED ISSUE #1440: Improve path customization tests, `__dmr_converter_schema__` is never tested

HarshRajSinghania:
## Summary

Adds tests for the documented `__dmr_converter_schema__` attribute on custom Django path converters.

Closes #1440

## Motivation

`ComponentParserGenerator` already reads `__dmr_converter_schema__` when building path-parameter schemas, but that path was untested.

## Implementation

- Valid customization: register a converter whose `__dmr_converter_schema__` is `int` and assert the generated path parameter schema type is `integer`.
- Invalid customization: register a converter whose schema type cannot be serialized and assert `UnsolvableAnnotationsError` is raised during `build_schema(...).convert()`.

## Testing

Behavior of both cases was checked against the current generator by constructing a `Router` with `register_converter` and calling `build_schema(...).convert()`:

- valid converter produced `{type: integer}` for the path parameter
- unserializable converter type raised `UnsolvableAnnotationsError` matching `Cannot generate OpenAPI schema`

Full project `just unit` was not run in this environment because the sandbox could not install the complete extra/dev stack (system PyJWT / missing venv tooling). CI on this PR should run the usual unit and lint jobs.

### DIFF
--- a/tests/test_unit/test_openapi/test_schema_snapshots.py
+++ b/tests/test_unit/test_openapi/test_schema_snapshots.py
@@ -3,7 +3,7 @@ from http import HTTPStatus
 from typing import Annotated, ClassVar, TypeAlias
 
 import pydantic
-from django.urls import path, re_path
+from django.urls import path, re_path, register_converter
 from syrupy.assertion import SnapshotAssertion
 
 from dmr import Body, Controller, Cookies, Path, Query, ResponseSpec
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
Sorry, https://github.com/wemake-services/django-modern-rest/pull/1442 was first. Please, take some other issue and please make sure to read our AI policy :)

PR #1444, opened 2026-09-12
Remove OpenAPI 3.0.x support

### LINKED ISSUE #1435: Remove OpenAPI 3.0.x support
We claimed that we supported OpenAPI 3.0.x, but this was never the case, we always supported a subset of 3.0.x, because only 3.1 started supporting JsonSchema that we use for our models.

We need to remove the support from both docs and code.

vyhuholl:
# I have made things!

`OpenAPIConfig` now raises `ValueError` for any `openapi_version` below `3.1.0`.

There was no 3.0 code path to remove. Nothing branches on the version except `openapi_version_info >= (3, 2)` in `dmr/metadata.py:139`, for `itemSchema`. The string went straight into the `openapi` field at `dmr/openapi/core/merger.py:27` and nothing else read it.

It did not even produce a valid 3.0 document. With the `[openapi]` extra installed, the test app's schema declared as `3.0.3` fails `openapi-spec-validator` on `info.license.identifier`, and on `info.summary` once you drop the license. Both are 3.1-only fields, so the validator stops there and never reaches the model schemas the issue is about.

The check sits in `__post_init__` and compares `(major, minor)`, so `'3.1'` passes along with `'3.1.1'` and `'3.2.0'`. You get the error when you build the config, not when you build the schema.

I left `load_schema` alone. It reads a foreign spec into our dataclasses rather than generating one, and the fixture behind it, `django-allauth.yml`, is a 3.0.3 document. Refusing to read those is a different call from the one the issue asks for, so I left it.

The changelog entry went into a new `0.16.0 WIP` section, since 0.15.0 is already out and this breaks anyone passing `3.0.x`. Renumber it if you had something else in mind.

## AI Policy

## Checklist

## Related issues

Closes #1435

### DIFF
--- a/dmr/openapi/config.py
+++ b/dmr/openapi/config.py
@@ -81,6 +81,10 @@ class OpenAPIConfig:
             ValueError: if ``openapi_version`` is older than ``'3.1.0'``.
 
         """
+        # NOTE: we don't limit the upper bound on purpose,
+        # we would only limit in the future if case
+        # of a real incompatibility. There's a high chance
+        # that it will just work (c)
         if self.openapi_version_info[:2] < _MIN_OPENAPI_VERSION:
             raise ValueError(
                 'OpenAPI versions before 3.1.0 are not supported, because '
--- a/dmr/openapi/config.py
+++ b/dmr/openapi/config.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import cast
+from typing import Final, cast
 
 from dmr.openapi.objects import (  # noqa: WPS235
     Components,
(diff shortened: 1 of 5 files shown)


dmr/openapi/config.py
```diff
@@ -64,6 +73,21 @@ class OpenAPIConfig:
...
 
+    def __post_init__(self) -> None:
+        """
+        Validates that ``openapi_version`` is supported.
+
+        Raises:
+            ValueError: if ``openapi_version`` is older than ``'3.1.0'``.
+
+        """
+        if self.openapi_version_info[:2] < _MIN_OPENAPI_VERSION:
```
sobolevn, line 88:
```suggestion
        # NOTE: we don't limit the upper bound on purpose,
        # we would only limit in the future if case
        # of a real incompatibility. There's a high chance
        # that it will just work (c)
        if self.openapi_version_info[:2] < _MIN_OPENAPI_VERSION:
```

sobolevn (approved):

PR #1449, opened 2026-09-12
Increase default DMR_MAX_CACHE_SIZE to 1024

### LINKED ISSUE #1448: Increase `DMR_MAX_CACHE_SIZE` to be `1024`
https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/dmr/envs.py#L8

I think that we defined a very small default for this value. There can be way more schemes than `256` on a big project. We want to have cache by default on them.

So, use a bigger number :)

wized2:
Fixes #1448

The default LRU cache size of 256 is too small for projects with many schemas. This raises \`DMR_MAX_CACHE_SIZE\` to **1024** so more schemas stay cached by default.

- \`dmr/envs.py\` default: \`256\` → \`1024\`
- docs \`configuration.rst\` updated to match
- changelog note under Unreleased

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,13 @@ are stored as descriptions in version releases on GitHub, example:
 https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
 
 
+## Unreleased
+
+### Features
+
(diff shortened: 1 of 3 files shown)


sobolevn (approved):
Thanks!

wized-1:
> Thanks!

No prob!

PR #1451, opened 2026-09-12
Explain why `openapi_examples_seed` is global only

### LINKED ISSUE #1436: Explain why `openapi_examples_seed` option is not supported per controller and per endpoint
Most of our settings can be modified on endpoint and controller levels. But, not this one. Why? Because schemas are generated only globally and can be reused by multiple endpoints and controllers. We need to explain this in the setting docs.

vyhuholl:
# I have made things!

Most settings can be set per controller and per endpoint. `openapi_examples_seed` cannot, and the docs did not say why.

The reason is where the generated example ends up. `generate_example` runs while a `Schema` is being built for `components/schemas`, and `SchemaRegistry.register` keys those by name: the second endpoint that needs the same type gets a `$ref` back, and the schema already registered is left alone. So the type becomes a schema once, the example is generated once with it, and every endpoint that uses the type points at that one entry. Nothing there belongs to a single endpoint, so there is no local seed to read. If two endpoints sharing a type asked for different seeds, whichever one the generator reached first would win.

The long version is a note under the setting in `docs/pages/configuration.rst`. `docs/pages/openapi/openapi.rst` gets two sentences and a link to it, because the "Examples generation" section is where people meet the feature before they go looking for the setting.

One thing in the diff that is not the issue: the snippet under that setting imported `HttpSpec` and `OpenAPIConfig` and used neither, which read as if the setting had something to do with them. Dropped both.

I am not sure the second note is wanted. If the explanation belongs only in the settings reference, say so and I will drop it from `openapi.rst`.

## AI Policy

## Checklist

## Related issues

Closes #1436

### DIFF
--- a/docs/pages/configuration.rst
+++ b/docs/pages/configuration.rst
@@ -480,12 +480,33 @@ OpenAPI
   .. code-block:: python
     :caption: settings.py
 
-    >>> from dmr.settings import HttpSpec
-    >>> from dmr.openapi.config import OpenAPIConfig
-
     >>> DMR_SETTINGS = {
     ...     Settings.openapi_examples_seed: 10,
     ... }
+
+  .. note::
+
+    Unlike most other settings, this one can only be set globally.
+    There are no controller and endpoint level counterparts for it.
+
+    Examples are attached to schemas in ``components/schemas``,
+    and schemas are generated per spec, not per endpoint:
+    a type is turned into a schema once,
+    and every endpoint and controller that uses this type
+    points to that single schema with a ``$ref``.
+
+    So, a schema does not belong to any particular endpoint
+    or controller, and there is nothing to read a local seed from.
+    If several endpoints sharing a schema declared different seeds,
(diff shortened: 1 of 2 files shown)


docs/pages/configuration.rst
```diff
@@ -480,13 +480,34 @@ OpenAPI
...
+    If several endpoints sharing a schema declared different seeds,
+    the winner would be whichever endpoint happened
+    to be visited first during the spec generation.
+    We prefer one predictable seed
+    over a value that depends on the traversal order.
+
+    Hand-written examples don't have this problem,
+    because they are defined on the type itself,
+    which is exactly where the schema comes from.
+    See :ref:`openapi-examples-generation`.
```
sobolevn: Please, make this shorter. Just one paragraph like in the second doc should be fine, but keep the `ref` :)

vyhuholl: Done

sobolevn (approved):
Thanks!

PR #1459, opened 2026-09-12
Copy `re_path` group patterns into parameter schemas

### LINKED ISSUE #1439: Improve `re_path` default parameter schema
Currently we default to `str` as schema for all patterns in `re_path`. However, this is not the only thing we can do. We can also copy the pattern itself to the schema. Example:

```python
re_path(r'^v(?P<version>\d+)/$', VersionController.as_view())  # schema: {type: string}
```

We can copy the pattern as well:

```python
# re_path (?P<version>\d+)
- name: version
  in: path
  schema: {type: string, pattern: '^\d+$'}
```

This way users would have more info about what is expected.
See https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/dmr/openapi/generators/component_parsers.py#L136-L152 for current schema generation.

vyhuholl:
# I have made things!

Named groups of `re_path()` routes now carry their own regex into the schema, so `r'^v(?P<version>\d+)/$'` documents `version` as `{'type': 'string', 'pattern': '^(?:\d+)$'}` and not just `{'type': 'string'}`. The type stays `str`, a regex does not tell us anything better than that.

Getting the sub-pattern out needs the group boundaries and `re` does not expose them, so `dmr/internal/regex.py` scans the source with one tokenizer regex that matches escapes and character classes too, because `\(` and `[(]` must not count as group starts. `_add_group_patterns` then writes the pattern onto the parameters that were just built from the same route.

Two decisions worth a look:

- The sub-pattern is wrapped before it is anchored: `^(?:json|xml)$`, not the `^\d+$` from the issue. A JSON Schema `pattern` is an unanchored search, and a top-level alternation anchored as `^json|xml$` says something else entirely. Four characters per parameter buys that.
- Sub-patterns are copied verbatim, so a nested named group reaches the spec as `(?P<year>\d{4})`, which is Python syntax and not the ECMA-262 that JSON Schema asks for. Translating regex dialects is a much bigger job than this issue, so the caveat is in the docs instead.

Only `re_path()` is touched here. Custom `path()` converters have a `.regex` too and could get the same treatment, but that is its own change.

`parse_named_groups` is tested against `re.compile(...).groupindex` on every case: whatever group `re` finds, the parser has to find as well. The snapshot in `test_raw_path_schema` picked up the two new patterns.

If you would rather have the plain `^\d+$` from the issue, I can wrap only when the sub-pattern has a top-level `|`.

## AI Policy

## Checklist

## Related issues

Closes #1439

### DIFF
--- a/dmr/internal/regex.py
+++ b/dmr/internal/regex.py
@@ -46,8 +46,8 @@ def parse_named_groups(source: str) -> dict[str, str]:
 
 def _closed_named_group(
     source: str,
-    opened: 're.Match[str]',
-    closed: 're.Match[str]',
+    opened: re.Match[str],
+    closed: re.Match[str],
 ) -> dict[str, str]:
     group_name = opened.group('name')
     if group_name is None:  # unnamed group, like `(\d+)`
--- a/dmr/internal/regex.py
+++ b/dmr/internal/regex.py
@@ -1,0 +1,55 @@
+import re
+from typing import Final
+
+#: Tokens that can change the group nesting inside a regex pattern.
+#: Escaped characters and character classes are matched to be skipped:
+#: `\(` and `[(]` do not open any groups.
+_GROUP_TOKENS: Final = re.compile(
+    r"""
+    \\. |                                # escaped char, like `\(`
+    \[ \^? \]? (?: [^\]\\] | \\. )* \] | # char class, like `[^\]]`
+    \(\?P<(?P<name>\w+)> |               # named group, like `(?P<year>`
+    \( |                                 # any other group start
+    \)                                   # group end
+    """,
+    re.VERBOSE | re.DOTALL,
+)
+
+
+def parse_named_groups(source: str) -> dict[str, str]:
+    r"""
+    Parse sub-patterns of all named groups in a regex pattern.
+
+    We only look at the group nesting, everything that cannot open
+    or close a group is skipped: like ``(`` in ``\(`` and in ``[(]``.
+    Unmatched ``)`` tokens are ignored, they can only happen
+    in patterns we don't fully understand, like verbose mode comments.
+
+    Args:
+        source: Regex pattern source, like ``r'^(?P<year>\d{4})/$'``.
+
+    Returns:
+        Named groups with their sub-patterns, like ``{'year': r'\d{4}'}``.
+
+    """
+    named_groups: dict[str, str] = {}
+    opened_groups: list[re.Match[str]] = []
+    for token in _GROUP_TOKENS.finditer(source):
+        if token.group().startswith('('):
+            opened_groups.append(token)
+        elif token.group() == ')' and opened_groups:
+            named_groups.update(
+                _closed_named_group(source, opened_groups.pop(), token),
+            )
+    return named_groups
+
+
+def _closed_named_group(
+    source: str,
+    opened: 're.Match[str]',
+    closed: 're.Match[str]',
+) -> dict[str, str]:
+    group_name = opened.group('name')
+    if group_name is None:  # unnamed group, like `(\d+)`
+        return {}
+    return {group_name: source[opened.end() : closed.start()]}
--- a/dmr/openapi/generators/component_parsers.py
+++ b/dmr/openapi/generators/component_parsers.py
@@ -1,11 +1,12 @@
 import dataclasses
 import uuid
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Any, ClassVar, TypeAlias
+from typing import TYPE_CHECKING, Any, ClassVar, TypeAlias, cast
 
 from django.urls import URLPattern, converters
 from typing_extensions import TypedDict
 
+from dmr.internal.regex import parse_named_groups
 from dmr.openapi.objects import (
     MediaType,
     Parameter,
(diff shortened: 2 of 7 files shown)


dmr/internal/regex.py
```diff
@@ -0,0 +1,55 @@
...
+    Args:
+        source: Regex pattern source, like ``r'^(?P<year>\d{4})/$'``.
+
+    Returns:
+        Named groups with their sub-patterns, like ``{'year': r'\d{4}'}``.
+
+    """
+    named_groups: dict[str, str] = {}
+    opened_groups: list[re.Match[str]] = []
+    for token in _GROUP_TOKENS.finditer(source):
```
sobolevn: well, I am not quite sure about this at all :) 
this solution seems very brittle, I can really understand what is going on here.

why can't we use `replace_named_groups` from django itself? what is the benefit of our own design?

dmr/internal/regex.py
```diff
@@ -0,0 +1,55 @@
...
+            named_groups.update(
+                _closed_named_group(source, opened_groups.pop(), token),
+            )
+    return named_groups
+
+
+def _closed_named_group(
+    source: str,
+    opened: 're.Match[str]',
+    closed: 're.Match[str]',
```
sobolevn:
```suggestion
    opened: re.Match[str],
    closed: re.Match[str],
```

you can use real object here

dmr/openapi/generators/component_parsers.py
```diff
@@ -141,16 +142,40 @@ def _parse_pattern(
...
                 ),
             )
         return params_list or None
 
+    def _add_group_patterns(
+        self,
+        params_list: list[Parameter | Reference],
+        regex_source: str,
+    ) -> list[Parameter | Reference]:
+        # Named groups are wrapped, because otherwise top-level alternations
```
sobolevn: Sorry, I don't understand this comment. Please, explain this better, maybe with examples.

dmr/openapi/generators/component_parsers.py
```diff
@@ -141,16 +142,40 @@ def _parse_pattern(
...
+        # like `a|b` would only be anchored on one side: `^a|b$`.
+        named_groups = {
+            group_name: f'^(?:{group_source})$'
+            for group_name, group_source in parse_named_groups(
+                regex_source,
+            ).items()
+        }
+        # Named groups are the only source of `re_path()` parameters,
+        # so all the parameters here are plain `str` schemas we've just built.
+        for param_spec in cast('list[Parameter]', params_list):
```
sobolevn: I would prefer to have an API without the `cast` :)

dmr/internal/regex.py
vyhuholl: **Why not `replace_named_groups`:** because `replace_named_groups` throws away subpatterns — the exact part that we need (for example, `(?P<year>\d{4})` will be converted to `<year>`, skipping the `\d{4}` part). And Django exposes no public way to get the sub-pattern back.

The private `_get_group_start_end` (which `replace_named_groups` is built on) can do it, but it is not in `django-stubs` — importing it would need `noqa: PLC2701` plus suppressions for all four type checkers. Also, `_get_group_start_end` does not understand character classes: on `(?P<x>[)]+)` it would return `[`, and we would publish a broken regex into the schema.

So I simplified the code by removing the custom tokenizer. `regex.py` now reuses Django's `named_group_matcher` for the group start and asks `re` itself for the end: the first `)` that leaves a compilable sub-pattern behind is the closing one. Any earlier `)` is either escaped, inside a character class, or closing a nested group — and all three leave a prefix that does not compile `(\d+\`, `[0-9`, `(x|y)`.

dmr/openapi/generators/component_parsers.py
vyhuholl: Rewritten with examples: `pattern` in JSON Schema is a search, while a url group matches the whole value, hence the anchors; the wrapper is needed because anchors bind weaker than `|` — `^json|xml$` means "starts with `json`" or "ends with `xml`", while `^(?:json|xml)$` is what `(?P<format>json|xml)` actually matches.

dmr/openapi/generators/component_parsers.py
vyhuholl: Replaced `cast`s, by `assert isinstance` with the invariant spelled out — the parameters are built from a `TypedDict` of `str` fields, one per named group, so their schemas are always inline. There is precedent in the repo ([dmr/files.py:234](https://github.com/wemake-services/django-modern-rest/blob/master/dmr/files.py#L234)) , and it does not affect branch coverage. Widening `ParameterGenerator.__call__` to return `list[Parameter]` did not work out: `list` is invariant, so it would break the four `get_schema` implementations in `components.py` that must return `list[Parameter | Reference]`.

Side effect: `[)]` inside a character class now parses correctly (Django's own scanner has a blind spot there), and groups like `(?P<same>\1)` simply end up without a `pattern` — both are covered by tests.

sobolevn (commented):
please, test that a controller with `re_path` and `Path[]` works correctly, `Path[]` must have a priority.

this test might exist already. then, it is all fine :)

vyhuholl:
Tested that `Path` component wins over `re_path` groups.

sobolevn (approved):

Issue #1437, opened 2026-09-12
Make sure all `bool` values in `dmr.openapi.objects` can also accept `None`


sobolevn:
Why is this important? Because we ignore `None` values from the schema dump. But, now all `deprecated: False` objects are dumped. This is the default, no need to do this.

We need to:
- Update `bool = False` to be `bool | None = None`
- Update all snapshots to remove exising values

All `None`s will be ignored.

max0nT:
Let me do it

sobolevn:
Please, go ahead! Please, don't forget about the refactoring task as well :)

PR #1463, opened 2026-09-12
Let `slug` and `path` converters customize the parameter schema

### LINKED ISSUE #1441: Support `SlugConverter` and `PathConverter` from Django in path parameters
Currently we only support `int` and `uuid` converters from django: https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/dmr/openapi/generators/component_parsers.py#L36-L41

However, there are two more that we can support: 
- `path`, we can add `{"description": "Can contain slashes"}` to the schema
- `slug`, we can add the pattern from `SlugConverter`

This needs to be tested. Probably we can also change the storage API. I am not sure how currently represent in for all possible providers. Probably we can store prepared schemas for types that exist.

helo060228:
Closes #1441

## What

`slug` and `path` converters now bring their own info into the path
parameter schema:

```yaml
# path('tags/<slug:tag>/', ...)
- name: tag
  in: path
  schema: {type: string, pattern: '^(?:[-a-zA-Z0-9_]+)$'}

# path('files/<path:file_path>/', ...)
- name: file_path
  in: path
  schema: {type: string, description: 'Can contain slashes'}
```

`int` and `uuid` stay as they are.

## How

The issue asked whether the storage API should change, so `_converters`
now maps a converter class to a prepared `ConverterSchema` (`model` plus
an optional `pattern` and `description`) instead of a bare type.

`ConverterSchema` is public and `@final`, and it is exported from
`dmr.openapi`, so custom converters can use it too:

```python
from dmr.openapi import ConverterSchema


class LowercaseConverter:
    regex = '[a-z]+'
    __dmr_converter_schema__ = ConverterSchema(
        model=str,
        pattern='^(?:[a-z]+)$',
        description='Lowercase letters only',
    )
```

A plain model (`__dmr_converter_schema__ = int`) keeps working.

Values that you provide are applied after the generated ones, so they
always override what we generate from `model`. If `model` is a serializer
model, the generated schema is a component reference, and a reference
cannot be extended inline, so `pattern` and `description` are ignored
for it. Both rules are documented.

## Checks

- [x] `ruff check`, `ruff format --check`
- [x] `flake8` (wemake-python-styleguide), `slotscheck`
- [x] `mypy`, `pyright`, `pyrefly`, `ty`
- [x] `import-linter`
- [x] tests use the public API: `build_schema(Router(...))` with
      `inline_snapshot`, next to the existing converter tests
- [x] `CHANGELOG.md` and `docs/pages/components/path.rst` updated,
      including a detailed explanation of the anchoring


dmr/openapi/generators/component_parsers.py
```diff
@@ -23,7 +23,54 @@
...
+    model: Any = str
+    pattern: str | None = None
+    description: str | None = None
+
+
+_ConvertersMapping: TypeAlias = Mapping[type[Any], _ConverterSchema]
+
+
+def _anchored(regex: str) -> str:
+    """Anchor a regex on both sides with a non-capturing group.
```
sobolevn: Sorry, why non-capturing group is required? Please, explain this in great detail in the docs.

dmr/openapi/generators/component_parsers.py
```diff
@@ -23,7 +23,54 @@
...
+    prepared: Mapping[str, _ConverterSchema],
+) -> dict[str, Any]:
+    """Build the ``TypedDict`` fields for the prepared converters."""
+    return {
+        converter_name: converter_schema.model
+        for converter_name, converter_schema in prepared.items()
+    }
+
+
+def _with_converter_schema(
```
sobolevn: Please, move all the protected functions after the public class

dmr/openapi/generators/component_parsers.py
```diff
@@ -34,8 +81,14 @@ class ComponentParserGenerator:
 
     # Class API:
     _converters: ClassVar[_ConvertersMapping] = {
-        converters.IntConverter: int,
-        converters.UUIDConverter: uuid.UUID,
+        converters.IntConverter: _ConverterSchema(model=int),
```
sobolevn: I don't quite like this design. Now, imagine that some person wants to do something similar with `__dmr_converter_schema__`. For example: provide a model with a description for all possible serializers. How can one do that? 

Maybe we should do this model `@final` and public? And document that this feature is possible?

dmr/openapi/generators/component_parsers.py
```diff
@@ -23,7 +23,54 @@
...
+    param_spec: Parameter | Reference,
+    prepared: Mapping[str, _ConverterSchema],
+) -> Parameter | Reference:
+    """Add the converter specific info to a generated parameter."""
+    if not isinstance(param_spec, Parameter):
+        return param_spec
+    converter_schema = prepared.get(param_spec.name)
+    if converter_schema is None or not isinstance(param_spec.schema, Schema):
+        return param_spec
+    if converter_schema.pattern is not None:
```
sobolevn: Please, document that providing explicit `ConverterSchema` will always override any generated values.

tests/test_unit/test_openapi/test_generators/test_component_generator.py
```diff
@@ -63,3 +68,73 @@ def test_fake_component(openapi_context: OpenAPIContext) -> None:
...
+    assert path_params is not None
+    schema = _single_schema(path_params)
+    assert schema.pattern == expected_pattern
+    assert schema.description == expected_description
+
+
+def test_reference_keeps_its_schema() -> None:
+    """Ensure that we never touch references."""
+    reference = Reference(ref='#/components/schemas/User')
+    assert _with_converter_schema(reference, {}) is reference
```
sobolevn: Sorry, these tests are not really correct. Please, test public API. I am not sure that they are even realistic.

tests/test_unit/test_openapi/test_generators/test_component_generator.py
```diff
@@ -63,3 +68,73 @@ def test_fake_component(openapi_context: OpenAPIContext) -> None:
...
+        ('files/<path:file_path>/', None, 'Can contain slashes'),
+        ('articles/<int:article_id>/', None, None),
+        ('users/<uuid:user_id>/', None, None),
+        ('plain/<str:name>/', None, None),
+        ('bare/<name>/', None, None),
+    ],
+)
+def test_path_converter_schemas(
+    openapi_context: OpenAPIContext,
+    raw_path: str,
```
sobolevn:
```suggestion
    *,
    raw_path: str,
```

helo060228:
Thanks for the review! All points are addressed in the latest push.

**1. Why the non-capturing group is required**

Explained in detail in `docs/pages/components/path.rst`, in a new
"Copying regexes into the schema" section, plus a code comment next to
the pattern. The short version:

- In JSON Schema, `pattern` is a *search*, it matches anywhere inside the
  value, while a url parameter always matches the whole value. That's why
  we anchor it on both sides.
- Anchors bind weaker than the `|` operator. Adding `^` and `$` around
  `json|xml` produces `^json|xml$`, which means "starts with `json`"
  *or* "ends with `xml`". Wrapping the regex into a non-capturing group
  keeps the meaning of the original branch: `^(?:json|xml)$`.

**2. Protected functions after the public class**

Done. Nothing private is defined before `ComponentParserGenerator`
anymore, the helpers live at the bottom of the file.

**3. The design**

`ConverterSchema` is now public and `@final`, and it is exported from
`dmr.openapi`, so a custom converter can describe itself with a model
*and* extra keywords:

```python
from dmr.openapi import ConverterSchema


class LowercaseConverter:
    regex = '[a-z]+'
    __dmr_converter_schema__ = ConverterSchema(
        model=str,
        pattern='^(?:[a-z]+)$',
        description='Lowercase letters only',
    )
```

A plain model (`__dmr_converter_schema__ = int`) still works. Both forms
are covered by tests now.

Happy to move it to another module if `dmr.openapi` is not the home you
would pick for it.

**4. Overriding generated values**

Documented. We generate from `model` first, and only then apply the
values that were provided explicitly, so `pattern` and `description`
always override the generated ones.

While documenting it I found one case worth calling out: if `model` is
a serializer model, the generated schema is a component reference, and
a reference cannot be extended inline, so `pattern` and `description`
are ignored for it. That is documented as well.

**5. Testing the public API**

The private helper tests are gone. The new tests build a real schema with
`build_schema(Router(...))` and assert with `inline_snapshot`, in
`test_converter_schema.py`, next to the existing converter tests.

**6. `*,`**

The test that had that signature was replaced by the public-API tests
above, so there is nothing to mark keyword-only at the moment. Noted for
the next parametrized test I write.

helo060228:
While updating this I noticed #1471, which covers the same issue from a
branch in this repository. I had not seen it when I started here: it was
opened after this PR, and there was no `issue-1441` branch or PR when I
checked the issue.

Both are yours to pick from. If you want this one to go a different
direction, or to drop it in favour of the other one, just say so.

helo060228:
An update after checking the full timeline, because my earlier note was incomplete.

```
09-12 11:02 UTC  @vyhuholl self-assigned this issue
09-12 11:22 UTC  this PR was opened
09-12 11:57 UTC  #1471 was opened
```

I checked this issue before starting and there was no assignee, no branch and no PR at that moment, so this work was not started against a claim. But my final check before opening the PR only looked for a branch and a PR, and it missed the assignment that was already there 20 minutes earlier. That one is on me, and the assignee is part of that check now.

To be explicit about where I stand: if you go by "the assignee holds the claim", I am happy to close this PR — no hard feelings, and I would rather not waste your review time. If you would rather keep both options open, this one is up to date with your review and its checks are green.

Either way, thanks for the review. The design notes were useful, and the public `ConverterSchema` is a better result than what I opened with.

helo060228:
Rebased onto `d3ff707`, because the PR had started conflicting with today's merges.

- `CHANGELOG.md` conflicted with the entries that landed today. I kept the upstream entry next to mine, nothing was dropped.
- `tests/test_unit/test_openapi/test_converter_schema.py` needed one snapshot update: `#1437` made bool objects accept `None`, so `deprecated` is no longer generated when it is unset, and my snapshots still carried `'deprecated': False`. They now match the current output.

The two failures remaining in `tests/test_unit/test_openapi/test_core/test_context_customization.py` are **not from this PR**. `d3ff707` is red on `master` itself, and a clean `d3ff707` checkout fails the same two tests with the same diff. Reported separately in #1535 so it does not get mixed up with this one.

`ruff check` and `ruff format --check` pass, and `tests/test_unit/test_openapi` is otherwise green.

helo060228:
Rebased again, onto `2c77fd8` (#1536). The two `test_context_customization` failures I mentioned above are gone now that the fix is in the base, so the noise is off this PR's checks.

Locally `tests/test_unit/test_openapi` is green: 242 passed, 35 snapshots.
One note on the checks: the Read the Docs build on `3e1de0b` failed in **3 seconds** with *"Error while checking out the repository"*, which is their side running into the force-push, not a docs problem. The previous commit, which has the same docs content, built fine in 236s. I am not pushing an empty commit just to re-trigger it, so that the workflow approvals stay as they are.

sobolevn (approved):
Thank you!

PR #1465, opened 2026-09-12
Support default values for TypeVar in controller type var inference

### LINKED ISSUE #1452: Support `default` values for `TypeVar` when using controller type var inference
We have a `TODO` here: https://github.com/wemake-services/django-modern-rest/blob/405ba9b05fe583954d49683be168a9fc584a3169/dmr/types.py#L318-L326

We either need to:
1. Support the typevars with defaults, add tests, remove `TODO`
2. Or just add tests and remove `TODO` if it is already supported

TheNetherWatcher:
Fixes #1452

## Summary

`TypeVarInference` (used to resolve generic type vars declared on `Controller` subclasses) ignored `TypeVar(..., default=...)` (PEP 696) values: if a type var was never bound to a concrete type anywhere in the inheritance chain, resolution would eventually raise `UnsolvableAnnotationsError` instead of falling back to the declared default.

In the common case (a type var re-declared and eventually concretely subscribed through the inheritance chain), Python's own runtime substitution already fills in defaults transparently, so this was already working. The gap was the edge case where a generic base is inherited without ever being subscripted at all (e.g. `class Concrete(Base):` where `Base` is `Generic[T]` with `T` having a default) — `__orig_bases__` doesn't even list `Base` in that case, so dmr's own walk never saw a concrete value and the type var stayed unresolved.

## Changes

- `dmr/types.py`: `TypeVarInference._infer`/`_resolve` now check `type_param.has_default()` when a type var can't be resolved further, and use `type_param.__default__` instead of raising.
- Removed the now-resolved `TODO` about type var defaults.
- Added tests in `tests/test_unit/test_controllers/test_reusable_generics.py`:
  - `test_typevar_default_used_when_unbound`: full `Controller` chain with a `default=`-carrying `TypeVar`, proving both "left unbound → uses default" and "explicitly overridden → uses override".
  - `test_infer_annotation_uses_default`: exercises the specific fallback branch directly (fully unparametrized base).
- `CHANGELOG.md` entry.

## Testing

- `just lint` — passes (ruff, flake8/wemake-python-styleguide, slotscheck, import-linter)
- `uv run python -m mypy .` — passes
- `just unit` — 3367 passed, 11 skipped, 0 failed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
 ### Features
 
 - Increased default `DMR_MAX_CACHE_SIZE` from `256` to `1024`, #1448
+- `TypeVar` `default=` values (PEP 696) are now honored during controller
+  type var inference when a type var is left unbound, #1452
 
 
 ## 0.15.0 (2026-09-11)
--- a/dmr/types.py
+++ b/dmr/types.py
@@ -317,10 +317,6 @@ class TypeVarInference:
 
             is_needed = type_map.get(type_param.__name__)
             if is_needed:
-                # TODO: most likely this will require
-                # some extra work to support
-                # type var defaults. Right now defaults
-                # are ignored in the resolution.
                 type_map.update({type_param.__name__: type_arg})
                 if isinstance(type_arg, TypeVar):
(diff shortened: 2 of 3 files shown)


sobolevn:
Closing, there are many mistakes here.

PR #1468, opened 2026-09-12
Allow bool fields in dmr.openapi.objects to accept None

### LINKED ISSUE #1437: Make sure all `bool` values in `dmr.openapi.objects` can also accept `None`

TheNetherWatcher:
## Summary
- Change `bool = False` fields that are dumped into the OpenAPI schema (`Parameter.deprecated`, `Parameter.required`, `Operation.deprecated`, `XML.attribute`, `XML.wrapped`) to `bool | None = None`
- `None` values are already ignored by the schema dump, so this stops emitting the redundant default `false` (e.g. `"deprecated": false`) into every generated OpenAPI document
- Regenerated all affected snapshots; diffs only remove the now-implicit `false` keys, nothing else changed

Fixes #1437

## Test plan
- [x] `just lint`
- [x] `just unit` (full suite green; only pre-existing, environment-only failures remain: Redis-backend tests requiring local Docker)
- [x] Manually reviewed every snapshot diff to confirm only default-`false` keys were removed

### DIFF
--- a/dmr/endpoint.py
+++ b/dmr/endpoint.py
@@ -303,7 +303,8 @@ class Endpoint:  # noqa: WPS214
                 if self.metadata.description is None
                 else str(self.metadata.description)
             ),
-            deprecated=self.metadata.deprecated or router_metadata.deprecated,
+            deprecated=(self.metadata.deprecated or router_metadata.deprecated)
+            or None,
             security=context.generators.security_scheme(
                 self.metadata.auth,
                 serializer,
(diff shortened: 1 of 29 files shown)


dmr/endpoint.py
```diff
@@ -303,7 +303,8 @@ def get_schema(
                 if self.metadata.description is None
                 else str(self.metadata.description)
             ),
-            deprecated=self.metadata.deprecated or router_metadata.deprecated,
+            deprecated=(self.metadata.deprecated or router_metadata.deprecated)
+            or None,
```
sobolevn, line 307:
```suggestion
            deprecated=(self.metadata.deprecated or router_metadata.deprecated
            or None),
```

nit

PR #1469, opened 2026-09-12
Support customizable schema generation parameters for pydantic and msgspec plugins

### LINKED ISSUE #1462: Support better schema generation parameters
Currently we don't pass quite a lot of options to both `pydantic` and `msgspec` schema generation:
https://github.com/wemake-services/django-modern-rest/blob/405ba9b05fe583954d49683be168a9fc584a3169/dmr/plugins/pydantic/schema.py#L27-L30

When `pydantic` defines this method as:

```python
def json_schema(
    self,
    *,
    by_alias: bool = True,
    ref_template: str = DEFAULT_REF_TEMPLATE,
    union_format: Literal['any_of', 'primitive_type_array'] = 'any_of',
    schema_generator: type[GenerateJsonSchema] = GenerateJsonSchema,
    mode: JsonSchemaMode = 'validation',
) -> dict[str, Any]:
```

The same with `msgspec`: https://github.com/wemake-services/django-modern-rest/blob/405ba9b05fe583954d49683be168a9fc584a3169/dmr/plugins/msgspec/schema.py#L24-L27

Signature:

```python
def schema(
    type: Any,
    *,
    schema_hook: Optional[Callable[[type], dict[str, Any]]] = None,
    ref_template: str = _REF_TEMPLATE,
) -> dict[str, Any]:
```

So, what we need to do:
1. Provide way to easily customize the schema generation for both `pydantic` and `msgspec`
2. Both would have different API, matching the current defaults
3. Document how to customize the schema generation this way: create your own serializer, subclass schema generator, use it
4. Add tests for 2 of these cases, especially with custom `schema_hook` and `schema_generator` options

TheNetherWatcher:
## Summary

Both `pydantic` and `msgspec` accept extra parameters that control the whole schema generation process (not just a single model), but `django-modern-rest` never passed them through. This adds a way to customize those, matching each library's own idiomatic API:

- **pydantic**: `PydanticSchemaGenerator` gains overridable `schema_generator` (a `type[GenerateJsonSchema]`) and `union_format` class attributes, passed through to `TypeAdapter.json_schema(...)`.
- **msgspec**: `MsgspecSchemaGenerator` gains an overridable `schema_hook` class attribute (`Callable[[type], dict[str, Any]] | None`), passed through to `msgspec.json.schema(...)`.

Both default to today's exact behavior — no change for anyone not opting in. To customize, subclass the relevant `*SchemaGenerator`, override the attribute, and point your `Serializer` subclass's `schema_generator` at it (documented in `docs/pages/openapi/openapi.rst` under a new "Customizing schema generation parameters" section, with runnable examples for both plugins).

## Test plan

- [x] `just lint` (ruff, flake8/wemake, slotscheck, import-linter)
- [x] `uv run python -m mypy .` on changed/added files
- [x] `just unit` — full suite green, 100% coverage maintained
- [x] `just docs` (`-W`, warnings as errors) — new rst section builds clean
- [x] Added tests: custom `schema_generator` for pydantic (marker key injected via `GenerateJsonSchema` subclass), custom `schema_hook` for msgspec (handles an otherwise-unsupported type), plus default-behavior-unchanged tests for both

Fixes #1462

### DIFF
--- a/docs/examples/openapi/msgspec_schema_hook_customization.py
+++ b/docs/examples/openapi/msgspec_schema_hook_customization.py
@@ -1,0 +1,29 @@
+from decimal import Decimal
+from typing import Any
+
+from typing_extensions import override
+
+from dmr.plugins.msgspec import MsgspecSerializer
+from dmr.plugins.msgspec.schema import MsgspecSchemaGenerator
+
+
+class MyMsgspecSchemaGenerator(MsgspecSchemaGenerator):
+    """Plugs a custom `schema_hook` into msgspec schema generation."""
+
+    __slots__ = ()
+
+    @override
+    @classmethod
+    def schema_hook(cls, source_type: type) -> dict[str, Any]:
+        """Teach msgspec how to build a schema for extra types."""
+        if source_type is Decimal:
+            return {'type': 'string', 'format': 'decimal'}
+        raise NotImplementedError(f'Unsupported type: {source_type!r}')
+
+
+class MySerializer(MsgspecSerializer):
+    """Serializer using the customized schema generator."""
+
+    __slots__ = ()
+
+    schema_generator = MyMsgspecSchemaGenerator
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
 - Added `tags` controller attribute to apply OpenAPI tags
   to all endpoints of this controller. They are merged
   with router-level and endpoint-level tags, #1434
+- Allows customizing JSON schema generation for `pydantic` (via
+  `PydanticSchemaGenerator.schema_generator` and `.union_format`) and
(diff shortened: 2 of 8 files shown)


docs/examples/openapi/msgspec_schema_hook_customization.py
```diff
@@ -0,0 +1,29 @@
...
+class MyMsgspecSchemaGenerator(MsgspecSchemaGenerator):
+    """Plugs a custom `schema_hook` into msgspec schema generation."""
+
+    __slots__ = ()
+
+    @override
+    @classmethod
+    def schema_hook(cls, source_type: type) -> dict[str, Any]:
+        """Teach msgspec how to build a schema for extra types."""
+        if source_type is Decimal:
```
sobolevn, line 19: https://github.com/msgspec/msgspec/blob/0830fd1581e5d2d0a8d4dab3c941aec17cd30f41/src/msgspec/_json_schema.py#L292-L294

sobolevn (commented):
This is also incorrect :( 

Please, stop.

PR #1470, opened 2026-09-12
Cache negotiation header_value -> parser/renderer class resolution

### LINKED ISSUE #1455: Cache the negotiation parts
Most of the headers are simple and repeatable. Most users will just send `Content-Type: application/json` and `Accept: application/json` like 99% of the time. So, there's no need to run the negotiation process the second time. We already know the final result. So, go through the https://github.com/wemake-services/django-modern-rest/blob/master/dmr/negotiation.py and find places where we can cache `header_value -> parser_class` or `header_value -> renderer_class` computations.

Prefer not to have any API changes here.

TheNetherWatcher:
## Summary
- Adds a bounded per-endpoint cache (respecting `DMR_MAX_CACHE_SIZE`) for `RequestNegotiator`'s wildcard `Content-Type` resolution and for `ResponseNegotiator`'s `Accept` header resolution.
- Most clients send the same `Content-Type`/`Accept` value on every request (e.g. `application/json`), so repeated requests now skip re-running the O(n) wildcard-matching / renderer-negotiation algorithm and hit an O(1) cache lookup instead.
- No public API changes, as requested in the issue.
- Cache is keyed per-endpoint instance (not globally), so two endpoints with different configured parsers/renderers never share or contaminate each other's cached results.
- Once the cache reaches `DMR_MAX_CACHE_SIZE`, new values are simply not cached (still resolved correctly via the full algorithm), preventing unbounded memory growth.

Fixes #1455

## Test plan
- Added `tests/test_unit/test_negotiation/test_negotiation_caching.py` covering: repeated stable resolution, unsupported values are never falsely cached, cross-endpoint cache isolation, and correct behavior once the cache is full (bounded growth).
- `just lint` and `just unit` pass locally (100% coverage maintained).
- `mypy .` passes with no issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@ https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
 
 - Increased default `DMR_MAX_CACHE_SIZE` from `256` to `1024`, #1448
 
+### Misc
+
+- Caches `Content-Type` / `Accept` header negotiation results per endpoint,
+  so repeated requests with the same header value skip re-running
+  the negotiation algorithm, #1455
+
 
 ## 0.15.0 (2026-09-11)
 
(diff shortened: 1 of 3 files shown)


sobolevn (commented):
Closing as low quality.

PR #1471, opened 2026-09-12
Support slug and path converters in path parameters

### LINKED ISSUE #1441: Support `SlugConverter` and `PathConverter` from Django in path parameters

vyhuholl:
# I have made things!

Django's `slug` and `path` url converters now carry their own OpenAPI schemas: `slug` gets `pattern: ^[-a-zA-Z0-9_]+$`, `path` gets a `Can contain slashes` description. Until now both came out as plain strings, the same as any converter we do not know.

The issue asks whether the storage API has to change, and it did. `_converters` mapped a converter class to a single type, which is all the serializer needs, but `pattern` and `description` are not types. Values are now `_ConverterSpec(annotation, prepared_schema)`: the annotation still goes into the generated `TypedDict` for the serializer to build the base schema from, and `prepared_schema` is merged on top of the result. That is the only serializer-neutral place I found for it. `msgspec.Meta` or `pydantic.Field` constraints in the annotation would mean a plugin import inside `dmr.openapi`, and this way both serializers come out with the same keys.

The slug pattern is `converters.SlugConverter.regex` with `^` and `$` around it. Django's regexes are unanchored, because they are fragments of a bigger url regex, while JSON Schema `pattern` is a partial match — without the anchors, `a/b` passes as a slug.

`__dmr_converter_schema__` and the `re_path` branch are untouched, and no existing snapshot moved. The new test goes through `build_schema` for `int`, `uuid`, `str`, `slug`, `path` and a registered custom converter, with pydantic and msgspec both, and compares the dumped parameter without `title`, which belongs to the serializer. Docs get a table of all five converter schemas.

One call I am unsure about: the `path` description goes into the schema, the way the issue puts it, so `Parameter.description` stays empty and Swagger UI shows the hint under the type instead of next to the name. Merging before the parameter is built would fill both. Also, `.. versionchanged::` in the docs says `0.16.0` — tell me if the next release is not that.

## AI Policy

## Checklist

## Related issues

Closes #1441


vyhuholl:
Closing this, since @helo060228 was the first to create a PR closing this issue.

PR #1472, opened 2026-09-12
Raise `EndpointMetadataError` on `as_view` of abstract controllers

### LINKED ISSUE #1445: Ensure that you can include abstract controllers in `Router`
Currently `Router` will say nothing if you try to include `Controller` with `is_abstract == True` into the url patterns.

I propose changing `as_view` logic here: https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/dmr/controller.py#L209-L223

`.as_view()` must be impossible to call on the abstract controller.
Raise `EndpointMetadataError` for this case.

vyhuholl:
# I have made things!

`Controller.as_view` now raises `EndpointMetadataError` when the controller is abstract, before it builds the view. The check is the first thing in `as_view`, where the issue asked for it. `Router` needed no changes, because everything reaches it through `as_view()` already.

Both kinds of abstract controller are covered, and they used to fail in different ways. One with a real serializer and no endpoints answered every request with a `405` and an empty `Allow` header, since `api_endpoints` is `{}` and `dispatch` falls through to `handle_method_not_allowed`. A generic one never gets `api_endpoints` at all, so `dispatch` raised `AttributeError: 'GenericController' object has no attribute 'api_endpoints'` and the request came back as a `500`. Neither of them said anything when the url patterns were built.

One thing in the diff that is not the change itself: `tests/test_unit/test_openapi/test_collector.py` routed an `_EmptyController` through `_process_pattern` to cover the empty path string. That controller cannot be routed anymore, so I dropped it and gave the empty-path case to `_PostController`. `_process_pattern` does not look at endpoints, so it loses no coverage. The new tests are in `tests/test_unit/test_controllers/test_abstract_as_view.py`.

I put the changelog entry under "Breaking changes" and not under features, because code that registers an abstract controller today stops importing after this. It is one line to move if you read it as a fix instead.

The docs get a note in the "Reusable controllers" section of `docs/pages/reusable-code.rst`. That is where reusable controllers are introduced, so it is where someone is most likely to try routing one.

## AI Policy

## Checklist

## Related issues

Closes #1445

### DIFF
--- a/tests/test_unit/test_controllers/test_abstract_as_view.py
+++ b/tests/test_unit/test_controllers/test_abstract_as_view.py
@@ -5,7 +5,6 @@ import pytest
 from dmr import Controller
 from dmr.exceptions import EndpointMetadataError
 from dmr.plugins.pydantic import PydanticSerializer
-from dmr.routing import Router, path
 from dmr.serializer import BaseSerializer
 
 _SerializerT = TypeVar('_SerializerT', bound=BaseSerializer)
@@ -51,13 +50,3 @@ def test_concrete_controller_as_view() -> None:
 
     assert not _Custom.is_abstract
     assert callable(_Custom.as_view())
-
-
-def test_abstract_controller_in_router() -> None:
-    """Ensure that abstract controllers cannot be added to a `Router`."""
-
-    class _Custom(Controller[PydanticSerializer]):
-        """Empty."""
-
-    with pytest.raises(EndpointMetadataError, match='_Custom'):
-        Router(urls=[path('custom/', _Custom.as_view())])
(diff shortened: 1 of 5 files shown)


tests/test_unit/test_controllers/test_abstract_as_view.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def test_abstract_controller_in_router() -> None:
+    """Ensure that abstract controllers cannot be added to a `Router`."""
+
+    class _Custom(Controller[PydanticSerializer]):
+        """Empty."""
+
+    with pytest.raises(EndpointMetadataError, match='_Custom'):
+        Router(urls=[path('custom/', _Custom.as_view())])
```
sobolevn: We don't actually test anything new, please remove this test.

vyhuholl: Removed.

sobolevn (approved):

PR #1473, opened 2026-09-12
Parse `summary` and `description` from controller docstrings

### LINKED ISSUE #1446: Parse `summary` and `description` from `Controller` for `PathItem` docs
Currently we don't parse `summary` and `description` from docstrings for controllers, only for endpoints. This is not good. 

https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/dmr/controller.py#L553-L561

What we can do instead?
- Use `Empty` as the default value for `Controller.summary` and `Controller.description`
- If it is empty, parse `summary` and `description` from the docstring, just like `EndpointMetadata` does
- If it is explicitly `None`, then do nothing
- Test this
- Update all existing snapshots

vyhuholl:
# I have made things!

`Controller.summary` and `Controller.description` now fall back to the controller's docstring, split the way endpoint docstrings already are. The note under "Customizing path items" has promised this since #632, but only endpoints ever did it.

Both attributes default to `EMPTY` instead of `None`, and `Controller._build_summary` and `_build_description` resolve them one at a time. Left alone, the value comes from the docstring. An explicit string wins over it, and an explicit `None` keeps the key out of the schema altogether. The parsing moved out of `EndpointMetadataBuilder._build_description` into `dmr/internal/docstrings.py`, so both callers reach `parse_docstring` through the same helper.

Resolving per attribute rather than all at once is the call I am unsure about. `@modify` goes the other way: pass it either `summary` or `description` and the endpoint docstring is dropped whole. The issue's bullets read per attribute to me, and it buys something real — a controller can take its summary from the docstring and still write `description` by hand, which is what `docs/examples/openapi/path_item_customization.py` now shows. Matching `@modify` is a few lines in those two methods if you would rather have them agree.

Class `__doc__` is not inherited in Python, so a subclass without its own docstring still gets nothing. That is what keeps the controllers we ship out of user schemas. `ObtainTokensSyncController` and the rest carry `Attributes:` blocks and `.. versionchanged::` directives in their docstrings, and their `convert_auth_payload` and `make_api_response` are abstract, so nobody routes to those classes directly.

No existing snapshot moved. The last bullet of the issue expected otherwise, but no controller in the test suite has a docstring, so the only new `.ambr` file is the one for the new tests.

## AI Policy

## Checklist

## Related issues

Closes #1446

### DIFF
--- a/dmr/controller.py
+++ b/dmr/controller.py
@@ -14,7 +14,7 @@ from dmr import throttling as dmr_throttling
 from dmr.cookies import NewCookie
 from dmr.endpoint import Endpoint
 from dmr.errors import ErrorModel, ErrorType, format_error
-from dmr.exceptions import UnsolvableAnnotationsError
+from dmr.exceptions import EndpointMetadataError, UnsolvableAnnotationsError
 from dmr.internal.docstrings import parse_summary_and_description
 from dmr.internal.io import identity
 from dmr.metadata import ResponseSpec
@@ -119,6 +119,8 @@ class Controller(View, Generic[_SerializerT_co]):  # noqa: WPS214
         is_abstract: Whether or not this controller is abstract.
             We consider controller "abstract" when it does not have
             exact serializer type or exact ``api_endpoints`` instances.
+            Abstract controllers cannot be routed, ``as_view`` raises
+            :class:`~dmr.exceptions.EndpointMetadataError` for them.
         is_async: Whether or not this controller is async.
         streaming: Does this controller work with streaming responses like SSE?
         controller_validator_cls: Runs full controller validation on definition.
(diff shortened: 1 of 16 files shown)


dmr/controller.py
```diff
@@ -573,6 +582,32 @@ def view_is_async(cls) -> bool:  # noqa: N805  # pyright: ignore[reportIncompati
 
     # Protected API:
 
+    @classmethod
+    def _build_summary(cls) -> str | None:
```
sobolevn: I would prefer not to have this API on the controller. Can we instead move it to `internal/docstrings.py`? Moreover, they are almost identical.

docs/examples/openapi/path_item_customization.py
```diff
@@ -4,7 +4,9 @@
 
 
 class UserController(Controller[MsgspecSerializer]):
-    description = 'Create new users'
+    """Users API."""  # This docstring becomes the `summary`
```
sobolevn:
```suggestion
    """
    Users API.
    
    Replaced description.
    """  # This docstring becomes the `summary`, but not `description`
```

sobolevn (commented):
Thanks, two comments.

sobolevn:
And please fix the merge conflict.

dmr/controller.py
vyhuholl: Combined into one function, moved to `internal/docstrings.py`

dmr/internal/docstrings.py
```diff
@@ -0,0 +1,61 @@
...
+from django.contrib.admindocs.utils import parse_docstring
+from typing_extensions import Sentinel, TypedDict
+
+if TYPE_CHECKING:
+    from django.utils.functional import (
+        _StrOrPromise,  # pyright: ignore[reportPrivateUsage]
+    )
+
+
+class SummaryAndDescription(TypedDict, closed=True):
```
sobolevn:
```suggestion
@final
class _SummaryAndDescription(TypedDict, closed=True):
```

dmr/validation/endpoint_metadata.py
sobolevn: This code now is out of sync with the controller code. It skips all the parsing if either `summary` or `description` is set. Let's unify the logic.

vyhuholl: Done in https://github.com/wemake-services/django-modern-rest/commit/00f9f0f0f168ecddfc89974c6c55e3364deb6459

sobolevn (approved):
LGTM now! Thanks!

PR #1474, opened 2026-09-12
Run jwt auth examples in the docs

### LINKED ISSUE #1450: Create pair of JWT tokens for the docs
Currently we can't run examples that involve JWT auth, but we have a lot of such examples. So, we need to create a pair of JWT tokens and be able to use them in `# run` comments the same way we use / run examples with `Token` auth.

Examples: 
- Auth: https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/docs/examples/auth/token/revoke_token.py#L11-L23
- https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/docs/tools/sphinx_ext/run_examples.py#L951-L952

Btw, also remove `Token` support from `Query`, since we don't use it anymore: https://github.com/wemake-services/django-modern-rest/blob/87a02f59fd96a5d9e2da0d1e80d35c84af248e05/docs/tools/sphinx_ext/run_examples.py#L840-L844

vyhuholl:
# I have made things!

`$JWT_ACCESS_TOKEN` and `$JWT_REFRESH_TOKEN`, two templates that `# run:` comments use the way they already use `$X_API_TOKEN`. We had one for the opaque token and nothing for jwt, so none of the examples behind jwt auth could run.

Unlike an opaque token, a jwt is not stored anywhere. It only has to be signed with our secret and to point at a user that exists. So `_IssuedJWTokens` writes no file: `_StoredToken` needs one because `Token.issue` returns the raw string once and the database keeps a hash, and that constraint does not apply here. It signs one pair for `test_user` the first time an example asks and reuses it for the rest of the build. That is the part I am least sure about. The issue says to use these the way we use `Token` auth, and the `# run:` side is identical, but the storage side is deliberately not. Say so and I will write the pair to `_build/` next to `token.txt`.

`_add_headers` compared `header_value == _TOKEN_TEMPLATE`, and `_add_cookies` had the same line. It is now one `_substitute_tokens` that walks a value and replaces any template it finds, driven by a `{template: loader}` map. That is what makes request bodies work, and two examples need it: `jwt_refresh_tokens.py` and `jwt_verify_tokens.py` carry the token in the json body. `$X_API_TOKEN` is gone from `"query"`, as the issue asks. No example passed a token there.

One thing in the diff that is not the issue: `SECRET_KEY`. It was 22 bytes, and `pyjwt` refuses to decode an `HS256` token signed with fewer than 32. That would not even have surfaced as a `401`. `InvalidKeyError` does not inherit from `InvalidTokenError`, so the `except` in `JWToken.decode` misses it and the example gets a `500`. It is 42 bytes now. `'test_user'` became `_TEST_USERNAME` in the same file, since the issuer looks the user up by it.

Five examples got a `# run:`. `using_jwt.py` and `using_jwt_cookie.py` both answer `"authed"`, `blocklist_tokens.py` goes through the blocklist mixin, `jwt_refresh_tokens.py` comes back with a fresh pair, and `jwt_verify_tokens.py` returns `204` with no body, so it passes `-D -` to have something to print. The cookie controllers still do not run, for a reason that has nothing to do with tokens: `jwt_cookie_refresh.py`, `jwt_cookie_logout.py`, `jwt_cookie_logout_blocklist.py`, `blocklist_refresh.py`, and `jwt_cookie_obtain_body.py` all set `jwt_refresh_cookie_path = reverse_lazy('api:jwt_refresh')`. The single-controller urlconf from `_generate_urls` has no route under that name. Giving the runner named and namespaced routes is a bigger change than this issue asks for, so I left them.

Last, something you may hit while reviewing rather than something I changed. Rebuild the docs without cleaning and `using_token_header.py` answers `401`: `revoke_token.py` revokes the one stored token, so after the first build the row in the kept `_build/test.db` stays revoked. `just docs` cleans first, so CI never sees it.

## AI Policy

## Checklist

## Related issues

Closes #1450

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@ https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
 
 ## 0.16.0 WIP
 
+### Breaking changes
+
+- `OpenAPIConfig` now raises `ValueError` for `openapi_version` below `3.1.0`.
+  OpenAPI `3.0.x` was never really supported: it predates JSON Schema,
+  which is what `pydantic` and `msgspec` generate for our models, #1435
+
 ### Features
 
 - Added `tags` controller attribute to apply OpenAPI tags
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@
 - [x] Fully typed and checked with `mypy`, `pyright`, `ty`, and `pyrefly` in strict modes
 - [x] Supports content negotiation, has default implementations for `json`, `msgpack`, SSE, Json Lines, and more
 - [x] Strict schema validation of both requests and responses, including errors
-- [x] Supports OpenAPI 3.0 / 3.1 / 3.2 semantic schema generation out of the box
+- [x] Supports OpenAPI 3.1 / 3.2 semantic schema generation out of the box
 - [x] Supports all your existing `django` primitives and packages, no custom runtimes
 - [x] Great testing tools with [schemathesis](https://github.com/schemathesis/schemathesis), [polyfactory](https://github.com/litestar-org/polyfactory), [tracecov](https://django-modern-rest.readthedocs.io/en/latest/pages/testing/tracecov.html), bundled `pytest` plugin, and default Django's testing primitives
 - [x] 100% test coverage with 3000+ of carefully designed unit, integration, typing, and property-based tests
--- a/dmr/openapi/config.py
+++ b/dmr/openapi/config.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import cast
+from typing import Final, cast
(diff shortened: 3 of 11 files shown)


sobolevn (commented):
Looks like we missed quite a lot of files that we can run:

Image

Shouldn't we run them as well? :)

vyhuholl:
Added the rest of jwt examples in https://github.com/wemake-services/django-modern-rest/commit/2b949736ae6ed5f357f53c76e9fcfa7f2c207b4d

sobolevn (approved):
Nice! 🎉

PR #1477, opened 2026-09-12
Remove OpenAPI 3.0.x support

### LINKED ISSUE #1435: Remove OpenAPI 3.0.x support

Vir-007:
We claimed support for OpenAPI 3.0.x, but only ever supported a subset of it, since 3.1 was the first version to adopt JSON Schema for models.

`OpenAPIConfig.__post_init__` now raises `ValueError` when `openapi_version` is below `3.1.0`. Updated the README, `docs/pages/openapi/openapi.rst`, and the `openapi.yml` external-views example to stop claiming 3.0.x support, and added a CHANGELOG entry.

Tested with two new unit tests in `tests/test_unit/test_openapi/test_spec.py`: one confirming `openapi_version='3.0.3'` raises, one confirming `3.1.0`/`3.2.0` are still accepted. Ran `just lint`, `just type-check`, and the full unit suite locally, all green.

Fixes #1435

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,15 @@ are stored as descriptions in version releases on GitHub, example:
 https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
 
 
+## 0.16.0 WIP
+
+### Breaking changes
+
+- `OpenAPIConfig` now rejects `openapi_version` values before `3.1.0`.
+  We only ever supported a subset of `3.0.x`, since `3.1` was the first
+  version to adopt JSON Schema for models, #1435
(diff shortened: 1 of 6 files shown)


sobolevn (commented):
This was fixed in https://github.com/wemake-services/django-modern-rest/pull/1444

Vir-007:
Ah, missed that — thanks for pointing it out!

PR #1478, opened 2026-09-12
Support better schema generation parameters

### LINKED ISSUE #1462: Support better schema generation parameters

hxxxi-malog:
# I have made things!

## AI Policy

## Checklist

## Related issues

- Closes #1462

## What I did

Both schema generators can now be customized, as requested in #1462:

- `PydanticSchemaGenerator` got a new `json_schema_kwargs` attribute. It is a closed `TypedDict` that forwards extra keyword arguments to pydantic's `TypeAdapter.json_schema()`: `by_alias`, `union_format`, and `schema_generator`. `ref_template` and `mode` are left out on purpose, they are always computed by `dmr` itself.
- `MsgspecSchemaGenerator` got a new `schema_hook` attribute. It is passed to `msgspec.json.schema()` and is called for each custom type that `msgspec` cannot describe natively.

The customization path is subclassing the plugin's schema generator and creating your own serializer that uses it.

## Notes

- Defaults are unchanged: `json_schema_kwargs` is `{}` (a no-op on expand) and `schema_hook` is `None`, which is `msgspec.json.schema`'s own default.
- Docs: new "Customizing schema generator" section in the OpenAPI page with two runnable examples, short sections in the plugins page, and `JsonSchemaKwargs` in the public API reference.
- The docs also warn that generated schemas are cached per annotation and are shared between all serializers of the same API, and that `msgspec` custom types still need a `serialize_hook` override to be serialized at runtime (the example shows both sides).
- Tests cover `schema_generator`, `by_alias` (with a default comparison), `union_format='primitive_type_array'`, and `schema_hook` (including the fallback error for unsupported types).
- CHANGELOG entries are added under `0.16.0 WIP`.


docs/examples/openapi/msgspec_schema_hook.py
```diff
@@ -0,0 +1,58 @@
...
+
+class Point:  # noqa: B903
+    """Custom type that ``msgspec`` cannot describe natively."""
+
+    def __init__(self, x: float, y: float) -> None:
+        self.x = x
+        self.y = y
+
+
+def point_schema(type_: type) -> dict[str, Any]:
```
sobolevn:
```suggestion
def point_schema(type_: type[Any]) -> dict[str, Any]:
```

docs/examples/openapi/msgspec_schema_hook.py
```diff
@@ -0,0 +1,58 @@
...
+            'required': ['x', 'y'],
+        }
+    raise NotImplementedError(type_)
+
+
+class SchemaGenerator(MsgspecSchemaGenerator):
+    schema_hook: ClassVar[SchemaHook | None] = point_schema
+
+
+class PointSerializer(MsgspecSerializer):
```
sobolevn:
```suggestion
class PointMsgspecSerializer(MsgspecSerializer):
```

docs/examples/openapi/msgspec_schema_hook.py
```diff
@@ -0,0 +1,58 @@
...
+class Polygon(msgspec.Struct):
+    points: list[Point]
+
+
+class PolygonController(Controller[PointSerializer]):
+    def get(self) -> Polygon:
+        return Polygon(points=[Point(1, 2), Point(3, 4)])
+
+
+# openapi: {"controller": "PolygonController", "openapi_url": "/docs/openapi.json/"}  # noqa: ERA001, E501
```
sobolevn: please add `# run` comment as well. we need to see that it works as it should.

docs/examples/openapi/msgspec_schema_hook.py
```diff
@@ -0,0 +1,58 @@
...
+        """Serialize ``Point`` objects, ``msgspec`` cannot do it natively."""
+        if isinstance(to_serialize, Point):
+            return {'x': to_serialize.x, 'y': to_serialize.y}
+        return super().serialize_hook(to_serialize)
+
+
+class Polygon(msgspec.Struct):
+    points: list[Point]
+
+
```
sobolevn:
```suggestion
```

docs/examples/openapi/msgspec_schema_hook.py
```diff
@@ -0,0 +1,58 @@
...
+class PolygonController(Controller[PointSerializer]):
+    def get(self) -> Polygon:
+        return Polygon(points=[Point(1, 2), Point(3, 4)])
```
sobolevn:
```suggestion
class PointsController(Controller[PointSerializer]):
    def get(self) -> list[Point]:
        return [Point(1, 2), Point(3, 4)]
```

let's not overly complicate this controller with extra types :)

docs/examples/openapi/pydantic_schema_generator.py
```diff
@@ -0,0 +1,50 @@
...
+
+from dmr import Controller
+from dmr.plugins.pydantic import PydanticSerializer
+from dmr.plugins.pydantic.schema import (
+    JsonSchemaKwargs,
+    PydanticSchemaGenerator,
+)
+
+
+class NoTitleJsonSchema(GenerateJsonSchema):
```
sobolevn: Please, use the same exact example with `Point` from `msgspec`. These two examples must be semantically the same, but their _implementation_ must be different.

docs/examples/openapi/pydantic_schema_generator.py
```diff
@@ -0,0 +1,50 @@
...
+class UserModel(pydantic.BaseModel):
+    email: str
+
+
+class UserController(Controller[NoTitleSerializer]):
+    def get(self) -> UserModel:
+        return UserModel(email='user@example.com')
+
+
+# openapi: {"controller": "UserController", "openapi_url": "/docs/openapi.json/"}  # noqa: ERA001, E501
```
sobolevn: add `# run` here as well

sobolevn (commented):
Nice! Thanks! Several comments

docs/examples/openapi/pydantic_schema_generator.py
hxxxi-malog: Okay, I’ll fix these issues and then resubmit.

docs/examples/openapi/pydantic_schema_generator.py
```diff
@@ -22,10 +23,26 @@ def generate(
...
     ) -> dict[str, Any]:
-        """Generate a JSON schema and remove its title."""
+        """Generate a JSON schema and remove all titles from it."""
         json_schema = super().generate(schema, mode=mode)
-        json_schema.pop('title', None)
-        return json_schema
+        return _drop_titles(json_schema)
+
+
+def _drop_titles(json_schema: dict[str, Any]) -> dict[str, Any]:
```
sobolevn: nope, this code is too complex for the example. we need something simple:
- it must work the same way for `msgspec` and `pydantic`
- implementation must be simple for both serializers
- it must be declarative. just like current `msgspec` implementation: here's the type, here's its schema

setup.cfg
```diff
@@ -91,5 +91,8 @@ per-file-ignores =
   docs/tools/sphinx_ext/*.py: WPS201, WPS202, WPS203, WPS204, WPS226, WPS402
   # Tests in doc examples are special:
   docs/examples/testing/*.py: WPS226
+  # Axis names are short and repeated in schema hooks by design:
+  docs/examples/openapi/msgspec_schema_hook.py: WPS111, WPS226
+  docs/examples/openapi/pydantic_schema_generator.py: WPS111
```
sobolevn: it is better to always follow the rules, use `coord_x` and `coord_y`

hxxxi-malog: Okay, I found where the problem is. My earlier understanding was off, and I’ll now fix it again.

hxxxi-malog: In the tests, WPS226 still triggers: the string `coord_x` appears 4 times in `point_schema`, `serialize_hook`, and `deserialize_hook`, exceeding the threshold of 3. This one still needs to remain exempted.

docs/examples/openapi/pydantic_schema_generator.py
hxxxi-malog: `NoTitleJsonSchema` was changed from recursive post-processing to two simple overrides, removing the two helper functions `_drop_titles` and `_iter_nested`:
- `field_title_should_be_set` returns `False`, a declarative switch: do not generate field titles
- `generate` has two `pop` calls, handling only the two fixed locations, the top level and `$defs`, with a simple loop and no recursion

The annotation uses `schema: Any` because `CoreSchemaOrField` only exists in the private `pydantic._internal` module

hxxxi-malog:
Let me think about it for a bit longer...

hxxxi-malog:
Thanks for the review! All comments are addressed:

- initial round: `type[Any]` annotation, renames, `Polygon` removed in favor of `PointsController` returning `list[Point]`, `# run` comments added
- `coord_x` / `coord_y` renames, lint rules followed
- both examples are now declarative and symmetric: msgspec `schema_hook` vs pydantic `model_schema` override, both literally "here's the type, here's its schema"

Verified locally: both examples run (`200` responses), their OpenAPI outputs are semantically identical, and `flake8` / `ruff` / `pyrefly` / `ty` are all green with no new per-file ignores.

docs/examples/openapi/msgspec_schema_hook.py
```diff
@@ -0,0 +1,71 @@
...
+                'coord_x': {'type': 'number'},
+                'coord_y': {'type': 'number'},
+            },
+            'required': ['coord_x', 'coord_y'],
+        }
+    raise NotImplementedError(type_)
+
+
+class SchemaGenerator(MsgspecSchemaGenerator):
+    schema_hook: ClassVar[SchemaHook | None] = point_schema
```
sobolevn: let's just define `@classmethod schema_hook` and copy `point_schema` logic there

tests/test_unit/test_plugins/test_msgspec/test_msgspec_schema.py
```diff
@@ -467,3 +477,45 @@ def test_unsupported_type(schema_generator: SchemaGenerator) -> None:
...
+class _HookedSerializer(MsgspecSerializer):
+    schema_generator = _HookedSchemaGenerator
+
+
+def test_custom_schema_hook(schema_generator: SchemaGenerator) -> None:
+    """Ensure custom ``schema_hook`` option is respected."""
+    schema = schema_generator(_CustomType, _HookedSerializer)
+
+    assert isinstance(schema, Schema)
+    assert schema.type == OpenAPIType.STRING
```
sobolevn: assert the full object:
```suggestion
    assert schema == Schema(type=OpenAPIType.STRING)
```

tests/test_unit/test_plugins/test_pydantic/test_pydantic_schema.py
```diff
@@ -505,3 +510,115 @@ def test_unsupported_type(schema_generator: SchemaGenerator) -> None:
         match='Cannot generate OpenAPI schema',
     ):
         schema_generator(_TestClass, PydanticSerializer)
+
+
+class _NoTitleJsonSchema(GenerateJsonSchema):
```
sobolevn: please, use the same test as `msgspec` test does. this one is too complex and does a lot of unrelated things

tests/test_unit/test_plugins/test_pydantic/test_pydantic_schema.py
```diff
@@ -505,3 +510,115 @@ def test_unsupported_type(schema_generator: SchemaGenerator) -> None:
...
+    field_name: str = pydantic.Field(alias='fieldAlias')
+
+
+@final
+class _NoAliasSchemaGenerator(PydanticSchemaGenerator):
+    json_schema_kwargs: ClassVar[JsonSchemaKwargs] = {'by_alias': False}
+
+
+@final
+class _NoAliasSerializer(PydanticSerializer):
```
sobolevn: test both `PydanticSerializer` and `PydanticFastSerializer`, parametrize this inside a test. both should be covered for green and red paths.

hxxxi-malog: OK, I'll fix them.

PR #1479, opened 2026-09-12
Pinned schemathesis version.

s3rius:
# Sup, another update!

Schemathesis [v4.27.0](https://github.com/schemathesis/schemathesis/releases/tag/v4.27.0) introduced new integer tests, according to OpenAPI3.1 specification.

> Integer fields in JSON bodies draw float spellings (2.0) under Open API 3.1 and 3.2.

According to [OpenAPI3.1 specs](https://swagger.io/specification/v3.1/), that should be true for all integers

> Since there is no distinct JSON integer type, JSON Schema defines integers mathematically. This means that both 1 and 1.0 are [equivalent](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-4.2.2), and are both considered to be integers.

However, pydantic in `strict` mode, doesn't comply with such specification. And reject such values. I made an issue about that. We can upgrade schemathesis when this issues is resolved. 

https://github.com/pydantic/pydantic/issues/13799

As a solution for now: 
* Revert my previous changes;
* Pin schemathesis version to 4.26.1 (was working correctly).

## AI Policy

## Checklist


sobolevn (approved):

sobolevn:
Oh, this is complex, I am not really sure that this is a pydantic bug. 
But, this is fine to fix the CI for now.
What does @Stranger6667 recommend in this case?

s3rius:
@sobolevn, since Pydantic tries to be as close as possible to the OpenAPI schema (since it generates it), it should comply with its specification. Otherwise, the generated schema doesn't reflect the actual parsing logic, which will always be incorrect.

Stranger6667:
@sobolevn I agree with @s3rius - pydantic should follow the spec. It is also was my intention with that change in schemathesis 4.27 without a knob to disable - to find more cases where something doesn't comply with the spec. That being said pinning should work :)

sobolevn:
Thanks! The only risk I can see here is if `pydantic` team would just close the issue now as wontfix.

@s3rius btw, did you test `msgspec` for the same thing? Maybe we should fix it as well.

PR #1481, opened 2026-09-12
Refactor `endpoint_metadata.py` file

MasoNord:
# I have made things!
Removed redundant helper functions such as `_check_modification_cookie_syntax()`, `_check_response_header_syntax()`, etc. This reduces boilerplate code in the core

## AI Policy

## Checklist

## Related issues
- Refs #1341


sobolevn (commented):
There are still multiple similar lines:

```python
if (
    HttpSpec.empty_response_body
    not in self.metadata.no_validate_http_spec
):
    self._check_empty_response_body(responses)
if (
    HttpSpec.header_name_server_managed
    not in self.metadata.no_validate_http_spec
):
    self._check_header_name_server_managed(responses)
```

is a full duplicate.

Also, these methods also are:

```python
def _check_header_syntax(
        self,
        responses: list[ResponseSpec],
    ) -> None:

        header_names = []

        modification = self.metadata.modification

        if modification and modification.headers:
            header_names = list(modification.headers.keys())

        for response in responses:
            if not response.headers:
                continue

            header_names += list(response.headers.keys())

        invalid_header = self._check_invalid_tokens(header_names)

        if invalid_header:
            raise EndpointMetadataError(
                f'Header name {invalid_header!r} is not following http spec.',
            )
```

and 

```python
def _check_cookie_syntax(
        self,
        responses: list[ResponseSpec],
    ) -> None:
        cookie_names = []

        modification = self.metadata.modification

        if modification and modification.cookies:
            cookie_names = list(modification.cookies.keys())

        for response in responses:
            if not response.cookies:
                continue

            cookie_names += list(response.cookies.keys())

        invalid_cookie = self._check_invalid_tokens(cookie_names)
        if invalid_cookie:
            raise EndpointMetadataError(
                f'Cookie name {invalid_cookie!r} is not following http spec.',
            )
```

they are different in very small details.

maybe we can create a new class for httpspec validation? and abstract these checks into it? and remove all code duplication :)

MasoNord:
About the first take, I'm a little bit confused. I don't understand why they are considered full duplicates, since `_check_empty_response_body` and `_check_header_name_server_managed` do different things. Or did you mean that they are full duplicates because of the number of `if` statements?

About the second one, you're right. We can combine `_check_cookie_syntax` and `_check_header_syntax` into a separate HTTP spec validation class, since they basically do the same thing: check whether names contain valid tokens. Should we call it `http_name_syntax`?

sobolevn:
1. Because they are the same:

```python
if (
    HttpSpec.CHECK_NAME
    not in self.metadata.no_validate_http_spec
):
    self.CHECK_CALLBACK(responses)
```

2. Create an abstraction that accepts callbacks and attributes, it must be a single method for all checks ideally :) or maybe two methods total

MasoNord:
If I understood you correctly, the proposed approach would be the following:

1. Create a separate helper function that handles the `no_validate_http_spec` check and invokes the corresponding callback:

```python
def _validate_http_spec(
    self,
    responses: list[ResponseSpec],
) -> None:
    self._check_http_spec_rule(
        HttpSpec.empty_response_body,
        self._check_empty_response_body,
        responses=responses,
    )

    self._check_http_spec_rule(
        HttpSpec.header_name_server_managed,
        self._check_header_name_server_managed,
        responses=responses,
    )
def _check_http_spec_rule(
    self,
    rule: HttpSpec,
    callback: Callable[..., None],
    **kwargs
) -> None:
    if rule not in self.metadata.no_validate_http_spec:
        callback(kwargs)
```

2. Combine the cookie and header syntax validation into a single `_check_http_syntax()` function:

```python
def _check_http_syntax(
    self,
    responses: list[ResponseSpec],
    attribute: Literal["cookie", "header"],
) -> None:
    names = []

    modification = self.metadata.modification

    if modification:
        if attribute == "header" and modification.headers:
            names = list(modification.headers.keys())
        elif attribute == "cookie" and modification.cookies:
            names = list(modification.cookies.keys())

    for response in responses:
        if attribute == "header" and response.headers:
            names += list(response.headers.keys())
        elif attribute == "cookie" and response.cookies:
            names += list(response.cookies.keys())

    invalid_name = self._check_invalid_tokens(names)

    if invalid_name:
        raise EndpointMetadataError(
            f"{attribute.capitalize()} name {invalid_name!r} "
            "is not following the HTTP spec.",
        )
```
Since this approach combines two checks in one, should we introduce new `HttpSpec` rule, like `http_name_syntax` ?

Does this match what you had in mind?

sobolevn:
yes, seems like something that I was talking about :)

dmr/validation/endpoint_metadata.py
```diff
@@ -178,33 +178,43 @@ def _validate_http_spec(
...
+            responses=responses,
+            field_type='header',
+        )
 
         # TODO: add more checks
 
+    def _check_http_spec_rule(
+        self,
+        rule: HttpSpec,
+        callback: Callable[..., None],
```
sobolevn, line 212: Please, use `ParamSpec` and `*args`, `**kwargs` to fully type the callback and its args.

dmr/validation/endpoint_metadata.py
```diff
@@ -178,33 +178,43 @@ def _validate_http_spec(
...
 
-        if (
-            HttpSpec.header_name_syntax
-            not in self.metadata.no_validate_http_spec
-        ):
-            self._check_modification_header_syntax()
-            self._check_responses_header_syntax(responses)
+        self._check_http_spec_rule(
+            HttpSpec.header_name_server_managed,
+            self._check_header_name_server_managed,
```
sobolevn, line 189: I would recommend creating a new class with these callbacks. It would be much cleaner.

dmr/settings.py
```diff
@@ -98,18 +98,15 @@ class HttpSpec(enum.StrEnum):
...
+            field names conform to the token syntax
+            defined in RFC 9110, Section 5.6.2.
     """
 
     empty_request_body = 'empty_request_body'
     empty_response_body = 'empty_response_body'
     header_name_server_managed = 'header_name_server_managed'
-    header_name_syntax = 'header_name_syntax'
-    cookie_name_syntax = 'cookie_name_syntax'
+    http_field_name_validation = 'http_field_name_validation'
```
sobolevn, line 109: Please, revert this change. Let it be two different fields.

sobolevn (commented):
Thanks! We are getting close to a good design. 

Don't forget to use `ClassVar[type[YourClass]]` to make the class customizable.

dmr/validation/endpoint_metadata.py
MasoNord: Do you mean extracting the HTTP spec rule/callback pairs into a separate class, so that `HttpSpecValidator` only orchestrates validation?

sobolevn: Yes :)

sobolevn (approved):
amazing! thanks!

sobolevn:
We can now continue adding new rules :)

PR #1492, opened 2026-09-13
Reject conflicting security scheme names in SecuritySchemeRegistry

### LINKED ISSUE #1488: Make sure that `SecuritySchemeRegistry` can't re-register existing auth schemes
Currently `SecuritySchemeRegistry` does not protect from registerting two auth schemes with the same name: https://github.com/wemake-services/django-modern-rest/blob/f822e35bd2db672dcc43fbd4967a22835a616971/dmr/openapi/core/registry.py#L127-L142

This needs to be fixed. It must raise an error, no error must pass silently.

HarshRajSinghania:
Fixes #1488

## Summary
`SecuritySchemeRegistry.register` now raises `ValueError` when the same name is registered with a different `SecurityScheme` or `Reference`. Silent overwrite of a conflicting scheme is no longer possible.

## Motivation
#1488 pointed out that two auth schemes could be registered under one name with no error. That produced an incomplete OpenAPI document instead of failing loudly.

## Implementation
- If the name is new, register as before.
- If the name already exists and the scheme is equal, keep the existing entry (idempotent). The generator calls `register` once per endpoint, so the same auth provider on multiple operations must remain valid.
- If the name already exists and the scheme differs, raise `ValueError` with the scheme name in the message, matching `OperationIdRegistry`.

## Testing
Added `tests/test_unit/test_openapi/test_core/test_security_scheme_registry.py`:
- unique names succeed
- equal scheme re-register is a no-op
- different `SecurityScheme` under one name raises
- different `Reference` under one name raises

Could not run the project's full pytest suite in this environment (`pip`/`hatch` install failed on a 502 from the package index). The registry logic was exercised with a standalone import of `SecuritySchemeRegistry` against the new tests' cases.


tests/test_unit/test_openapi/test_core/test_security_scheme_registry.py
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    for name, scheme in schemes.items():
+        registry.register(name, scheme)
+
+    assert registry.schemes == schemes
+
+
+def test_duplicate_security_scheme_raises_error() -> None:
+    """Test that registering a duplicate security scheme raises ``ValueError``."""
```
sobolevn:
```suggestion
    """Test that registering a duplicate security scheme raises `ValueError`."""
```

PR #1498, opened 2026-09-13
fix: no empty `description` in merged `requestBody`

### LINKED ISSUE #1495: It is possible to produce an empty `description` for `ResponseBody`
https://github.com/wemake-services/django-modern-rest/blob/f822e35bd2db672dcc43fbd4967a22835a616971/dmr/openapi/generators/component_parsers.py#L193-L199

It misses `or None` part. Because we only skip `None` values from dumping.

larryxs:
Closes #1495

# Made some fixes on `requestBody` scheme

## Checklist


dmr/openapi/generators/component_parsers.py
```diff
@@ -196,7 +196,8 @@ def _merge_bodies(
                 (schema.description or '')
                 + ' '
                 + (new_schema.description or '')
-            ).strip(),
+            ).strip()
+            or None,
```
sobolevn:
```suggestion
            description=(
                (
                    (schema.description or '')
                    + ' '
                    + (new_schema.description or '')
                ).strip() or None
            ),
```

sobolevn (commented):
nit

PR #1508, opened 2026-09-13
Fix OpenAPI security for auth-disabled endpoints

### LINKED ISSUE #1497: Test explicit `OpenAPIConfig.security` definition with `auth=None` per-endpoint
Looks like we have a potential bug in how we define `security` in OpenAPI.
Here we always return `None`: https://github.com/wemake-services/django-modern-rest/blob/f822e35bd2db672dcc43fbd4967a22835a616971/dmr/openapi/generators/security_scheme.py#L24-L37

But, if per-document `security` is set, we need to return `[]` as the value. Otherwise, `None` won't be dumped by our `None`-ignoring `dump_schema` rules. And `security` rules per this endpoint will be inherited from the document.

Which is wrong. 

This would require a test and a fix. Maybe we should look at current config's `security` to define the return type?

Deepak-raja-123:
## What does this PR do?

Fixes #1497.

When global OpenAPI security is configured, endpoints with
`auth=None` now explicitly generate an empty `security` requirement
instead of omitting the field.

This prevents the endpoint from inheriting the document-level
security requirement.

## Changes

- Return `[]` from `SecuritySchemeGenerator` when `auth_providers`
  is empty and global security is configured.
- Add a regression test for the security scheme generator.
- Add a schema-level regression test for `auth=None`.
- Update the changelog.

## Tests

- `uv run pytest tests\test_unit\test_openapi` — 232 passed
- Ruff check — passed
- Ruff format check — passed
- Flake8 — passed
- Slotscheck — passed
- Import-linter — passed

The full test suite was also attempted locally, but unrelated
Allauth snapshot failures and Redis/Docker integration errors
prevented a completely green full-suite run.


tests/test_unit/test_openapi/test_generators/test_security_scheme.py
```diff
@@ -99,3 +99,19 @@ def test_security_scheme_generator_with_schemes(
     assert openapi_context.registries.security_scheme.schemes == snapshot({
         'testScheme': SecurityScheme(type='http', scheme='bearer'),
     })
+
+
+def test_no_auth_with_global_security() -> None:
```
sobolevn: This test duplicates `test_auth_none_with_global_security`, please remove it

tests/test_unit/test_openapi/test_schema_snapshots.py
```diff
@@ -132,6 +132,36 @@ def test_auth_and_cookies_schema(snapshot: SnapshotAssertion) -> None:
...
+                path(
+                    'public/',
+                    _AuthDisabledController.as_view(),
+                ),
+            ],
+        ),
+        config=config,
+    ).convert()
+
+    assert schema['paths']['/api/public/']['get']['security'] == []
```
sobolevn: we assert the full snapshot, not just a single part.
please, add the proper check.

CHANGELOG.md
```diff
@@ -64,26 +64,10 @@ https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
   it is copied from the sub-pattern of the matching named group:
   `r'^v(?P<version>\d+)/$'` documents `version`
   as `{'type': 'string', 'pattern': '^(?:\d+)$'}`, #1439
-- Added `json_schema_kwargs` attribute to `PydanticSchemaGenerator`
-  to pass extra keyword arguments like `by_alias`, `union_format`,
-  and `schema_generator` to `pydantic`'s `TypeAdapter.json_schema`, #1462
-- Added `schema_hook` class method to `MsgspecSchemaGenerator`
-  to customize JSON schema generation for custom types, #1462
 
 ### Bugfixes
```
sobolevn, line 88: Please, fix the merge conflict.

Deepak-raja-123:
Addressed the review feedback:
- Removed the duplicate test.
- Changed the endpoint test to assert the full OpenAPI snapshot.
- Merged the latest master to resolve the changelog conflict.

PR #1510, opened 2026-09-13
Add tests against different dbs.

s3rius:
### Tests against different DBs

previously we were only using sqlite for testing the project. But in order to be extra sure, we need to test against real popular databases. 

This PR adds tests against `postgresql` and `mysql` using testcontainers. SQLite is still in use, and runs by default.


.github/workflows/test.yml
```diff
@@ -22,10 +22,14 @@ jobs:
       matrix:
         python-version: ['3.11', '3.12', '3.13', '3.14']
         redis-image: ['redis:8-alpine']
+        database-image:
```
sobolevn: my idea is not to duplicate anything in the CI. it take quite a long time. so, we can run:

- py311 / sqlite / redis:7
- py312 / oracle / redis:8
- py313 / postgres / valkey
- py314 / mysql / ...

This way we still test all, but not all combitations. Because we don't actually use anything super specific.

s3rius: I think that at least one version should be consistently tested against all databases. This would be easier to pinpoint the issue in case if it's python specific.

sobolevn (approved):
Looks solid!

sobolevn (approved):

Issue #1494, opened 2026-09-13
Test and document that default values for components are supported


sobolevn:
Currently we don't ever test or document that a code like this:

```python
from dmr import Body, Controller
from dmr.plugins.pydantic import PydanticSerializer


class MyController(Controller[PydanticSerializer]):
    def post(
        self,
        parsed_body: Body[dict[str, str] | None] = None,
    ) -> str:
        print(parsed_body)
        return 'done'
```

is even supported.

What is important here?
1. We must call the endpoint function as-is, without any work from our part. Defaults must be real python defaults. And work the same way
2. But, if there are default for `Body`, we must issue `required: false` schema for the whole body

Snapshot test is required for this change.

soil0119:
Hi! I'd like to work on this issue.

I'm interested in Python/Django and typed API frameworks, especially how
request validation and OpenAPI schema generation work internally. I think
this issue would be a great opportunity to learn more about those areas
while contributing to the project.

My understanding is that we need to:
1. add a regression test showing that an endpoint's Python default is used
   when the request body is omitted,
2. mark the request body as optional in the generated OpenAPI schema, and
3. document the behavior.

Could you confirm whether the scope is limited to the Body component, and
whether the tests should cover both Pydantic and msgspec?

If the scope is correct and the issue is still available, I'd be happy to
work on it. Thank you!

sobolevn:
@soil0119 sorry, I've missed your message. Assigned! :)

PR #1511, opened 2026-09-14
fix: require all parsers to support file parsing in FileMetadata

### LINKED ISSUE #1493: Change how we work with `FileMetadata` validation
Currently, this code is allowed:

```python
from typing import Literal

import pydantic

from dmr import Body, Controller, FileMetadata
from dmr.plugins.pydantic import PydanticSerializer
from dmr.parsers import JsonParser, MultiPartParser


class Text(pydantic.BaseModel):
    content_type: Literal['text/plain']


class Files(pydantic.BaseModel):
    text: Text


    parsers = [JsonParser(), MultiPartParser()]

    def post(
        self,
        parsed_body: Body[dict[str, str]],
        parsed_file_metadata: FileMetadata[Files],
    ) -> str:
        return 'done'
```

But, how can we ever satisfy?
- `JsonParser()`
- `parsed_file_metadata: FileMetadata[Files]`

It is not optional. Every request here will fail.
What we can do? Change https://github.com/wemake-services/django-modern-rest/blob/f822e35bd2db672dcc43fbd4967a22835a616971/dmr/components.py#L866-L889 from `any` to `all`.

jarik2014:
## What this does

`FileMetadata` now requires **all** declared parsers to implement `SupportsFileParsing`
instead of at least one, as requested in #1493.

## Why

`FileMetadata` is not optional in the endpoint signature, so any parser that cannot parse
files makes the endpoint permanently broken: the request is negotiated to that parser and
guaranteed to fail at runtime. With the previous `any(...)` check this controller passed
import-time validation:

```python
    parsers = (MultiPartParser(), JsonParser())

    def post(
        self,
        parsed_file_metadata: FileMetadata[Files],
    ) -> str:
        ...
```

but every request negotiated to `JsonParser` returned a 400.
Now this is rejected at import time:

```
EndpointMetadataError: Class <class 'MyController'> requires all parsers
to support file parsing, found: ['multipart/form-data', 'application/json']
```

The runtime guard in `FileMetadata.provide_context_data` is kept as defence in depth
(and its test now forces the negotiator to return a non files-ready parser directly,
since such a controller can no longer be defined).

## Changes

- `dmr/components.py`: `any` → `all` in `FileMetadata.validate`, docstring and error
  message updated; also keeps raising when no parsers are declared at all
- `tests/test_unit/test_components/test_file_metadata.py`: existing test updated,
  new test for the exact example from this issue, runtime-guard test reworked
- `tests/test_unit/test_openapi/test_files_snapshots.py`: the mixed-parsers controller
  from the issue is now covered as an import-time rejection instead of a schema snapshot;
  the stale syrupy snapshot was removed
- `CHANGELOG.md`: entry under Bugfixes

## Checks

- `uv run pytest -q` — full suite: see the run in the checks section below
- `uv run mypy .` — no issues in 707 source files
- `uv run ruff check` / `ruff format --check` — clean

Closes #1493


sobolevn (commented):
I am not sure that my description was fully correct :) 
Now, when thinking about it again, including https://github.com/wemake-services/django-modern-rest/issues/1494, I think that we should do it a bit differently.

`JsonParser()` can parse `Body[...]` and `FileMetadata[None] = None` from `def post(self, parsed_body: Body[...], parsed_file_metadata: FileMetadata[Files | None] = None)`

So, we need to first complete the default values support. Next, we should change the validation again. If `FileMetadata` does not have a default value - then we require `all()` if it does, only `any()`.

PR #1515, opened 2026-09-14
Allow customizing OpenAPI generators and ConfigMerger

### LINKED ISSUE #1461: Make sure that it is possible to customize `operation_id` generation
If people need to tailor the `operation_id` exact look, we need to be able to provide a callback to do that.

I think that we need to do it somewhere here: 
- https://github.com/wemake-services/django-modern-rest/blob/405ba9b05fe583954d49683be168a9fc584a3169/dmr/openapi/core/context.py#L29
- https://github.com/wemake-services/django-modern-rest/blob/405ba9b05fe583954d49683be168a9fc584a3169/dmr/openapi/core/context.py#L80

Probably via a callback? Other options are also fine :) 
But, make sure that the default implementation will be kept.

### LINKED ISSUE #1487: Make `ConfigMerge` class customizable
Currently, there's no way to customize the `ConfigMerger` class here: https://github.com/wemake-services/django-modern-rest/blob/f822e35bd2db672dcc43fbd4967a22835a616971/dmr/openapi/core/context.py#L61-L69

This goes against our design, where we provide all classes as `ClassVar[type[ConfigMerger]]` objects.

It should be:

```python
class OpenAPIContext:
     config_merger_cls: ClassVar[type[ConfigMerger]] = ConfigMerger

     # then:
     # self.config_merger = self.config_merger_cls(self)
```

soil0119:
Refs #1461 and #1487

## Summary

Made the generator classes used by `OpenAPIContext` replaceable so users can customize OpenAPI `operationId` generation.
Applied the same pattern to all six generators and `ConfigMerger`, while keeping the existing default implementations.
Added tests to verify that explicit operation IDs take precedence and duplicate IDs are still detected. Also added a usage example, documentation, and a CHANGELOG entry.

Lint and type checks passed!

## AI Policy

## Checklist

## Related issues

- Closes #1461
- Closes #1487


docs/examples/openapi/custom_context.py
```diff
@@ -0,0 +1,59 @@
...
+        title = schema.info.title
+        schema.info.title = f'{title} (custom)'
+        return schema
+
+
+class CustomContext(OpenAPIContext):
+    operation_id_cls: ClassVar[type[OperationIdGenerator]] = (
+        PathOperationIdGenerator
+    )
+    config_merger_cls: ClassVar[type[ConfigMerger]] = CustomConfigMerger
```
sobolevn: please, remove all code about `ConfigMerger` from this example. you show it as an example that changes operation_id generation. so, let's not overly complicate this.

tests/test_unit/test_openapi/test_core/test_context_customization.py
```diff
@@ -0,0 +1,241 @@
...
+    router = Router('api/', [path('users/', _UserController.as_view())])
+    config = OpenAPIConfig(title='Users API', version='1.0.0')
+    default_schema = build_schema(router, config=config).convert()
+
+    custom_schema = build_schema(
+        router,
+        context=_CustomContext(config),
+    ).convert()
+
+    assert custom_schema['paths']['/api/users/']['get']['operationId'] == (
```
sobolevn: This is not how it is done :) 
Use `== snapshot` to create a `syrupy` schema snapshot file, see how we do it in other tests.

tests/test_unit/test_openapi/test_core/test_context_customization.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+    assert build_schema(router, config=config).convert() == default_schema
+    # Everything except the intentionally customized fields stays the same:
+    default_schema['info']['title'] = 'Users API (custom)'
+    default_schema['paths']['/api/users/']['get']['operationId'] = 'getApiUsers'
+    assert custom_schema == default_schema
+
+
+@pytest.mark.parametrize('serializer', serializers)
+def test_custom_generated_id_collision(
```
sobolevn: this test is not needed.

soil0119:
I have incorporated all your feedback and completed the work. Please take a look. Have a nice day.

tests/test_unit/test_openapi/test_core/test_context_customization.py
```diff
@@ -0,0 +1,178 @@
...
+        ('response', _CustomResponseGenerator, ResponseGenerator),
+        (
+            'security_scheme',
+            _CustomSecuritySchemeGenerator,
+            SecuritySchemeGenerator,
+        ),
+        ('parameter', _CustomParameterGenerator, ParameterGenerator),
+    ],
+)
+def test_generator_classes(
```
sobolevn: This test is not needed.

tests/test_unit/test_openapi/test_core/test_context_customization.py
```diff
@@ -0,0 +1,178 @@
...
+    generator = getattr(context.generators, name)
+
+    assert isinstance(generator, custom_cls)
+    assert generator is not getattr(other_context.generators, name)
+    default_generator = getattr(OpenAPIContext().generators, name)
+    assert isinstance(default_generator, default_cls)
+    assert not isinstance(default_generator, custom_cls)
+
+
+def test_config_merger_class() -> None:
```
sobolevn: this test is not needed.

sobolevn (commented):
You can remove `msgspec` serializer from the test. Since it produces a bit different result.

sobolevn (approved):
Now this seems great! Thank you!

PR #1517, opened 2026-09-14
Return security=[] for endpoints without auth when global security is set

### LINKED ISSUE #1497: Test explicit `OpenAPIConfig.security` definition with `auth=None` per-endpoint

Aditya-myst:
Closes #1497

## What was wrong

If `OpenAPIConfig.security` is set, endpoints with `auth=None` still got `security=None`.
`None` is dropped on dump, so the `security` key was missing and the endpoint inherited
the global security. So `auth=None` endpoints looked protected in the schema.

## What I changed

In `SecuritySchemeGenerator.__call__`, when there are no auth providers:
- global `security` is set -> return `[]` (explicit opt-out)
- otherwise -> return `None` as before

## Tests

Two new unit tests in `test_security_scheme.py`: no global security -> `None`,
global security set -> `[]` (both `None` and `[]` inputs). CHANGELOG updated.

## Local run

Windows, full suite. Only failures: 3 `test_allauth_session_flow` cases
(fail the same on clean `master`, inline_snapshot `Is(...)` error) and the Redis
throttling tests (no Docker here). Neither is related to this change.


sobolevn (approved):

Aditya-myst:
Thanks for the quick review and the CI fix!.

PR #1519, opened 2026-09-14
feat: support explicitly abstract controllers

### LINKED ISSUE #1458: Support explicitly abstract controllers
Currently we only infer `Controller.is_abstract` ourself: https://github.com/wemake-services/django-modern-rest/blob/405ba9b05fe583954d49683be168a9fc584a3169/dmr/controller.py#L206

But, there might be valid cases where people want to define controller to be reused without any customizations or with just the input / return type customizations.

We must respect top-level `is_abstract` controller definition. Child classes without explicit `is_abstract = True` must be treated as concrete if they have the serializer.

This way people can create generic controller only on some parts. This would also need a doc update in the "Reusable code" section with the example.

jarik2014:
## What this does

Explicit `is_abstract` controller definitions are now respected, as requested in #1458.

## Why

`Controller.is_abstract` used to be fully inferred, so it was impossible to have a controller
with an exact serializer and endpoints that is deliberately *not* routable. People had to
defeat the inference (for example by using a `TypeVar` instead of a real serializer),
which in turn lost the types.

With this change:

```python
class MyBaseController(Controller[PydanticSerializer]):
    is_abstract = True  # has a serializer and an endpoint, but is not routed

    def get(self) -> str:
        return 'hello'


class MyController(MyBaseController):
    # does not declare `is_abstract`, so it is concrete again
    ...
```

`MyBaseController` keeps its `api_endpoints` metadata, but `as_view()` raises
`EndpointMetadataError`, while `MyController` is routable with the very same logic.

## Changes

- `dmr/controller.py`: an explicitly declared `is_abstract` wins over the inference.
  `cls.__dict__` is used on purpose, so subclasses of abstract controllers are
  concrete again without any extra declarations; the attribute docs
  document the new behavior
- `tests/test_unit/test_controllers/test_abstract_as_view.py`: tests for an explicitly
  abstract controller, for its concrete subclass, and for a subclass that opts in
  to being abstract as well
- `docs/pages/reusable-code.rst`: new "Explicitly abstract controllers" part
  in the "Reusable code" section
- `docs/examples/reusable_code/explicit_abstract.py`: the example from the docs
- `CHANGELOG.md`: Features entry

## Checks

- The new tests fail on the old inference and pass with the fix (verified by running
  them against the unmodified `dmr/controller.py`)
- `uv run pytest -q` — full suite
- `uv run mypy .` — no issues in 708 source files
- `uv run ruff check`, `ruff format --check`, `flake8`, `slotscheck`, `import-linter` — clean

Closes #1458


jarik2014:
Heads-up on the single red check: `lychee` failed on this branch during a GitHub 503 window (`blob/master/LICENSE`, `.github/SECURITY.md` and other untouched URLs returned 503). The same job failed on `master` in the same window and passes again since 09:57 UTC, so it is not related to the diff — nothing in this PR adds or changes a link.

docs/examples/reusable_code/explicit_abstract.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def get(self) -> str:
+        return 'hello from the base controller'
+
+
+class MyController(MyBaseController):
+    """It does not declare `is_abstract`, so it is concrete again."""
+
+    # It serves the same `GET` request as `MyBaseController`,
+    # but now it can be routed.
```
sobolevn, line 18: Please add `# run` and `# openapi` comments for `MyController` example.

docs/pages/reusable-code.rst
```diff
@@ -82,6 +82,25 @@ But, this is just the first step. We can do much more!
     :linenos:
     :language: python
 
+Explicitly abstract controllers
```
sobolevn, line 85: This should be the first section in this document. And use `---` for this header.

tests/test_unit/test_controllers/test_abstract_as_view.py
```diff
@@ -50,3 +50,52 @@ def get(self) -> str:
...
+
+    class _Custom(Controller[PydanticSerializer]):
+        is_abstract = True
+
+        def get(self) -> str:
+            raise NotImplementedError
+
+    assert _Custom.is_abstract
+    # It still has a real endpoint, it is just not routed:
+    assert _Custom.api_endpoints
```
sobolevn, line 66: Nope :) 
This should not happen. It should be empty. This is exactly the difference between concrete and abstract controllers.

We require them to build endpoints in the correct non-absctract context. Please, fix this :)

dmr/controller.py
```diff
@@ -118,8 +118,16 @@ class Controller(View, Generic[_SerializerT_co]):  # noqa: WPS214
...
             We consider controller "abstract" when it does not have
             exact serializer type or exact ``api_endpoints`` instances.
+            You can also set it to ``True`` explicitly to make
+            a controller with a serializer and endpoints reusable
+            without routing it. Subclasses that don't declare
+            ``is_abstract`` themselves become concrete again.
             Abstract controllers cannot be routed, ``as_view`` raises
             :class:`~dmr.exceptions.EndpointMetadataError` for them.
+
+            .. versionchanged:: 0.16.0
```
sobolevn, line 128: I prefer to have this in a single place at the very bottom of the docstring.

sobolevn (commented):
Thanks! One major comment.

Issue #1491, opened 2026-09-13
Fix `Reference | Schema` conflict in our OpenAPI definitions


sobolevn:
Currently we mix up two different things in the code: `Schema` with `$ref` set and a `Reference` object with `$ref` set.

For example, this code:

```python
import pydantic
from pydantic.json_schema import GenerateJsonSchema
from dmr.openapi.mappers.schema_loader import load_schema


class WithDialect(GenerateJsonSchema):
    def generate(self, schema, mode='validation'):
        json_schema = super().generate(schema, mode=mode)
        json_schema['$schema'] = self.schema_dialect
        return json_schema


class Address(pydantic.BaseModel):
    model_config = pydantic.ConfigDict(
        json_schema_extra={'$anchor': 'address', '$comment': 'Postal address'},
    )
    city: str


class User(pydantic.BaseModel):
    model_config = pydantic.ConfigDict(
        json_schema_extra={'$comment': 'Internal note for schema readers'},
    )
    name: str = pydantic.Field(json_schema_extra={'$comment': 'Display name'})
    address: Address = pydantic.Field(
        default=Address(city='Moscow'),
        description='Where the user lives',
    )


raw = pydantic.TypeAdapter(User).json_schema(
    ref_template='#/components/schemas/{model}',
    schema_generator=WithDialect,
)
defs = raw.pop('$defs')
```

Produces this schema:

```json
{
  "$comment": "Internal note for schema readers",
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "properties": {
    "name": {"$comment": "Display name", "title": "Name", "type": "string"},
    "address": {
      "$ref": "#/components/schemas/Address",
      "default": {"city": "Moscow"},
      "description": "Where the user lives"
    }
  },
  "required": ["name"], "title": "User", "type": "object"
}
```

But, we would load it as `Reference` object in `load_schema` and we will loose `default` and `description`. Which are `Schema` object attributes.

So, what we need to do?
1. Analyze all places where `Reference` can't be even used based on https://spec.openapis.org/oas/v3.2.0.html We need to grep this page with `| Reference Object` and check that we don't have more places that can hold `Reference` objects
2. We must distinguish `$ref` in potential `Schema` objects and `Reference` objects
[...]

PR #1520, opened 2026-09-14
fix: load `$ref` with sibling keywords as a schema

jarik2014:
## What this does

`$ref` combined with sibling keywords is now loaded as a `Schema`, as asked in #1491.
This is the first part of the issue: distinguishing `$ref` in `Schema` objects
from real `Reference` objects, so keywords like `default` and `description`
are no longer lost.

## Why

A `Reference` Object cannot hold anything but `$ref`, `summary` and `description` —
the spec says that any other property added to it SHALL be ignored.
JSON Schema / OpenAPI schema objects, on the other hand, allow `$ref`
next to other keywords.

So, this raw schema (the example from the issue):

```json
{
  "$ref": "#/components/schemas/Address",
  "default": {"city": "Moscow"},
  "description": "Where the user lives"
}
```

was loaded as a `Reference` and `default` was silently dropped, even though
`Schema` already has both `default` and `ref` attributes.

## Changes

- `dmr/openapi/mappers/schema_loader.py`:
  - only values that can *only* be a `Reference` object (`$ref` plus optional
    `summary` / `description`) are loaded as `Reference`, everything else
    becomes a `Schema` with `$ref` and its siblings
  - `load_schema` now reads the `$ref` keyword, it used to not do it at all,
    so a `Schema` with a reference would have lost it
- `tests/test_unit/test_openapi/test_mappers/test_schema/test_jsonschema_load.py`:
  regression tests for a `$ref` with siblings in two nested positions
  (`properties` and `items`) and for a plain reference object
- `tests/test_unit/test_plugins/test_pydantic/test_pydantic_schema.py` and
  `tests/test_unit/test_plugins/test_msgspec/test_msgspec_schema.py`:
  enum parameter tests for both serializers. They used to assert
  `{'$ref': ...}` only; now they assert that `default` is kept next to `$ref`.
  This is the same data loss, but in a real parameter, not in a synthetic dict
- `CHANGELOG.md`: Bugfixes entry

## What is not here

This PR deliberately stops at the `Reference | Schema` distinction:

- resolving `$ref` inside `Schema` objects in `maybe_resolve_reference`
  (the `ResolvedSchema` idea from the issue) is a separate design decision,
  and it changes the resolution call sites, not the loading;
- preserving custom `x-` attributes needs an API decision as well:
  `Schema` is a slotted dataclass, so there's no place to keep unknown
  keywords yet. Happy to open a follow-up with a proposal (for example
  `Schema.extra` that is merged into the dumped schema), or to add it
  to this PR, if you prefer it in one go.

## Checks

- The new regression tests fail on the unmodified `schema_loader.py`
  and pass with the change (verified by running them against the old code)
- The two serializers now both show the preserved `default`, so the fix
  is not a `pydantic`-only behaviour
- `uv run pytest tests/test_unit/test_plugins tests/test_unit/test_openapi -q` —
  435 passed, 44 snapshots, so existing documents are not affected
- `uv run mypy .` — no issues in 707 source files
- `uv run ruff check`, `ruff format --check`, `flake8` — clean

Refs #1491


jarik2014:
Heads-up on the single red check: `lychee` failed on this branch during a GitHub 503 window (`blob/master/LICENSE`, `.github/SECURITY.md` and other untouched URLs returned 503). The same job failed on `master` in the same window and passes again since 09:57 UTC, so it is not related to the diff — nothing in this PR adds or changes a link.

dmr/openapi/mappers/schema_loader.py
```diff
@@ -22,6 +22,14 @@
 
 _EnumT = TypeVar('_EnumT', bound=Enum)
 
+_REF_KEY = '$ref'
```
sobolevn, line 25: All constants must be `Final`.

sobolevn (commented):
This is not all. What is also needed:
- schema `$ref` resolution to flat schema
- audit that `Reference` takes only places that are allowed by the spec

jarik2014:
rebased onto master, mergeable again.

Conflict was in `tests/.../test_jsonschema_load.py`: master had grown its own tests for that file (`test_load_schema_openapi_v32_fields`, `test_load_schema_without_xml_node_type`, `test_load_schema_format_preserve_type`). I kept all three and appended this PR's three `$ref`-with-siblings tests, adding `Reference` to the import list. The loader change itself applied cleanly and is unchanged.

```
uv run pytest tests/test_unit/test_openapi/test_mappers/test_schema/test_jsonschema_load.py \
  tests/test_unit/test_plugins/test_msgspec/test_msgspec_schema.py \
  tests/test_unit/test_plugins/test_pydantic/test_pydantic_schema.py -q
  56 passed, 2 skipped
```

(the coverage gate in that run is red only because a subset cannot reach the repo's 100% requirement; the full suite is CI's job.)

PR #1525, opened 2026-09-14
Support `TypeVar` defaults in controller type var inference

### LINKED ISSUE #1452: Support `default` values for `TypeVar` when using controller type var inference

vyhuholl:
# I have made things!

The `TODO` in `_apply_base_type_params` was right, but not about everything — one of the three cases already worked.

Partial subscription is fine today. `class Sub(Reusable[PydanticSerializer])` with a defaulted `_ModelT` gets the default, because CPython substitutes defaults at subscription time: `Reusable[PydanticSerializer]` is already `Reusable[PydanticSerializer, DefaultModel]` by the time we look at it. That holds on 3.11 through 3.14, for `typing_extensions.TypeVar` and for the native syntax. So that case only needed tests.

Two cases really were broken:

1. A bare subclass. `class Sub(Reusable): ...` raised `UnsolvableAnnotationsError`, because `get_original_bases()` hands back the plain class and there are no type args to read. mypy and pyright both apply defaults here, so code that type-checks clean blew up at import time.
2. A default that is itself a type var: `_LinkedT = TypeVar('_LinkedT', default=_ModelT)`. `TypeVarInference` recorded only the type params it already knew it needed, so when `_LinkedT -> _ModelT` and `_ModelT -> ExactModel` both came from the same base in that order, `_ModelT` was not in the map yet when the first link was read. It resolved to itself and hit the `must be concrete` error.

## What I changed

The code changes are all in `dmr/types.py`:

- New helper, `_resolve_type_args()`. For a subscripted base it returns `get_args()` untouched; for a bare one it substitutes the defaults. When no param has a default it returns `()`, so a bare base contributes nothing and `class _Custom(Controller)` still raises — `test_controller_without_serializer` passes unchanged.
- `infer_bases()` keeps bare bases now (`get_origin(base) or base`), otherwise they never reach the helper above.
- The real fix for case 2 is in `_apply_base_type_params()`: it records every type param, not just the ones already known to be needed. The `is_needed` gate was the bug. The `assert isinstance(type_param, TypeVar)` above it already ran for every param, so dropping the gate doesn't widen what that assert can hit.
- When nothing in the MRO provides a value, `_resolve_type_var()` now falls back to the type var's own default.
- `_resolve_type_var_default()` reads `getattr(type_var, '__default__', NoDefault)`. I used that rather than `has_default()` because typeshed only declares those members from 3.13: a `type: ignore` that 3.11 needs reads as unused on 3.14, and `just type-check` runs on both. The `getattr` works everywhere, for native type vars and for `typing_extensions` ones.

## One thing I'd like a second opinion on

`class Sub(Reusable[PydanticSerializer, _ModelT])` — explicitly passing a type var that has a default — used to raise, and now builds with the default. I think that's right, since mypy infers the same thing when such a class is used bare, and `class Grandchild(Sub[ExactModel])` still overrides it. Type vars without defaults raise exactly as before. But it does change an error path into a success path, so I'd rather you looked at it than not.

## AI Policy

## Checklist

## Related issues

Closes #1452


docs/pages/reusable-code.rst
```diff
@@ -188,6 +188,51 @@ And then - implementations:
...
+  :linenos:
+  :language: python
+
+Now, both type args are optional for the subclasses:
+
+.. tabs::
+
+  .. tab:: defaults
+
+    .. literalinclude:: /examples/reusable_code/defaults_pydantic.py
```
sobolevn, line 216: Let's both examples use `pydantic`. Because changing the serializer is not a part of the logic here. Just show how we can change the request model type with / without defaults.

docs/pages/reusable-code.rst
```diff
@@ -188,6 +188,51 @@ And then - implementations:
...
 
+Type variable defaults
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 0.16.0
+
+Type variables can have defaults, as described in :pep:`696`.
+We use them when a subclass does not provide some of the type args.
+
+Use ``typing_extensions.TypeVar`` to define them,
```
sobolevn:
```suggestion
Use ``typing_extensions.TypeVar`` on <= 3.12 to define them,
```

docs/pages/reusable-code.rst
```diff
@@ -188,6 +188,51 @@ And then - implementations:
...
+the native ``class Reusable[_ModelT = MyModel]`` syntax
```
sobolevn: also mention `typing.TypeVar` on 3.13+, we support it as well.

docs/pages/reusable-code.rst
```diff
@@ -188,6 +188,51 @@ And then - implementations:
...
+  .. tab:: exact types
+
+    .. literalinclude:: /examples/reusable_code/defaults_msgspec.py
+      :caption: views.py
+      :linenos:
+      :language: python
+
+Defaults can also point to other type variables:
+``_ResponseBodyT = TypeVar('_ResponseBodyT', default=_RequestModelT)``
+means "the response body is the request model, unless told otherwise".
```
sobolevn, line 230: Add a note that serializers can also have defaults. What will happen in this case? Will the controller be abstract?

tests/test_unit/test_controllers/test_reusable_generics_defaults.py
```diff
@@ -0,0 +1,228 @@
...
+                def post(self, parsed_body: Body[_ModT]) -> str:
+                    ...
+
+            class NativeBare(_NativeController): ...
+            """,
+        ),
+        ns,
+    )
+
+    controller = ns['NativeBare']
```
sobolevn: please, assert `NativeController` as well. It is very important to handle this case correctly and describe it in the docs as well.

tests/test_unit/test_controllers/test_reusable_generics_defaults.py
```diff
@@ -0,0 +1,228 @@
...
+    Controller[_SerializerT],
+    Generic[_SerializerT, _ModelT],
+):
+    """Both the serializer and the request model have defaults."""
+
+    def post(self, parsed_body: Body[_ModelT]) -> str:
+        raise NotImplementedError
+
+
+class _LinkedController(
```
sobolevn: please, define controller near tests that use them. it is hard to read test code that is a long distance apart.

tests/test_unit/test_controllers/test_reusable_generics_defaults.py
```diff
@@ -0,0 +1,228 @@
...
+        """It is still generic, but it can be routed with the default."""
+
+    class ExactController(StillGenericController[_ExactModel]):
+        """And it can still be subclassed with an exact model."""
+
+    assert _body_models(StillGenericController) == [_DefaultModel]
+    assert _body_models(ExactController) == [_ExactModel]
+
+
+def test_type_var_without_default_stays_abstract() -> None:
```
sobolevn: not a single `async` controller is tested. let's have one.

setup.cfg
```diff
@@ -55,6 +55,8 @@ per-file-ignores =
   dmr/throttling/*.py: WPS226
   # It is fine to have many exceptions:
   dmr/exceptions.py: WPS202
+  # Typing helpers and the type var inference engine live together:
```
sobolevn: We can move internal parts to `internal/`, there's no need to ignore this error. Create new `internal/type_inference.py` module and place all needed parts there. New files are free.

sobolevn (commented):
Thank you! Looks like a solid start, but there are some things to fix :)

docs/pages/reusable-code.rst
vyhuholl: Both tabs now use pydantic. `_SerializerT` in `reusable_defaults.py` is now a plain `TypeVar` with no default, so the example is only about the request model. Tabs are "with default" / "without default", both `PydanticSerializer`. `defaults_msgspec.py` is gone, replaced by `defaults_exact.py`.

docs/pages/reusable-code.rst
vyhuholl: Added a mention.

docs/pages/reusable-code.rst
vyhuholl: Added serializer defaults note. Answering your question directly: the subclass is concrete, the declaring controller is not. I verified both before writing it — `ReusableController` with every type var defaulted still reports `is_abstract=True` and has no serializer, because `infer_type_args` reads its own `Controller[_SerializerT]` base, which is subscripted with a type var. Only a subclass reaches the bare-base path where defaults apply. New example `defaults_serializer.py` shows both classes side by side, plus an `.. important::` block stating the rule.

tests/test_unit/test_controllers/test_reusable_generics_defaults.py
vyhuholl: Moved controllers next to their tests. `_LinkedController` and `_MixedController` are now nested in the single test that uses each — the two `_MixedController` tests merged into `test_type_var_without_default`, which is one scenario anyway. `_BaseController` stays at module level; four tests use it.

tests/test_unit/test_controllers/test_reusable_generics_defaults.py
vyhuholl: `test_defaults_in_async_controller` now asserts `is_async`, the defaulted serializer, and the defaulted body model.

setup.cfg
vyhuholl: Done.

tests/test_unit/test_controllers/test_reusable_generics_defaults.py
vyhuholl: `test_pep696_type_params` now asserts the base too: `NativeController.is_abstract`, no `serializer`, no `api_endpoints`. I also added `test_reusable_controller_is_still_abstract` for the `typing_extensions` path, since the same rule matters there.

docs/examples/reusable_code/defaults_exact.py
```diff
@@ -0,0 +1,18 @@
...
+    email: str
+
+
+class PydanticController(
+    ReusableController[PydanticSerializer, _RequestModel],
+):
+    """Defaults are only used when type args are missing."""
+
+
+# run: {"controller": "PydanticController", "method": "post", "body": {"email": "example@example.com"}, "url": "/api/example/"}  # noqa: ERA001, E501
```
sobolevn, line 22: I think we should add an example with the default model and show that it won't validate.

docs/pages/reusable-code.rst
```diff
@@ -188,6 +188,72 @@ And then - implementations:
...
+.. tabs::
+
+  .. tab:: with default
+
+    .. literalinclude:: /examples/reusable_code/defaults_pydantic.py
+      :caption: views.py
+      :linenos:
+      :language: python
+
+  .. tab:: without default
```
sobolevn:
```suggestion
  .. tab:: substituting default
```

docs/examples/reusable_code/defaults_exact.py
vyhuholl: Added. `defaults_exact.py` now has a second `# run:` that posts the *default* model's payload to a controller that substituted it away.

sobolevn (approved):
Thanks! LGMT now!

PR #1529, opened 2026-09-15
Support missing OpenAPI 3.2 features

### LINKED ISSUE #1485: Support missing OpenAPI 3.2 features
All of these was found by Claude. It would be very hard to do by hand.

Please, be really careful that I did not manually reviewed all of the missing items in the spec. This is something that must be done by the contributor.

| 3.2 feature | Field | Status |
|---|---|---|
| Self-identifying documents | `OpenAPI.$self` | ❌ absent from `dmr/openapi/openapi.py` |
| Named servers | `Server.name` | ❌ `objects/server.py` |
| Tag hierarchy / classification | `Tag.summary`, `Tag.parent`, `Tag.kind` | ❌ `objects/tag.py` |
| Whole-query-string params | `Parameter.in: "querystring"` | ❌ no component, no `in` value |
| Reusable media types | `Components.mediaTypes` | ❌ `objects/components.py` |
| Response short label | `Response.summary` | ❌ `objects/response.py`, `ResponseSpec` |
| Media type prose | `MediaType.description` | ❌ `objects/media_type.py` |
| Nested / sequential encoding | `Encoding.encoding`, `Encoding.itemEncoding`, `Encoding.prefixEncoding` | ❌ `objects/encoding.py` |
| Structured vs serialized examples | `Example.dataValue`, `Example.serializedValue` | ❌ `objects/example.py` |
| XML node typing | `XML.nodeType` (`element`/`attribute`/`text`/`cdata`/`none`) | ❌ `objects/xml.py`; `attribute`/`wrapped` are now *deprecated* |
| Discriminator fallback | `Discriminator.defaultMapping` | ❌ `objects/discriminator.py`, `_try_discriminator` |
| OAuth 2.0 device grant | `OAuthFlows.deviceAuthorization` | ❌ `objects/oauth_flows.py` |
| OAuth metadata discovery | `SecurityScheme.oauth2MetadataUrl` | ❌ `objects/security_scheme.py` |
| Deprecating a scheme | `SecurityScheme.deprecated` | ❌ `objects/security_scheme.py` |

Also: 3.2 deprecates the OAS `example` keyword *inside Schema Objects* in favour of
JSON Schema `examples`. `SchemaGenerator._maybe_generate_example` and
`schema_loader.load_schema` both write `Schema.example`, i.e. the framework's
auto-generated examples land on a deprecated keyword.

vyhuholl:
# I have made things!

## The table in the issue checks out

You asked the contributor to verify it, so that came first. I compared every row against `versions/3.2.0.md` and `versions/3.2.1.md` in the OAI repo, and against the released schema at https://spec.openapis.org/oas/3.2/schema/2025-09-17. All 14 rows are real gaps.

One row needs a note. `MediaType.description` is missing from the fixed-fields table in both markdown versions of the spec, but it is in the released JSON Schema, and `openapi-spec-validator` accepts it. I added it on that basis. Say the word and I'll drop it until the prose catches up.

I also hit something the issue doesn't list. `MediaType.prefix_encoding` was typed as a single `Encoding`, but the spec defines `prefixEncoding` as an array of them. Fixed here, because the same commit reworks `Encoding` anyway.

## What I changed

`CHANGELOG.md` has the field-by-field list, so I won't repeat it.

I gated the `example` deprecation on the version. Generated examples go to the JSON Schema `examples` list for `3.2.0` and newer, and stay on `example` below that; examples written by hand are never touched. The gate lives in `set_generated_example` in `dmr/openapi/mappers/example.py`. This let me pull example generation out of `schema_loader.load_schema` entirely — that function can't know the target version, and it no longer takes `should_generate_example`, `annotation`, or `serializer`.

Three other things ended up breaking, all listed in the changelog: `XML.attribute` and `XML.wrapped` now default to `None` (3.2 forbids them next to `node_type`, and we were dumping `attribute: false, wrapped: false` into every `xml` object), `prefix_encoding` became a list, and the `content` maps on `Response`, `RequestBody`, `Parameter` and `Header` now accept `Reference` — without that, `Components.media_types` is declarable but not referenceable.

Where the spec states an invariant I validate it in `__post_init__`: `querystring` parameters must use `content`, `Example.value` excludes the three new value fields, `node_type` excludes `attribute` and `wrapped`, and `self_uri` on a pre-3.2 config raises.

## What I did not do

No `Querystring` request component. The issue's row reads "no component, no `in` value" — I did the `in` value, so `ParameterLocation` now includes `'querystring'` and the object validates, but I did not write a parser that reads `QUERY_STRING` and feeds a model. That is request-parsing work next to `Query` and `Body`, not an OpenAPI object gap, and folding it in here would make a 42-file diff harder to review. Happy to do it as a follow-up if you want it in this milestone instead.

## AI Policy

## Checklist

## Related issues

Closes #1485


dmr/openapi/generators/component_parsers.py
```diff
@@ -207,13 +207,17 @@ def _merge_contents(
...
         for media_name, media_type in new_schema.content.items():
+            # We've just built these bodies from component parsers,
+            # so all of them have inline media types, never references:
+            assert isinstance(media_type, MediaType)  # noqa: S101
             media_items: list[Reference | Schema] = []
             if media_type.schema:  # pragma: no cover:
                 media_items.append(media_type.schema)
             existing_content = schema.content.get(media_name)
+            assert not isinstance(existing_content, Reference)  # noqa: S101
             # TODO: remove pragma after implementing conditional types
```
sobolevn, line 221: Can you please create a new issue to look at this `TODO`? Because we now support conditional `FileMetadata` bodies with `OctetStreamParser` and `MultiPartParser`

dmr/openapi/generators/response.py
```diff
@@ -77,6 +82,18 @@ def get_schema(
             ),
         )
 
+    def _get_summary(
```
sobolevn: I don't agree with this decision. One of our main design goals is to do not touch user's input in the schema. Hide nothing.

So, if some users wants to add `summary` to the response spec, valid options are:
- Get a validation error from `openapi-validator` early in the development and to remove this change / bump version
- To disable the validation and do it anyway

Forcefully hiding the value is not an option :)

dmr/openapi/generators/response.py
```diff
@@ -115,7 +132,13 @@ def _get_cookies(
             schema = context.generators.schema(str, serializer)
             # for mypy: `str` cannot return a reference, it is a primitive
             assert isinstance(schema, Schema)  # noqa: S101
-            schema = dataclasses.replace(schema, example=f'{name}=123')
+            # `replace` copies the schema, so we don't set the cookie
+            # example on the shared `str` schema:
+            schema = set_generated_example(
```
sobolevn, line 128: Sorry, this does not look correct to me :( 
There was a hidden bug here. For some reason, this place does not use our `example` generation logic and always produces `123` as an example. This is not cool.

Please, reuse our example generation logic here. Right now it is not configurable, it is static, it uses a very generic data.

I don't remember why I wrote this code :)

dmr/openapi/mappers/example.py
```diff
@@ -1,11 +1,42 @@
...
+    is ``None`` and we don't write anything at all.
+
+    .. versionadded:: 0.16.0
+    """
+    if example is None:
+        return schema
+    if context.config.openapi_version_info[:2] < _JSON_SCHEMA_EXAMPLES_VERSION:
+        schema.example = example
+    else:
+        schema.examples = [example]
```
sobolevn, line 24: Can we always use the same field? Can we just ignore `example` field on 3.1 and 3.2?

dmr/openapi/objects/encoding.py
```diff
@@ -8,10 +8,30 @@
...
     style: str | None = None
     explode: bool | None = None
     allow_reserved: bool | None = None
+
+    # OpenAPI 3.2+ fields:
+    encoding: dict[str, 'Encoding'] | None = None
+    item_encoding: 'Encoding | None' = None
+    prefix_encoding: list['Encoding'] | None = None
+
+    def __post_init__(self) -> None:
```
sobolevn: I think we should remove all runtime validation from these new types. 

The same is true for `dmr/openapi/objects/media_type.py`. I only added that as a sanity-check. We rely on `openapi-validator` for this, let's not re-implement it.

dmr/openapi/config.py
```diff
@@ -91,6 +104,14 @@ def __post_init__(self) -> None:
                 'they are not based on JSON Schema, which is what we use '
                 f'to generate model schemas, got {self.openapi_version!r}',
             )
+        if (
```
sobolevn: Please, remove this condition.

docs/pages/openapi/openapi.rst
```diff
@@ -17,6 +17,70 @@ We support OpenAPI versions from ``3.1.0`` through ``3.2.0``.
...
 
+.. _openapi-3-2:
+
+OpenAPI 3.2 features
+--------------------
+
+Set ``openapi_version='3.2.0'`` on :class:`dmr.openapi.OpenAPIConfig`
+to target `OpenAPI 3.2 <https://spec.openapis.org/oas/v3.2.0.html>`_.
+
+Everything that 3.2 adds to the spec objects is available:
```
sobolevn: Do not list everything. Just saying that we support all the spec is enough.

docs/pages/openapi/openapi.rst
```diff
@@ -17,6 +17,70 @@ We support OpenAPI versions from ``3.1.0`` through ``3.2.0``.
...
+  and ``deprecated``
+
+Here's how several of them look together:
+
+.. literalinclude:: /examples/openapi/openapi_v32.py
+  :caption: urls.py
+  :language: python
+  :linenos:
+
+Some of these fields do not exist in older versions of the spec.
```
sobolevn: Again, this is not needed.

docs/pages/openapi/openapi.rst
```diff
@@ -17,6 +17,70 @@ We support OpenAPI versions from ``3.1.0`` through ``3.2.0``.
...
+- ``self_uri`` raises a ``ValueError``
+  when ``openapi_version`` is older than ``'3.2.0'``
+- :attr:`ResponseSpec.summary <dmr.metadata.ResponseSpec>` is only rendered
+  for ``'3.2.0'`` and newer, older versions silently drop it
+- Autogenerated examples land on the JSON Schema ``examples`` keyword
+  for ``'3.2.0'`` and newer, because 3.2 deprecates the OAS ``example``
+  keyword inside Schema Objects.
+  See :ref:`openapi-examples-generation`
+
+.. versionadded:: 0.16.0
```
sobolevn: Let's remove it. We claimed 3.2 support a long ago, we just fixed the missing parts.

sobolevn (commented):
Not a full review.

dmr/openapi/generators/component_parsers.py
vyhuholl: I've discovered that the TODO's premise is already satisfied — conditional types for `FileMetadata[]` shipped in 0.15.0 and are tested. But removing the pragmas still doesn't go green, because the misses are the false arms of both ifs, and `_merge_contents` only runs when two components each return a `RequestBody` — which no conditional-file test controller does.

And chasing that arm turned up an additional real bug. I've created an issue about it: see #1537

dmr/openapi/generators/response.py
vyhuholl: Fixed in https://github.com/wemake-services/django-modern-rest/commit/02ca1bed5d3e3d4b3f14b6b7db2daa5c9689edaf

dmr/openapi/generators/response.py
vyhuholl: Done

dmr/openapi/config.py
vyhuholl: Done

docs/pages/openapi/openapi.rst
vyhuholl: Removed

docs/pages/openapi/openapi.rst
vyhuholl: Removed

docs/pages/openapi/openapi.rst
vyhuholl: Removed

dmr/openapi/objects/encoding.py
vyhuholl: Removed

dmr/openapi/mappers/example.py
vyhuholl: Done

tests/test_unit/test_openapi/__snapshots__/test_translated_schema.ambr
```diff
@@ -26,8 +26,7 @@
                 },
                 "Set-Cookie: translated-cookie": {
                   "schema": {
-                    "type": "string",
-                    "example": "translated-cookie=123"
+                    "type": "string"
```
sobolevn, line 29: We need at least one example for cookies with generated data. Currently it is not tested at all.

tests/test_unit/test_openapi/test_generators/test_response.py
```diff
@@ -167,3 +169,40 @@ def test_response_multiple_content_types(
...
+            description='The task was put into the queue',
+        ),
+    )
+
+    def get(self) -> str:
+        raise NotImplementedError
+
+
+@pytest.mark.parametrize('openapi_version', ['3.1.0', '3.2.0'])
+def test_response_summary(openapi_version: str) -> None:
```
sobolevn: this test is not needed, there's no logic / behavior to test. we just dump all fields. `summary` is a field.

tests/test_unit/test_openapi/test_objects/test_openapi_v32_fields.py
```diff
@@ -0,0 +1,209 @@
...
+    SecurityScheme,
+    Server,
+    Tag,
+)
+from dmr.routing import Router
+
+# Makes sure that the spec we build is validated against the real 3.2 schema:
+pytest.importorskip('openapi_spec_validator')
+
+_REUSABLE_MEDIA_TYPE = '#/components/mediaTypes/Reusable'
```
sobolevn:
```suggestion
_REUSABLE_MEDIA_TYPE: Final = '#/components/mediaTypes/Reusable'
```

tests/test_unit/test_openapi/test_objects/test_openapi_v32_fields.py
```diff
@@ -0,0 +1,209 @@
...
+        ),
+    )
+
+
+def test_openapi_v32_document_is_valid() -> None:
+    """Ensure that all the new 3.2 fields pass the real 3.2 validation."""
+    # `convert` validates the result, it raises when something is off:
+    dumped = build_schema(Router('api/v1/'), config=_config()).convert()
+
+    assert dumped['$self'] == 'https://example.com/openapi.json'
```
sobolevn: assert the full document with `== snapshot`, it is worth it.

tests/test_unit/test_openapi/test_objects/test_openapi_v32_fields.py
```diff
@@ -0,0 +1,209 @@
...
+    assert dumped['servers'] == snapshot([
+        {'url': 'https://example.com', 'name': 'production'},
+    ])
+    assert dumped['tags'] == snapshot([
+        {'name': 'external', 'summary': 'External', 'kind': 'audience'},
+        {'name': 'partner', 'parent': 'external', 'kind': 'audience'},
+    ])
+
+
+def test_openapi_v32_components() -> None:
```
sobolevn: so this test will be removed

tests/test_unit/test_openapi/test_config.py
```diff
@@ -39,3 +39,15 @@ def test_unsupported_openapi_versions(openapi_version: str) -> None:
             version='1.0.0',
             openapi_version=openapi_version,
         )
+
+
+def test_self_uri() -> None:
```
sobolevn: I suggest to remove this test. `def _config() -> OpenAPIConfig:` covers it.

tests/test_unit/test_openapi/test_generated_examples.py
```diff
@@ -0,0 +1,67 @@
...
+@pytest.mark.parametrize('openapi_version', ['3.1.0', '3.2.0'])
+@pytest.mark.parametrize(
+    'find_schema',
+    [_inline_body_schema, _referenced_schema],
+)
+def test_generated_examples_keyword(
+    find_schema: Any,
+    openapi_version: str,
+    settings: LazySettings,
+) -> None:
```
sobolevn, line 52:
```suggestion
def test_generated_examples_keyword(
    settings: LazySettings,
    *,
    find_schema: Callable[[dict[str, Any]], Schema | Reference],
    openapi_version: str,
) -> None:
```

tests/test_unit/test_openapi/__snapshots__/test_translated_schema.ambr
vyhuholl: Added `test_response_generator_cookie_examples` with `openapi_examples_seed: 5`, which now pins:

```python
'Set-Cookie: first_cookie': Header(
    schema=Schema(
        type=OpenAPIType.STRING,
        examples=['first_cookie=GMPXMVbyXHUfymCDaloV'],
    ),
    ...
```
I kept the seedless test alongside it, since the two assert different things: one that no example appears when generation is off, the other that a generated one appears when it's on.

tests/test_unit/test_openapi/test_generators/test_response.py
vyhuholl: Removed.

tests/test_unit/test_openapi/test_objects/test_openapi_v32_fields.py
vyhuholl: Done.

tests/test_unit/test_openapi/test_objects/test_openapi_v32_fields.py
vyhuholl: Removed.

tests/test_unit/test_openapi/test_config.py
vyhuholl: Removed.

tests/test_unit/test_openapi/test_generators/test_response.py
```diff
@@ -167,3 +169,45 @@ def test_response_multiple_content_types(
...
+                type=OpenAPIType.STRING,
+                examples=['first_cookie=GMPXMVbyXHUfymCDaloV'],
+            ),
+            description='First',
+            required=True,
+        ),
+        'Set-Cookie: second_cookie': Header(
+            schema=Schema(
+                type=OpenAPIType.STRING,
+                examples=['second_cookie=GMPXMVbyXHUfymCDaloV'],
```
sobolevn, line 202: why all example values are the same? 🤔

sobolevn (commented):
Thanks! Just one question left.

sobolevn:
There are now merge conflicts 🙏

tests/test_unit/test_openapi/test_generators/test_response.py
vyhuholl: Good question! I've investigated it and found a bug in `generate_example` from `dmr/openapi/mappers/example.py`, which the test just made visible.

`polyfactory`'s `seed_random` sets `cls.__random__` and seeds `faker`, but never updates `cls.__random_seed__`:

```python
@classmethod
def seed_random(cls, seed: int) -> None:
    cls.__random__ = Random(seed)
    cls.__faker__.seed_instance(seed)
```

So in `dmr/openapi/mappers/example.py`, `_ExampleFactory.__random_seed__` stays at its class default of `10` forever, and this guard never stops being true:

```python
if _ExampleFactory.__random_seed__ != seed:  # pragma: no cover
    _ExampleFactory.seed_random(seed)
```

With `openapi_examples_seed = 5`, the factory is reseeded before every call, resetting the stream — so every top-level call for the same annotation returns the same value. Measured through the real `settings` fixture:

```
seed=5  → ['GMPXMVbyXHUfymCDaloV', 'GMPXMVbyXHUfymCDaloV', 'GMPXMVbyXHUfymCDaloV']
seed=10 → ['NiPhGgaTklZTpXCJkWJY', 'UpsihdpFaJrqQyqzKcYb', 'MRaOtEacjXkNWWrsBOfM']
```

At seed `10` the guard is false, nothing reseeds, the stream advances, and values vary. The behaviour flips on whether your seed happens to equal an internal default.

It's project-wide, not cookie-specific — `GMPXMVbyXHUfymCDaloV` already appears 4× in `test_example_snapshots.ambr`, as `msg` in three different `ErrorModel` examples and as a dict key in a fourth. Values only vary *inside* one generated model, because that's a single `get_field_value` call.

So I can't make the cookie values differ from the test side — all three cookies are `str`, and each gets its own `generate_example(str, ...)` call, which always returns the stream's first value.

I'd rather not fix the seeding inside this PR. Seeding once and letting the stream run does give variety, but then values depend on how many calls came before — I measured that too: after two unrelated calls the sequence starts at the third value. With a class-level factory shared across tests and `pytest-randomly` shuffling order, that makes every example snapshot in the repo flaky. The fix that actually works is reseeding once per `build_schema` rather than per call, which is a change to the example subsystem, churns every example snapshot, and has nothing to do with OpenAPI 3.2.

I've opened a new issue about it, see #1548

sobolevn (approved):

PR #1531, opened 2026-09-15
Improve testing of complex type-forms

### LINKED ISSUE #1460: Improve testing of complex type-forms
We test quite a lot of type forms in our tests. But, we never test code like:

```python
type UserBody = Body[User]

class MyController(...):
    def post(self, parsed_body: UserBody) -> str: ...
```

This must define the correct component parser for `Body`. The same should be tested with `UserBody: TypeAlias = Body[User]`

For all the serializers.

One more thing: union type with `ResponseSpecMetadata`.

```python
class MyController(...):
    def get(self) -> Annotated[UserModel, ResponseSpecMetadata(headers={'X-Id': HeaderSpec()})] | str: ...
```

I think that there can be bugs in this code, see `unwrap_annotation` in `litestar` on how we can properly unwrap all annotations and get the metadata.

Please, add tests first. Then fix bugs if present. Feel free to discuss the design if needed.

vyhuholl:
# I have made things!

The #1460  issue asked for tests first and guessed there might be bugs. There were two.

## Components behind a `type X = Y` alias were dropped

```python
type UserBody = Body[User]

    def post(self, parsed_body: UserBody) -> str: ...
```

`get_type_hints` hands back the `TypeAliasType` object as is, so the `get_origin(component) is Annotated` check in `ComponentParserBuilder._find_components` was never true and `parsed_body` never made it into `component_parsers`. The controller still imported fine. Every request then died with `TypeError: post() missing 1 required positional argument: 'parsed_body'`. I reproduced it on `PydanticSerializer`, `PydanticFastSerializer` and `MsgspecSerializer`.

`UserBody: TypeAlias = Body[User]` was never affected — in runtime that is a plain assignment, so it is already `Annotated[...]` by the time we look at it.

## `ResponseSpecMetadata` inside a union was ignored

```python
def get(self) -> Annotated[
    User, ResponseSpecMetadata(headers={'X-Id': HeaderSpec()}),
] | str: ...
```

`ResponseSpec.__post_init__` looked for the metadata on the whole return type, and `get_origin` of a union is not `Annotated`. So `headers` and `cookies` came out `None` and the specs were gone without a word: nothing in the schema, nothing validated.

## What I changed

`dmr/types.py` gets two helpers. `unwrap_type_alias` resolves an alias to what it points at. `typing.TypeAliasType` and `typing_extensions.TypeAliasType` are separate classes, so both are checked; aliases of aliases and subscripted generic aliases (`type Payload[T] = Body[T]`) unwrap too. The loop is bounded at 15 because `type A = B; type B = A` is legal Python and would otherwise spin forever. `iter_union_members` yields the members of a union, unwrapping aliases first.

`AnnotationsContext.__call__` now unwraps every resolved hint, so the rest of the pipeline keeps working on real types instead of learning about aliases. `get_annotated_metadata` unwraps as well, for types that never pass through it — an aliased `error_model`, or a `ResponseSpec(SomeAlias, ...)` handed straight to `@validate`.

`_merge_response_spec_metadata` walks the union members and merges what it finds. Merging rather than taking the first hit is deliberate: the members describe one response, so if two of them carry headers, both belong on it.

`setup.cfg` gains a `WPS202` exemption for `dmr/types.py`; two more helpers put it one over the module-member limit.

## Tests

`tests/test_unit/test_components/test_type_aliases.py` covers all six components behind an explicit `TypeAlias`, behind `TypeAliasType`, and behind `type X = Y` (3.12+, through `exec`, the way `test_type_validation.py` already does it). Both alias classes are worth testing: on 3.14 `typing.TypeAliasType is not typing_extensions.TypeAliasType`. Nested aliases, generic aliases, the depth bound and a request-level test across all three serializers are in there too.

Union metadata tests went into `test_response_spec_meta.py`: headers merged across members, aliased return types, and a response that now gets correctly rejected for missing the header its union member declares.

One gap, on purpose. The cross-serializer request test covers Query, Body, Path and Cookies, not Headers. `HeadersComponent` passes the raw `HttpHeaders` object straight through, and pydantic wants a `BaseModel` where msgspec wants a `Struct` — no model type all three accept. That is existing behaviour and has nothing to do with aliases, so I left it. Headers behind an alias is still covered at the `component_parsers` level.

## AI Policy

## Checklist

## Related issues

Closes #1460


dmr/types.py
```diff
@@ -164,6 +181,70 @@ def infer_annotation(annotation: Any, context: type[Any]) -> Any:
     return TypeVarInference(annotation, context)()[annotation]
 
 
+def unwrap_type_alias(annotation: Any) -> Any:
```
sobolevn: I think that `types.py` is not a proper place for internal type helpers. Please, put them into `internal/types.py`

dmr/metadata.py
```diff
@@ -107,10 +109,7 @@ class ResponseSpec:
 
     def __post_init__(self) -> None:
         """If headers and cookies are not set, look for metadata and use it."""
-        metadata = get_annotated_metadata(
-            self.return_type,
-            ResponseSpecMetadata,
-        )
+        metadata = _merge_response_spec_metadata(self.return_type)
```
sobolevn: I don't think that this is a great API :( 
What will happen to other places where we call `get_annotated_metadata`? Should we change all of them?

Please, provide similar tests for other places where we call `get_annotated_metadata`, since we now know that there's a bug in our typing inspection.

What to do instead? Change `get_annotated_metadata` itself. Maybe `ResponseSpecMetadata` must now how to be merged with another `ResponseSpecMetadata`.

docs/pages/components/index.rst
```diff
@@ -49,6 +49,73 @@ All components consist of two parts:
...
+
+.. tabs::
+
+  .. tab:: type
+
+    Requires Python 3.12 or higher.
+
+    .. code:: python
+
+      import pydantic
```
sobolevn: Please, never-ever use non doctest code in the docs. This literally means "dead code with errors".

Move examples to their own files.

docs/pages/components/index.rst
```diff
@@ -49,6 +49,73 @@ All components consist of two parts:
    For example, :data:`~dmr.components.Query`
 
 
+Components behind type aliases
```
sobolevn: I would make this section just a little `.. note`.
It just works as expected, we can just say that type aliases and unions have first-class support everywhere.

docs/pages/error-handling.rst
```diff
@@ -307,6 +307,22 @@ to add missing ``X-Error-Id`` headers for your error responses.
 You can do the same for all responses, not just failing ones.
 For this, override :meth:`~dmr.controller.Controller.to_response`.
 
+When a single response is described by a union of several models,
```
sobolevn: This is not correct. We must merge the metadata in a correct way:
- `User` will be returned with `X-Error-Id` header
- `str` will be returned without `X-Error-Id` header
This is what user said.

If it were `Annotated[User | str, ResponseSpecMetadata(headers={'X-Error-Id': HeaderSpec()})]` - then both responses would have `X-Error-Id` header.

sobolevn (commented):
Please, use version-based `pragma` comments, see https://github.com/asottile/covdefaults#version-specific--pragma-no-cover

dmr/types.py
vyhuholl: Moved `unwrap_type_alias` and `iter_union_members` to `dmr/internal/types.py`. The `WPS202` exemption for `dmr/types.py` in `setup.cfg` became unnecessary, so I dropped it too.

dmr/metadata.py
vyhuholl: Reworked the union handling. `ResponseSpec.__post_init__` no longer has its own metadata lookup: `get_annotated_metadata` itself now unwraps type aliases and, for `MergeableMetadata` subclasses, merges metadata of union members. Every other call site gets alias support for free, and metadata types that cannot be combined are never read off a union.

And added tests for every metadata type we look up (`ParameterMetadata`, `MediaTypeMetadata`, `ConditionalType`, `FieldInfo`, component parsers, `ResponseSpecMetadata`).

docs/pages/components/index.rst
vyhuholl: Moved this example to `docs/examples/error_handling/union_response_headers.py`.

docs/pages/components/index.rst
vyhuholl: Rewrote the Components section into a small note.

docs/pages/error-handling.rst
vyhuholl: Fixed, thank you for correcting.

docs/pages/components/index.rst
```diff
@@ -48,72 +48,17 @@ All components consist of two parts:
...
+  both work as component annotations, including aliases of aliases
+  and subscripted generic aliases like ``type Payload[ModelT] = Body[ModelT]``.
 
-.. versionchanged:: 0.16.0
+  .. versionchanged:: 0.16.0
 
-  ``type X = Y`` aliases used to be ignored:
-  their components were silently not parsed.
+    ``type X = Y`` aliases used to be ignored,
+    their components were silently not parsed.
```
sobolevn: this was a bug, not a feature we introduced, so `.. versionchanged:: 0.16.0` should be removed from the docs.

docs/pages/error-handling.rst
```diff
@@ -307,24 +307,55 @@ to add missing ``X-Error-Id`` headers for your error responses.
...
+  ...     str | int,
+  ...     ResponseSpecMetadata(headers={'X-User-Id': HeaderSpec()}),
+  ... ]
+
+Now ``X-User-Id`` is required for both ``str`` and ``int`` responses,
+and a response without it fails validation.
+
+The same rule applies when several members are annotated:
+a header is required only when every member of the union declares it
+as required. Specs of all members end up in the documentation either way.
```
sobolevn, line 354: Add a note that this is a bad way to design new APIs. We only support it for legacy code and migrations.

New APIs must always have the same required set of headers.

tests/test_unit/test_metadata/test_response_spec_meta.py
```diff
@@ -184,76 +184,129 @@ def test_error_model_with_metadata_schema(snapshot: SnapshotAssertion) -> None:
...
+
+def test_merge_of_empty_metadata() -> None:
+    """Ensure that members without metadata merge into nothing."""
+    assert ResponseSpecMetadata.merge(None, None) is None
+
+
+class _OptionalUnionHeaderController(Controller[PydanticSerializer]):
+    @validate(ResponseSpec(_AnnotatedMember | str, status_code=HTTPStatus.OK))
+    def get(self) -> HttpResponse:
+        return self.to_response('a string without the header')
```
sobolevn, line 293: This test must return both: `str` without a header and `_AnnotatedMember` with the header dependending on a query parameter, for example. Currently we are only testing one possible path.

sobolevn:
Please, resolve merge conflicts :)

docs/pages/components/index.rst
vyhuholl: Removed

tests/test_unit/test_metadata/test_response_spec_meta.py
vyhuholl: Split `test_union_member_headers_are_optional` into `test_union_member_with_header` and `test_union_member_without_header`.

docs/pages/error-handling.rst
vyhuholl: Added a warning about this.

sobolevn (approved):

PR #1532, opened 2026-09-15
Allow bool objects in dmr.openapi.objects accept None #1437

### LINKED ISSUE #1437: Make sure all `bool` values in `dmr.openapi.objects` can also accept `None`

max0nT:
# I have made things!

## AI Policy

## Checklist

## Related issues
 - Closes #1437


sobolevn (approved):
Great stuff 🎉

PR #1539, opened 2026-09-15
Removed django_db from schemathesis.

s3rius:
Possible speedup of schemathesis execution. 

This PR removes `django_db(transaction=True)` from schemathsis. Although schemathesis runs against real endpoints and probably needs it, tests should pass just fine. Let's test it on github runners and see if it works.


sobolevn (approved):
This is faster, thanks!

PR #1524, opened 2026-09-14
docs: improve more docs pages

mschoettle:
# I have made things!

I am in the process of reading the docs to learn about DMR and investigate whether this is the right API framework for us.

As I am reading I noticed some things (mostly missing articles) and thought I'll contribute back.

Thanks for creating this framework!

## AI Policy

- [x] I have read and agree to the [AI Policy](https://github.com/wemake-services/django-modern-rest/blob/master/.github/AI_POLICY.md), removed any "Co-Authored-By" lines attributing coding agents, and manually reviewed the final result

## Checklist

- [x] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)
- [ ] I have created at least one test case for the changes I have made
- [x] I have updated the documentation for the changes I have made

## Related issues

Follow-up to #1416

PR #1540, opened 2026-09-15
docs: restructure using-controller to controllers

mschoettle:
# I have made things!

Follow-up from this comment to be consistent with `components/` (and plural): https://github.com/wemake-services/django-modern-rest/pull/1524#issuecomment-5671484557

## AI Policy

## Checklist


sobolevn (commented):
Oh, now I get what you were talking about :(
This will break URLs for other people.

mschoettle:
RTD has a redirects feature that we could use: https://docs.readthedocs.com/platform/latest/user-defined-redirects.html#user-defined-redirects

Unless you are generally against changing it, that's fine too :)

sobolevn:
I think that we can keep this as-is :) 
But, thanks for the suggestion! 🤝

PR #1542, opened 2026-09-16
Add concrete views for auth

### LINKED ISSUE #1457: Provide concrete views
Currently we have a lot of reusable views for auth.
But, this is not really cool to do the manual stuff. Even if it take just several lines. Especially, when I don't need any customizations.

I propose providing new folders near the existing `views/`, let's call it `concrete_views/` and we will put all the same views with default implementations there.

This way people can just import and use them with no manual work.

This would require a bit of integration testing :) 
We would need to test default implementations the same way we test reusable views now.
And we should highlight these new views in the docs. And suggest using them over the more complex reusable ones.

vyhuholl:
# I have made things!

Adds `concrete_views` next to `views` for all three auth flows: `dmr/security/jwt/concrete_views/`, `dmr/security/token/concrete_views.py` and `dmr/security/django_session/concrete_views.py`.

We already ship the default bodies — `ObtainTokensPayload`, `ObtainTokensResponse`, `ObtainTokenPayload`, `DjangoSessionPayload` and the rest — but nothing was ever wired to them, so every project writes the same `convert_auth_payload` and `make_api_response` by hand. See `docs/examples/auth/jwt/jwt_obtain_tokens.py` for what that costs today. The new modules do the wiring. Same settings, same hooks, bodies filled in:

```python
class ObtainTokens(concrete_views.ObtainTokensSyncController[PydanticSerializer]):
    """Authenticates a user and issues both tokens."""
```

I kept the class names identical to the ones in `views`, the way the issue described it. The module path is what tells them apart.

Two settings still have to be set, because neither can be guessed:

- `token_cls` on the opaque token views, only the project knows which model holds its tokens
- `jwt_refresh_cookie_path` on the jwt cookie views, it has no default on purpose and `refresh_cookie_spec` raises without it

`CookieRefreshTokens*` and `CookieLogout*` need no defaults at all, so `jwt.concrete_views` re-exports them as-is: those names are the same objects as in `jwt.views`. I did that so a whole cookie flow comes from one import instead of two. Say if you'd rather they weren't there.

Tested through the test app the same way the reusable views are: fourteen new controllers routed in `django_test_app`, 80 cases across three new integration modules, covering obtain, refresh, verify, the cookie flow, the default and a custom `token_cls`, and the session login.

Each of the three auth pages gets a "Ready-to-use views" section above its reusable-controller section, with a tip pointing there first, plus API reference entries and four runnable examples. Two `slotscheck` `exclude-modules` entries in `pyproject.toml`. Importing these modules touches Django settings, same as the `views` modules already excluded there. CHANGELOG updated.

## AI Policy

## Checklist

## Related issues

Closes #1457


django_test_app/server/apps/django_session_auth/views.py
```diff
@@ -63,6 +64,23 @@ async def make_api_response(self) -> DjangoSessionResponse:
         return {_USER_ID: str((await self.request.auser()).pk)}
 
 
+# Concrete views, they are used as-is, without any customizations at all:
+
+
+@final
+class ConcreteSessionSyncController(
+    concrete_views.DjangoSessionSyncController[PydanticSerializer],
```
sobolevn: I was think about maybe some simplier API here?
I really don't want to write a new class.

Options?
1. Serializer with the default type var?
2. Some method to `DjangoSessionSyncController.as_view(serializer=PydanticSerializer)`?
3. Other ideas?

django_test_app/server/apps/token_auth/views/obtain.py
```diff
@@ -72,6 +72,27 @@ async def make_api_response(self) -> ObtainTokenResponse:
...
+    token_cls = CustomToken
+
+
+@final
+class ConcreteObtainTokenAsyncController(
+    concrete_views.ObtainTokenAsyncController[PydanticFastSerializer],
+):
+    """Concrete view to issue an opaque token."""
+
+    token_cls = Token
```
sobolevn: They should have all the default values. Including `Token` as the class.

django_test_app/server/apps/jwt_auth/views.py
```diff
@@ -357,3 +358,91 @@ class CookieLogoutSyncView(CookieLogoutSyncController[PydanticSerializer]):
...
+    jwt_refresh_cookie_path = _CONCRETE_REFRESH_COOKIE_PATH
+
+
+@final
+class ConcreteCookieRefreshSyncController(
+    concrete_views.CookieRefreshTokensSyncController[PydanticSerializer],
+):
+    """Concrete view to rotate both token cookies."""
+
+    jwt_refresh_cookie_path = _CONCRETE_REFRESH_COOKIE_PATH
```
sobolevn: This value must be optional as well.

docs/pages/auth/django-session.rst
```diff
@@ -48,12 +48,44 @@ Example, how to use the auth class and how to get ``self.request.user``:
...
+Every hook still works the same way.
+
+.. tip::
+
+  Start here. Move to the reusable controllers below
+  only when you need a different request or response body,
+  they are the same classes with the bodies left open.
+
+
 Reusing pre-existing views
```
sobolevn:
```suggestion
Customizing pre-existing views
```

here and in other similar places.

sobolevn (commented):
My whole idea was to enable users to have two lines API:
1. One for import
2. Second for including the view into routing

That's all.
That's why I think that maybe `.as_view()` magic here is actually a good idea?

What do you think? Maybe you have better ideas? :)

django_test_app/server/apps/token_auth/views/obtain.py
vyhuholl: Done. `token_cls` now defaults to the bundled `Token` model.

django_test_app/server/apps/jwt_auth/views.py
vyhuholl: Done, `jwt_refresh_cookie_path` now defaults to `/`.

django_test_app/server/apps/django_session_auth/views.py
vyhuholl: Done. I went with the second option:

```python
path('login/', concrete_views.DjangoSessionSyncController.as_view(
    serializer=PydanticSerializer,
))
```

Why not the TypeVar default. It's the more elegant option and #1525 just landed the machinery for it — but it can't work here. A TypeVar default is a static type, so `dmr` core would have to name `PydanticSerializer` or `MsgspecSerializer`, and the import-linter contract forbids either outside `plugins/`. It also wouldn't answer the actual complaint: `defaults_serializer.py` still writes `class PydanticController(ReusableController)`.

vyhuholl:
For your comment about two lines API — I realized that in my implementation there were three:

```python
from dmr.plugins.pydantic import PydanticSerializer      # ← the extra one
from dmr.security.django_session import concrete_views

path('login/', concrete_views.DjangoSessionSyncController.as_view(serializer=PydanticSerializer))
```

So I fixed it in https://github.com/wemake-services/django-modern-rest/commit/cad47065673cc011eb7b20d218c0b8edca6f1476. The fix: a new `serializer` setting — a project-wide fallback that `as_view` uses when it isn't given one. Named once:

```python
DMR_SETTINGS = {Settings.serializer: PydanticSerializer}   # settings.py, once
```

```python
from dmr.security.django_session import concrete_views     # line 1
path('login/', concrete_views.DjangoSessionSyncController.as_view())   # line 2
```

No default, so nothing changes for anyone who doesn't set it. An explicit `serializer=` still wins, so one route can differ. And it's never consulted for a controller that already has an exact serializer — which is every controller written the normal way, so this cannot silently change existing code.

The precedent is already there: `parsers`, `renderers`, `auth`, `throttling`, `responses` all have project-level defaults that controllers override. The serializer was the odd one out.

dmr/settings.py
```diff
@@ -142,6 +145,9 @@ class SettingsDict(TypedDict, total=False):
 
 #: Default settings for `django-modern-rest`.
 _DEFAULTS: Final[Mapping[str, Any]] = {  # noqa: WPS407
+    # Project-wide serializer for controllers that don't name one,
+    # there's no way for us to guess which plugin a project uses:
+    Settings.serializer: None,
```
sobolevn, line 150: Sorry, this is not what I meant. We can't use any global settings for serializers. I meant some API for providing required fields without explicit subclassing :)

sobolevn (changes requested):

dmr/settings.py
vyhuholl: Fixed in https://github.com/wemake-services/django-modern-rest/pull/1542/changes/4738c9ff761cfc515d54c3e2a3d6039de60fc70f.

vyhuholl:
Removed the global serializer setting in https://github.com/wemake-services/django-modern-rest/commit/4738c9ff761cfc515d54c3e2a3d6039de60fc70f. Instead, every keyword argument of `as_view` is now applied as a class attribute of the subclass it builds, so the required fields are given at the call site:

```python
    path('login/', ObtainTokenSyncController.as_view(
        serializer=PydanticSerializer,
        token_cls=CustomToken,
    ))
```

That covers `serializer`, `token_cls`, `jwt_refresh_cookie_path`, and anything else a controller declares, with one rule instead of three.

dmr/controller.py
```diff
@@ -220,35 +238,66 @@ def __init_subclass__(cls) -> None:
 
     @override
     @classmethod
-    def as_view(cls, **initkwargs: Any) -> Callable[..., HttpResponseBase]:
+    def as_view(cls, **class_attrs: Any) -> Callable[..., HttpResponseBase]:
```
sobolevn, line 241: I don't think that I agree with this design. It has several major problems.

1. Not passing `**initkwargs` to `.as_view()` is a breaking change. Moreover, this can kill some potential mixin class usages. We must not do that. Because we need Django packages compatibility.
2. Typing it as `Any` is a bad idea, we have no type-safety.
3. Having it on all controllers is not great as well, because now we provide a very strange API for common cases. 

What I suggest we do instead: each specific reusable view, for example only in `dmr/security/jwt/concrete_views.py` /others, can define its own `as_view` override. Like so:

```python
@override
@classmethod
def as_view(cls, *, serializer: type[BaseSerializer], jwt_refresh_cookie_path: str | None = None, **initkwargs: Any) -> Callable[..., HttpResponseBase]:
    # Implementation
```

This way:
1. API is not broken
2. It is fully typed
3. Only select few controllers have it. Only concrete views in our case. Nothing else can use it.

sobolevn (commented):
I had to think for some time about the design, this is what I came up with.

dmr/controller.py
vyhuholl: Implemented in https://github.com/wemake-services/django-modern-rest/commit/11bf20dd65cbaeb1cbc078ea225d6f1dd625ac08. Each of the 16 concrete views now overrides `as_view` with only the fields it needs, as typed keyword-only arguments. Everything else goes to Django as `initkwargs`:
Views | Typed parameters
-- | --
jwt body (6), django session (2) | `serializer`
jwt cookie (6) | `serializer`, `jwt_refresh_cookie_path: StrOrPromise \| None`
token (2) | `serializer`, `token_cls: type[TokenLikeSync[_UserT]] \| None` (the async view uses `TokenLikeAsync`)
```python
path('token/', ObtainTokenSyncController.as_view(
    serializer=PydanticSerializer,
    token_cls=CustomToken,
))
```
All 16 overrides build the subclass through one private helper, `dmr/internal/concrete.py`.

One change from your snippet — your snippet had `serializer: type[BaseSerializer]` as required. I made it `type[BaseSerializer] | None = None`. If it's required, subclassing a concrete view stops working: the subclass already has its serializer from its type argument, so `MyView.as_view()` would fail with "missing argument 'serializer'", both at runtime and in the type checkers. The test app and the docs both use subclasses like that.

Leaving it out on the concrete view itself still fails, with the same `EndpointMetadataError` any abstract controller gives. So the only difference is that the missing serializer is caught when the URLs are loaded, not by the type checker.

vyhuholl:
Implemented the new design in https://github.com/wemake-services/django-modern-rest/commit/11bf20dd65cbaeb1cbc078ea225d6f1dd625ac08. Each of the 16 concrete views now overrides `as_view` with only the fields it needs, as typed keyword-only arguments. Everything else goes to Django as `initkwargs`:
Views | Typed parameters
-- | --
jwt body (6), django session (2) | `serializer`
jwt cookie (6) | `serializer`, `jwt_refresh_cookie_path: StrOrPromise \| None`
token (2) | `serializer`, `token_cls: type[TokenLikeSync[_UserT]] \| None` (the async view uses `TokenLikeAsync`)
```python
path('token/', ObtainTokenSyncController.as_view(
    serializer=PydanticSerializer,
    token_cls=CustomToken,
))
```
All 16 overrides build the subclass through one private helper, `dmr/internal/concrete.py`.

One change from your snippet — your snippet had `serializer: type[BaseSerializer]` as required. I made it `type[BaseSerializer] | None = None`. If it's required, subclassing a concrete view stops working: the subclass already has its serializer from its type argument, so `MyView.as_view()` would fail with "missing argument 'serializer'", both at runtime and in the type checkers. The test app and the docs both use subclasses like that.

Leaving it out on the concrete view itself still fails, with the same `EndpointMetadataError` any abstract controller gives. So the only difference is that the missing serializer is caught when the URLs are loaded, not by the type checker.

dmr/internal/concrete.py
```diff
@@ -0,0 +1,67 @@
...
+            f'{controller_cls!r} already has {existing_serializer!r} '
+            'as its serializer, passing `serializer=` would contradict '
+            'the type arguments of this controller. '
+            'Drop the argument, or pass it to the concrete view instead',
+        )
+
+    # The new class is a subclass of *controller_cls* by construction,
+    # but type checkers cannot follow `type()` with a type var base:
+    base_cls: type[Any] = controller_cls
+    return cast(
```
sobolevn: please, replace `cast` with `# type: ignore` comment

dmr/security/django_session/concrete_views.py
```diff
@@ -31,12 +36,14 @@ class DjangoSessionSyncController(
...
     :class:`~dmr.security.django_session.views.DjangoSessionResponse`.
-    The only thing it needs is a serializer type:
+    The only thing it needs is a serializer type,
+    pass it to :meth:`as_view` and there is no class to write:
 
     .. code:: python
 
-        class Login(DjangoSessionSyncController[PydanticSerializer]):
-            ...  # nothing else to define
+        path('login/', DjangoSessionSyncController.as_view(
```
sobolevn: Please, make this a doctest

dmr/internal/concrete.py
```diff
@@ -0,0 +1,67 @@
...
+        raise EndpointMetadataError(
+            f'{controller_cls!r} already has {existing_serializer!r} '
+            'as its serializer, passing `serializer=` would contradict '
+            'the type arguments of this controller. '
+            'Drop the argument, or pass it to the concrete view instead',
+        )
+
+    # The new class is a subclass of *controller_cls* by construction,
+    # but type checkers cannot follow `type()` with a type var base:
+    base_cls: type[Any] = controller_cls
```
sobolevn: this is not needed, just use a type ignore

dmr/security/jwt/concrete_views/body.py
```diff
@@ -42,12 +45,14 @@ class ObtainTokensSyncController(
...
     and returns :class:`~dmr.security.jwt.views.ObtainTokensResponse`.
-    The only thing it needs is a serializer type:
+    The only thing it needs is a serializer type,
+    pass it to :meth:`as_view` and there is no class to write:
 
     .. code:: python
 
-        class ObtainTokens(ObtainTokensSyncController[PydanticSerializer]):
-            ...  # nothing else to define
+        path('login/', ObtainTokensSyncController.as_view(
```
sobolevn: and here, a doctest is always better

dmr/security/jwt/concrete_views/cookie.py
```diff
@@ -54,11 +53,13 @@ class CookieObtainTokensSyncController(
 
         .. code:: python
 
-            class Login(CookieObtainTokensSyncController[PydanticSerializer]):
-                jwt_refresh_cookie_path = reverse_lazy('api:refresh')
+            path('login/', CookieObtainTokensSyncController.as_view(
```
sobolevn: and here

tests/test_unit/test_security/test_concrete_views_as_view.py
```diff
@@ -0,0 +1,177 @@
...
+def test_as_view_with_token_cls(
+    view_cls: type[token_views.ObtainTokenSyncController[Any]],
+) -> None:
+    """Ensures that `token_cls` given to `as_view` beats the default."""
+    view = view_cls.as_view(
+        serializer=PydanticSerializer,
+        token_cls=_get_custom_token_model(),
+    )
+
+    assert view.view_class.token_cls is _get_custom_token_model()  # type: ignore[attr-defined]
```
sobolevn: please, always assert the view is not avstract. assert that api_endpoints are not empty.

tests/test_unit/test_security/test_concrete_views_as_view.py
```diff
@@ -0,0 +1,177 @@
...
+    assert routed_cls.__module__ == view_cls.__module__
+    assert routed_cls.__doc__ == view_cls.__doc__
+
+
+@pytest.mark.parametrize('view_cls', _CONCRETE_VIEWS)
+def test_as_view_leaves_the_view_alone(
+    view_cls: _AnyController,
+) -> None:
+    """Ensures that routing a concrete view does not mutate it."""
+    view_cls.as_view(serializer=PydanticSerializer)
```
sobolevn: please, parametrize serializers, test all 3 if possible.

typesafety/assert_type/test_security/test_concrete_views.py
sobolevn: I suggest to remove this file

tests/test_unit/test_security/test_concrete_views_as_view.py
```diff
@@ -0,0 +1,177 @@
...
+) -> None:
+    """Ensures that the bundled model is used when none is given."""
+    view = view_cls.as_view(serializer=PydanticSerializer)
+
+    assert view.view_class.token_cls is Token  # type: ignore[attr-defined]
+
+
+def test_subclass_as_view() -> None:
+    """Ensures that subclasses are routed the usual way."""
+    view = _ConcreteSubclass.as_view()
```
sobolevn: please, define this class just above the test

sobolevn (commented):
This now looks good! Please, also update the docs to state that `concrete_views` should be the default for common cases, but all custom logic should be applied to reusable views.

PR #1543, opened 2026-09-16
Merge request body content types by intersection

### LINKED ISSUE #1537: `_merge_contents` drops request content types depending on component order
### What's wrong?

Found while chasing the TODO in `_merge_contents`:
https://github.com/wemake-services/django-modern-rest/blob/d3ff707e2aa809a3d8ed94a534c81a183b4b2127/dmr/openapi/generators/component_parsers.py#L206-L225

## The TODO is stale

`FileMetadata[]` got conditional types in `0.15.0`. `FileMetadataComponent.conditional_types` is there, `get_schema` builds `conditional_schemas` from it, and `test_conditional_files_schema` plus `tests/test_unit/test_components/test_file_metadata.py` exercise it with `MultiPartParser` + `OctetStreamParser`. So the thing the pragma was waiting for has landed.

Dropping both pragmas still doesn't make coverage green, but not for the reason the TODO assumes. Every statement in the function runs under the current suite; what's missing is a partial branch on each `if` — the *false* arm of both. Conditional `FileMetadata` was never going to fix that, because `_merge_contents` only runs when two components each return a `RequestBody`, and every conditional-file test controller declares `FileMetadata[]` alone.

## The bug behind it

Try to reach the false arm of `if existing_content and existing_content.schema:` and something worse turns up. `_merge_contents` iterates `new_schema.content` only, so any content type in `schema` but not in `new_schema` gets dropped on the floor. Which body is `new_schema` comes from the order the components appear in the signature, so the documented content types change with parameter order:

```python
class _FileFirst(Controller[PydanticSerializer]):
    parsers = (MultiPartParser(), JsonParser())

    def post(
        self,
        parsed_file_metadata: FileMetadata[_Files],
        parsed_body: Body[_Payload],
    ) -> str: ...


class _BodyFirst(Controller[PydanticSerializer]):
    parsers = (MultiPartParser(), JsonParser())

    def post(
        self,
        parsed_body: Body[_Payload],
        parsed_file_metadata: FileMetadata[_Files],
    ) -> str: ...
```

`requestBody.content` for the two:

```text
_FileFirst: ['application/json', 'multipart/form-data']
_BodyFirst: ['multipart/form-data']
```

`Body[]` is documented for every parser in `metadata.parsers`, while `FileMetadata[]` is documented only for the `SupportsFileParsing` ones, so the two bodies disagree about `application/json` and the merge resolves that disagreement by declaration order.

`_BodyFirst` is the output I'd keep. Sending `application/json` to either controller is a 400 at runtime, since `FileMetadataComponent.provide_context_data` rejects a parser that isn't `SupportsFileParsing`:

```json
{"detail": [{"msg": "Trying to parse files with 'JsonParser' that does not support SupportsFileParsing protocol", "type": "value_error"}]}
```

So `_FileFirst` documents a content type that can never succeed, and `_BodyFirst` gets the right answer by accident.

### How it should be?

What I'd like

1. Merge content types by intersection, not by whichever body arrived second, and make the result independent of parameter order.
2. A snapshot test with both orders of the same two components, asserting the same `requestBody.content` keys for each.
3. Then the TODO and the `# pragma: no cover` on line 219 both go.

The other pragma is a separate question. `if media_type.schema:` on line 214 looks unreachable: `MediaType.__post_init__` rejects `schema` and `item_schema` both being `None`, and `BodyComponent` and `FileMetadataComponent` always pass `schema=`. If that holds, the `if` should be deleted rather than covered — worth confirming before someone writes a test for a branch that can't run.

Note the typo while you're in there: `# pragma: no cover:` on line 214 has a
trailing colon.

### Used versions

0.15.0

### OS information

macOS

vyhuholl:
# I have made things!

Makes the merged `requestBody` stop depending on the order body components are declared in.

`_merge_contents` iterated `new_schema.content` only, so a content type the other body had and this one didn't was dropped on the floor. Which body is `new_schema` comes from the order the components appear in the endpoint signature, so `Body[]` and `FileMetadata[]` on the same endpoint documented different content types depending on which of them came first. With `parsers = (MultiPartParser(), JsonParser())`, putting `FileMetadata[]` first documented `application/json`, putting `Body[]` first did not.

The disagreement itself is expected: `Body[]` is documented for every parser in `metadata.parsers`, `FileMetadata[]` only for the `SupportsFileParsing` ones. What the merge did with it was not. A request is parsed by every body component at once, so a content type that only one of them supports can never succeed — `application/json` on such an endpoint is a 400 from `FileMetadataComponent.provide_context_data`, which rejects `JsonParser`. So we keep the intersection now, and neither order documents it anymore.

Fixing the keys uncovered the same bug one level down. `dataclasses.replace(media_type, ...)` built the merged media type out of whichever body came second, so `encoding` — which only `FileMetadata[]` sets — disappeared whenever `Body[]` was the one declared last.

I've made three changes to `_merge_contents`:
- content types are the intersection of both bodies, not the keys of the second one
- `encoding` and `examples` are merged from both sides. They are keyed by property name and describe different parts of the same body, so they don't collide
- `allOf` members follow declaration order, they used to be reversed

Both `# pragma: no cover` and the TODO above them are gone, including the trailing colon in `# pragma: no cover:`. `if media_type.schema:` turned out to be unreachable, so it's an `assert` now with a comment saying why: both body components always pass `schema=`, `item_schema` is for streaming responses. The file measures 68 statements and 22 branches with nothing missed and nothing partial.

Two new tests instead of the snapshot the issue asks for: `test_merged_body_content_types_ignore_order` and `test_merged_body_keeps_encoding_in_any_order`, both over a pair of controllers that differ only in parameter order. They state the property directly, and a third full-document snapshot next to `test_merged_body_without_description_schema` would mostly repeat it. One existing snapshot did change: `test_body_and_file_schema` now lists `_DescriptionModel` before `_SeveralFiles` in `allOf`, which is the new ordering.

One thing I left alone: if two body components both set a plain `example` on the same content type, one still overwrites the other. Nothing in the suite does that, so I had no way to test a rule for it. CHANGELOG updated.

## AI Policy

## Checklist

## Related issues

Closes #1537


tests/test_unit/test_openapi/test_files_snapshots.py
```diff
@@ -313,3 +313,59 @@ def test_merged_body_without_description_schema(
         )
         == snapshot
     )
+
+
+class _FileFirstController(Controller[PydanticSerializer]):
```
sobolevn, line 318: This is not true. 

Currently we would require `parsed_file_metadata` for all requests. Including `json` ones.

Let's first wait for #1494

sobolevn (changes requested):
I propose closing this for now.

PR #1544, opened 2026-09-16
Cache the negotiation results per header value

### LINKED ISSUE #1455: Cache the negotiation parts

vyhuholl:
# I have made things!

`dmr/negotiation.py` — both negotiators now memoize their decision per header value with `lru_cache(maxsize=MAX_CACHE_SIZE)`, created per instance in `__init__`:

* `RequestNegotiator._decide` now takes `content_type: str | None` instead of the request, and returns `None` instead of raising. `__call__` raises `RequestSerializationError`.
* `ResponseNegotiator` gets two caches: `_negotiate` and `_negotiate_non_streaming`. The streaming `try/except NotAcceptableError` fallback became a plain `is None` check.

`dmr/internal/negotiation.py` — `negotiate_renderer` split into pure `find_renderer(accept, renderers, default) -> Renderer | None` (cacheable), `not_acceptable_error(...)` (builds the error), and a thin request-based `negotiate_renderer` wrapper so `dmr/routing.py`'s 404/500 handlers are untouched.

Only the *decision* is cached, never the exception — exceptions keep tracebacks alive, and `NotAcceptableError`'s message quotes `request.accepted_types` of that exact request. No public API changed.

## Measured (200k iterations, best of 5)

case | master | branch
-- | -- | --
Accept: application/json | 0.0192s | 0.0158s
browser Accept (7 types, q-values) | 1.0984s | 0.0159s (x69)
Content-Type: application/json | 0.0280s | 0.0242s
Content-Type hitting application/* | 0.3827s | 0.0246s (x15)

Also:
* `tests/test_unit/test_negotiation/test_negotiation_cache.py` — 5 tests: hits/misses per header, unsupported headers still raise on every request, caches not shared between endpoints.
* `benchmarks/tests/test_negotiation/test_negotiators.py` — codspeed benchmarks for both `Accept` shapes.
* CHANGELOG + docs (`negotiation.rst`, `configuration.rst` under `DMR_MAX_CACHE_SIZE`).

## AI Policy

## Checklist

## Related issues

Closes #1455


dmr/negotiation.py
```diff
@@ -59,6 +65,9 @@ def __init__(
         self._media_by_precedence = media_by_precedence(self._parsers.keys())
         # The last configured parser is the most specific one:
         self._default = next(iter(self._parsers.values()))
+        # Almost every client sends the very same `Content-Type` header
+        # over and over, so we only decide once per header value:
+        self._negotiate = lru_cache(maxsize=MAX_CACHE_SIZE)(self._decide)
```
sobolevn: This is not correct. This will keep `self` in `lru_cache` alive.

dmr/negotiation.py
```diff
@@ -166,6 +186,13 @@ def __init__(
             iter(self._non_streaming_renderers.values()),
         )
 
+        # Almost every client sends the very same `Accept` header
+        # over and over, so we only negotiate once per header value:
+        self._negotiate = lru_cache(maxsize=MAX_CACHE_SIZE)(self._decide)
```
sobolevn: same problem here.

sobolevn (commented):
really amazing numbers, but we have to fix the `self` bug :)

sobolevn (commented):
Btw, don't forget to add `clear_cache` method to negotiators. Otherwise, it would be impossible to fix their state.

dmr/negotiation.py
sobolevn: Proof:

```python
>>> from functools import lru_cache
>>> import sys
>>> class Some:
...     def __init__(self):
...         self.cached = lru_cache()(self.regular)
...     def regular(self, arg):
...         print(self, arg)
...     def __del__(self):
...         print('__del__')
...         
>>> s = Some()
>>> sys.getrefcount(s)  # should return 2, but returns 3 :(
3
>>> del s
```

Compare it with the correct way:

```python
>>> class Regular:
...     def __init__(self):
...         self.a = 1
...     def __del__(self):
...         print('__del__')
...         
>>> r = Regular()
>>> sys.getrefcount(r)
2
>>> del r
__del__
```

vyhuholl: Fixed in https://github.com/wemake-services/django-modern-rest/commit/d9c113d9af55ec84b8992ded9bd074ff6844e61b. The fix — bind the *state*, not `self`. The parser lookup moved out of `RequestNegotiator._decide` into `find_parser` in `dmr/internal/negotiation.py`, next to the existing `find_renderer`, and both are now bound with `functools.partial`:

```python
self._negotiate = lru_cache(maxsize=MAX_CACHE_SIZE)(
    partial(
        _find_parser,
        parsers=self._parsers,
        exact_parsers=self._exact_parsers,
        media_by_precedence=self._media_by_precedence,
        default=self._default,
    ),
)
```

The partial holds the parser/renderer dicts and defaults — none of which reference the negotiator back — so there is no cycle. `ResponseNegotiator`'s two caches get the same treatment over `find_renderer`. The `_decide` / `_decide_non_streaming` methods are gone; both `find_*` functions are keyword-only past the header value and document why.

After the fix, both `RequestNegotiator` and `ResponseNegotiator` report refcount of 2, the same as a bare `object()` local. Also, added `test_negotiators_are_not_self_referencing` to prevent this bug from happening again.

Speed re-measured against pre-caching https://github.com/wemake-services/django-modern-rest/commit/2defe58a61678656ba9aa8df16f723050f16dfe1, with almost exactly the same results (browser `Accept` speed improvement move down from x69 to x65 — corrected the CHANGELOG accordingly).

dmr/negotiation.py
vyhuholl: Again, fixed in https://github.com/wemake-services/django-modern-rest/commit/d9c113d9af55ec84b8992ded9bd074ff6844e61b

tests/test_unit/test_negotiation/test_negotiation_cache.py
sobolevn, line 1: This is testing an implementation detail that is not really important. Just test:
1. Call negotiation once
2. Call it the second time, assert result is the same
3. Clear the cache
4. Call it the last time, assert result is the same

Do it for both negotiators.

vyhuholl: Done in https://github.com/wemake-services/django-modern-rest/pull/1544/changes/b9af98fcfe306608f1020aaa959997399a04340f. The file went from 196 lines to 22: two tests, one per negotiator, each following this exact sequence:
```python
first = negotiator(dmr_rf.get('/whatever/', headers=headers))
second = negotiator(dmr_rf.get('/whatever/', headers=headers))

negotiator.clear_cache()
third = negotiator(dmr_rf.get('/whatever/', headers=headers))

assert first.content_type == _JSON
assert second is first
assert third is first

```

dmr/negotiation.py
```diff
@@ -214,6 +256,17 @@ def __call__(self, request: HttpRequest) -> Renderer:
...
         return renderer
 
+    def clear_cache(self) -> None:
+        """
+        Drop everything this negotiator has memoized so far.
+
+        Renderers are fixed for an endpoint in import time,
+        so this is only needed when they are modified in place:
+        in tests or in some very dynamic setups.
+        """
```
sobolevn, line 306:
```suggestion
        .. versionadded:: 0.16.0
        """
```

dmr/negotiation.py
```diff
@@ -84,34 +104,30 @@ def __call__(self, request: HttpRequest) -> Parser:
...
-            ),
-        )
+    def clear_cache(self) -> None:
+        """
+        Drop everything this negotiator has memoized so far.
+
+        Parsers are fixed for an endpoint in import time,
+        so this is only needed when they are modified in place:
+        in tests or in some very dynamic setups.
+        """
```
sobolevn, line 151:
```suggestion
        .. versionadded:: 0.16.0
        """
```

setup.cfg
```diff
@@ -52,6 +52,7 @@ per-file-ignores =
   dmr/routing.py: WPS114, WPS201
   dmr/problem_details.py: WPS211, WPS226
   dmr/negotiation.py: WPS202
+  dmr/internal/negotiation.py: WPS202
```
sobolevn: You can move `media_by_precedence` to `media_compat`. Will it help to remove this error?

sobolevn (commented):
Also: please, don't forget to describe the caching change in `RequestNegotiator` and `ResponseNegotiator` docstrings.

benchmarks/tests/test_negotiation/test_negotiators.py
```diff
@@ -0,0 +1,53 @@
...
+def test_response_negotiation_api_client(
+    benchmark: BenchmarkFixture,
+    dmr_rf: DMRRequestFactory,
+) -> None:
+    """Benchmark the `Accept` header that hits the exact match."""
+    negotiator = _NegotiatedController.api_endpoints['GET'].response_negotiator
+    request = dmr_rf.get('/test', headers={'Accept': _API_CLIENT_ACCEPT})
+
+    def factory() -> Renderer:
+        return negotiator(request)
```
sobolevn: Without `_REPEAT` it will fluke a lot. Just like the second one.

setup.cfg
vyhuholl: Moved.

benchmarks/tests/test_negotiation/test_negotiators.py
vyhuholl: Added `_REPEAT: Final = 1000` with the loop inside `factory()`.

sobolevn (approved):
Thanks!

PR #1549, opened 2026-09-17
Seed the example factory once per schema build

### LINKED ISSUE #1548: Every generated OpenAPI example of the same type gets the same value
### What's wrong?

`generate_example` reseeds the factory before every call, so two schemas of the same type always get the same example: https://github.com/wemake-services/django-modern-rest/blob/cd223ab66feb594c81639a815b7bc34d74bb883a/dmr/openapi/mappers/example.py#L41-L48

```python
if _ExampleFactory.__random_seed__ != seed:  # pragma: no cover
    # Reseed the factory, if it is required.
    _ExampleFactory.seed_random(seed)
```

The guard never becomes false. `polyfactory` sets `__random__` and seeds faker, but it never writes back `__random_seed__`:

```python
@classmethod
def seed_random(cls, seed: int) -> None:
    cls.__random__ = Random(seed)
    cls.__faker__.seed_instance(seed)
```

So `_ExampleFactory.__random_seed__` keeps the `10` it is declared with on [L27](https://github.com/wemake-services/django-modern-rest/blob/cd223ab66feb594c81639a815b7bc34d74bb883a/dmr/openapi/mappers/example.py#L27), the condition holds for every seed except `10` itself, and the random stream restarts before each value.

Three `str` calls per seed, one test run, seed set through the `settings` fixture:

```text
seed=5   ['GMPXMVbyXHUfymCDaloV', 'GMPXMVbyXHUfymCDaloV', 'GMPXMVbyXHUfymCDaloV']
seed=10  ['NiPhGgaTklZTpXCJkWJY', 'UpsihdpFaJrqQyqzKcYb', 'MRaOtEacjXkNWWrsBOfM']
```

The first row reseeds before each value, so all three come out equal. The second row ran after it: the guard is false at `10`, nothing reseeds, and the three calls simply continue the `Random(5)` stream the first row left behind — its values 2, 3 and 4. So whether examples repeat depends on whether the configured seed happens to match an internal default, and when it does, on whatever ran before.

Values do still vary *inside* one generated model, because that is a single `get_field_value` call. Across models they don't, which is what you see in `tests/test_unit/test_openapi/__snapshots__/test_example_snapshots.ambr`: `GMPXMVbyXHUfymCDaloV` appears four times there, as `msg` in three separate `ErrorModel` examples and as a dict key in the fourth.

It's cosmetic — the examples are valid, just repetitive, and identical strings across a document make the generated docs look broken.

### How it should be?

Distinct values for distinct schemas, and the same document for the same seed no matter what ran before it.

Seeding once and letting the stream run gives the first half but loses the second. The values then depend on how many calls came earlier — after two unrelated calls the same sequence starts at its third value:

```text
seed once, then 4 calls      ['GMPXMVbyXHUfymCDaloV', 'NiPhGgaTklZTpXCJkWJY', 'UpsihdpFaJrqQyqzKcYb', 'MRaOtEacjXkNWWrsBOfM']
after 2 unrelated calls      ['UpsihdpFaJrqQyqzKcYb', 'MRaOtEacjXkNWWrsBOfM', 'PSbXerFVrWCqqjehJZic', 'ZBvmEQxvcVbzRgMXGOfw']
```

`_ExampleFactory` is class-level state shared by the whole test session, so that alone would make every example snapshot depend on test order, and we run with `pytest-randomly`. Reseeding once per `build_schema` instead of once per call keeps a document reproducible on its own. `generate_example` has no access to the context today, so this needs a decision about where the seeding belongs — probably `SchemaGenerator`, which already holds `_context`.

Whatever the fix, it rewrites every generated example in `test_example_snapshots.ambr` and `test_msgspec_snapshots.ambr`, the only two snapshots holding generated values today, so it wants its own PR.

Two smaller things in the same three lines:

1. The `# pragma: no cover` on the guard is wrong. That branch runs on every call for any seed other than `10`; it is the *false* arm that never runs.
2. `__random_seed__ = 10  # just a random number` is load-bearing now, which the comment doesn't suggest. Whatever replaces this shouldn't leave a `polyfactory` attribute deciding our behaviour.

### Used versions

`django-modern-rest` 0.15.0, `polyfactory` 3.3.0, `faker` 40.39.0.

Reproduced on https://github.com/wemake-services/django-modern-rest/commit/cd223ab66feb594c81639a815b7bc34d74bb883a.

### OS information

macOS 27.0

vyhuholl:
# I have made things!

Seeds the example factory once per schema build, so generated examples stop repeating.

The issue assumed `__random_seed__` was a cache of the seed we last used. It isn't. `polyfactory` reads it once in `BaseFactory.__init_subclass__` and calls `seed_random` right there, so it's a way to seed a factory at class creation and nothing else. `seed_random` never writes it back, which is why `__random_seed__ != seed` in `generate_example` stayed true for every seed except the `10` we declared: the factory was reseeded before each example, the random stream restarted every time, and every schema of the same type got that stream's first value.

Seeding once was the intent all along, then. The declaration just can't reach a seed that comes from settings.

I've made three changes:
- `__random_seed__` is gone. It can't do this job, and keeping it leaves a `polyfactory` attribute deciding our behaviour
- `seed_examples` in `dmr/openapi/mappers/example.py` seeds from settings
- `OpenAPIContext.__init__` calls it. One context builds one schema, which the docs already say, so that is exactly one seeding per document

`generate_example` only checks whether generation is enabled now, and the `# pragma: no cover` went with the branch it sat on.

Every generated value changed — seven of them, across `test_example_snapshots.ambr` and `test_msgspec_snapshots.ambr`. One `loc` entry went from a string to an int, which is expected: `loc` is `list[int | str]`, and a different position in the stream picks a different member of that union.

Examples still repeat between documents, because every build reseeds and each one starts from the same place. `GMPXMVbyXHUfymCDaloV` is still in `test_example_snapshots.ambr` twice — but under two different snapshot entries now, instead of three times inside one. A real app builds one schema, so per-document reproducibility is the half that matters.

New tests in `tests/test_unit/test_openapi/test_example_seeding.py`: that fields of the same type get different values, that a schema comes out the same after three unrelated `generate_example` calls have moved the shared stream, and that nothing is generated without a seed. The middle one is the one worth keeping — the factory is process-wide state, and without per-build seeding a schema's values shift with whatever generated before it.

CHANGELOG updated.

## AI Policy

## Checklist

## Related issues

Closes #1548


sobolevn:
please, resolve merge conflicts

dmr/openapi/core/context.py
```diff
@@ -92,6 +93,10 @@ def __init__(
         """Initialize the OpenAPI context."""
         from dmr.openapi.config import default_config  # noqa: PLC0415
 
+        # One context builds one schema, so this seeds the examples
+        # of that schema exactly once:
+        seed_examples()
```
sobolevn: this should be a method, so we can customize it in subclasses.

vyhuholl: Done in https://github.com/wemake-services/django-modern-rest/commit/d5eaf521cd228b5662506191385ccc2b569a5615. `seed_examples` is now a context method. The mapper-level function is renamed `seed_example_factory` so the short name belongs to the hook:

```python
def seed_examples(self) -> None:
    """
    Seed the generation of examples for this schema.
    ...
    Override it to seed from something other than
    :data:`~dmr.settings.Settings.openapi_examples_seed`,
    or to leave the factory alone entirely.
    """
    seed_example_factory()
```

I moved the call to the end of `__init__` rather than the start. It used to run first, before `self.config` was even assigned — an override that wanted to read self.config for a per-document seed would have hit an uninitialised attribute. Seeding only has to happen before the first example is generated, which is during `router.get_schema(context)`, well after `__init__` returns. The class docstring gets a `versionchanged` for the new hook.

sobolevn (approved):

sobolevn:
Great catch!

PR #1502, opened 2026-09-13
Fix nested URL path parameters

kanagarajSCK:
Closes #1496

This change adds a regression test for nested URL patterns and preserves parent route converters when generating OpenAPI path parameters.

Tests:
- `pytest -q -o addopts="" tests/test_unit/test_openapi/test_collector.py`
- `pytest -q -o addopts="" tests/test_unit/test_openapi tests/test_unit/test_routing`

Results:
- 40 passed
- 290 passed, 1 skipped
- 34 snapshots passed

PR #1564, opened 2026-09-19
Fix nested URL patterns producing incomplete path parameters

### LINKED ISSUE #1496: Test that nested URL patterns produce correct parameters
We need to test a case like:

`Router('tenants/<int:tenant_id>/', [path('users/<int:pk>/', MyController.as_view())])`

It might contain a bug: we should get two path parameters, not one.

krisrajaryan27:
## Summary

Fixes #1496

When URL patterns are nested (e.g., a `Router` prefix with converters or a `URLResolver` wrapping a `URLPattern`), the OpenAPI schema only included path parameters from the leaf pattern, ignoring parent converters.

For example, `Router('tenants/<int:tenant_id>/', [path('users/<int:pk>/', C.as_view())])` only produced `pk` in the spec — `tenant_id` was missing.

### Changes

**`dmr/openapi/collector.py`**:
- `controller_mapping_collector` now accumulates parent resolver patterns as it recurses into `URLResolver`s
- `_merge_parent_patterns` builds a combined `URLPattern` with all converters from every nesting level
- `_prefix_patterns` handles the case where a `Router` prefix itself contains converter syntax
- When all parents and child use `path()` (`RoutePattern`), converters are preserved (e.g., `int` → `integer` in schema)
- When any pattern uses `re_path()` (`RegexPattern`), falls back to regex-based combination

**`tests/test_unit/test_openapi/test_schema_snapshots.py`** — 5 new snapshot tests:
- `test_nested_path_path_schema` — `path` parent + `path` child
- `test_nested_re_path_path_schema` — `re_path` parent + `path` child
- `test_nested_path_re_path_schema` — `path` parent + `re_path` child
- `test_nested_re_path_re_path_schema` — `re_path` parent + `re_path` child
- `test_router_prefix_nested_path_schema` — `Router` prefix with converter + `path` child (the exact case from the issue)

### Test plan

- [x] All 2501 unit tests pass (2496 existing + 5 new)
- [x] All 69 snapshot tests pass
- [x] Existing `test_raw_path_schema` snapshot unchanged
- [x] Verified both `tenant_id` and `pk` appear as path parameters in all nested test cases


sobolevn (approved):
Thanks! I've merged https://github.com/wemake-services/django-modern-rest/pull/1502 because it was first, but I forgot to add a `CHANGELOG.md` entry :)

So, I added one here.

Please, feel free to contribute more!

PR #1565, opened 2026-09-19
Improve AI skills

sobolevn:
This is a complete slop. I will review it later.


dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
```diff
@@ -0,0 +1,314 @@
+#!/usr/bin/env python3
```
vyhuholl: `just lint` fails on this branch right now. Ruff reports `EXE001` — the script has a shebang but mode `100644`:
```
EXE001 Shebang is present but file is not executable
 --> dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py:1:1
```
Either `chmod +x` (and commit the mode) or drop the shebang.

dmr/.agents/skills/dmr-upgrade/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+  or the pinned version in `uv.lock`, `poetry.lock`, or `requirements.txt`.
+- Target: the version the user asked for, otherwise the latest release
+  on https://pypi.org/project/django-modern-rest/.
+- Tell the user which releases lie in between and that each one
+  is applied in order. Never skip a release.
+
+### 2. Bump the dependency
+
+Update `pyproject.toml` (or the requirements file) to the target version
+and install it, so the codemod can detect the version
```
vyhuholl: The most important blocker — **the documented upgrade workflow now silently does nothing**. `SKILL.md:33-36` says bump the dependency and install it *first* "so the codemod can detect the version", then step 3 runs the script without --current. But `detect_current_version()` in `dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py` reads the installed version — which is now the target — so `select_releases` returns `[]` and it prints `Nothing to do`. `docs/pages/ai/dmr-upgrade.rst:24` repeats the same order. Detect before the bump, or read the lock file, or make `--current` required.

dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
```diff
@@ -0,0 +1,314 @@
...
+    prompt: str
+    symbols: list[SymbolRename]
+    keywords: list[KeywordRename]
+    removed: list[str]
+    manual: list[ManualCheck]
+
+
+def parse_version(text: str) -> tuple[int, ...]:
+    """Convert ``'0.16.0'`` into a comparable tuple."""
+    return tuple(int(part) for part in text.split('.'))
```
vyhuholl: Version parsing crashes on any PEP 440 version:
```
$ dmr_upgrade.py --current 0.16.0rc1 --target 0.16.0 .
ValueError: invalid literal for int() with base 10: '0rc1'
```
We ship `0.16.0 WIP`; anyone on an `rc`/`.dev` install hits this via `detect_current_version()`. Related: `--target 0.16` silently selects zero releases, because `(0, 16, 0) <= (0, 16)` is `False`.

dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
```diff
@@ -0,0 +1,314 @@
...
+    dry_run: bool,
+) -> tuple[bool, int]:
+    """Rewrite one file across all releases, return (changed, manual_hits)."""
+    original = path.read_text(encoding='utf-8')
+    source = original
+    manual_hits = 0
+    for release in releases:
+        if release['symbols'] or release['keywords']:
+            source = rewrite_module(source, str(path), release)
+        manual_hits += report_manual_checks(path, source, release)
```
vyhuholl: One unparsable file aborts the run mid-write — `cst.parse_module` raises a raw `ParserSyntaxError` traceback, and files already processed are already written. Real projects have at least one file libcst chokes on (fixtures, newer syntax). Catch per file, report, continue.

.readthedocs.yml
```diff
@@ -4,6 +4,13 @@ build:
   os: ubuntu-lts-latest
   tools:
     python: "3.12"
+  jobs:
+    post_build:
+      # Markdown twin of every page for LLMs and agents, served next
+      # to the HTML page: `pages/routing.md` for `pages/routing.html`.
+      # Reuses the doctrees of the HTML build, so examples are not re-run.
+      - cd docs && uv run python -m sphinx -M markdown . _build -W
```
vyhuholl:
```suggestion
      - cd docs && uv run --group docs --all-extras --frozen python -m sphinx -M markdown . _build -W
```

vyhuholl:
Reviewed the whole diff and ran the new code (codemod, tests, ruff, `agentskills` validate, wheel build). Structural direction is sound — skills in the package, validator + unit tests + evals, docs Markdown twins. The problems are concentrated in the codemod and in a few CI/doc details.

tests/test_unit/test_skills/test_skills.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@pytest.mark.parametrize('skill_dir', _SKILLS, ids=lambda path: path.name)
+def test_skill_local_links(skill_dir: Path) -> None:
+    """Every referenced local file exists, progressive disclosure works."""
+    markdown_files = [skill_dir / 'SKILL.md']
+    markdown_files.extend((skill_dir / 'references').glob('*.md'))
+
+    for markdown_file in markdown_files:
+        for link in _LOCAL_LINK.findall(markdown_file.read_text()):
```
vyhuholl:
```suggestion
        for link in _LOCAL_LINK.findall(markdown_file.read_text(encoding='utf-8')):
```

vyhuholl: Without explicit encoding argument the Windows CI job [is failing](https://github.com/wemake-services/django-modern-rest/actions/runs/35459111644/job/105939664239?pr=1565).

The same — for the other 7 `read_text()` calls in that file. Fixing only this one turns CI green but leaves the rest broken: 12 skill files contain non-ASCII, and `—` / `→` happen to be decodable as cp1252, so those tests keep passing while reading mojibake.

CHANGELOG.md
```diff
@@ -17,9 +17,11 @@ What is a public API for us (all criteria must be met)?
 Later on we will make the API more stable and decrease the amount
 of requirements for an API to count as public.
 
-All migration prompts since `0.13.0` release
-are stored as descriptions in version releases on GitHub, example:
-https://github.com/wemake-services/django-modern-rest/releases/tag/0.13.0
+All migration prompts for breaking releases live in the
+[`dmr-upgrade`](https://github.com/wemake-services/django-modern-rest/tree/master/dmr/.agents/skills/dmr-upgrade) agent skill,
```
vyhuholl, line 21:
```suggestion
[`dmr-upgrade`](dmr/.agents/skills/dmr-upgrade) agent skill,
```

vyhuholl: So lychee CI job won't fail. And the same for other absolute `blob/master`/`tree/master` URLs in the changelog. For the other 2 lychee faliures see https://github.com/wemake-services/django-modern-rest/pull/1565#issuecomment-5794835022 (since lychee config is not in the diff, I cannot directly suggest a change here).

tests/test_unit/test_skills/test_skills.py
```diff
@@ -0,0 +1,136 @@
...
+
+_SKILLS: Final = sorted(
+    path for path in _SKILLS_DIR.iterdir() if (path / 'SKILL.md').is_file()
+)
+
+
+def _split_frontmatter(skill_dir: Path) -> tuple[dict[str, object], str]:
+    _, frontmatter, body = (
+        (skill_dir / 'SKILL.md')
+        .read_text()
```
vyhuholl:
```suggestion
        .read_text(encoding='utf-8')
```

tests/test_unit/test_skills/test_skills.py
```diff
@@ -0,0 +1,136 @@
...
+            assert (skill_dir / link).is_file(), f'{markdown_file}: {link}'
+
+
+@pytest.mark.parametrize('skill_dir', _SKILLS, ids=lambda path: path.name)
+def test_skill_host_metadata(skill_dir: Path) -> None:
+    """Claude Code plugin and Codex metadata match the skill."""
+    plugin = json.loads(
+        (skill_dir / '.claude-plugin' / 'plugin.json').read_text(),
+    )
+    codex = yaml.safe_load((skill_dir / 'agents' / 'openai.yaml').read_text())
```
vyhuholl:
```suggestion
    plugin = json.loads(
        (skill_dir / '.claude-plugin' / 'plugin.json').read_text(
            encoding='utf-8',
        ),
    )
    codex = yaml.safe_load(
        (skill_dir / 'agents' / 'openai.yaml').read_text(encoding='utf-8'),
    )

```

tests/test_unit/test_skills/test_skills.py
```diff
@@ -0,0 +1,136 @@
...
+
+    assert plugin['name'] == skill_dir.name
+    assert plugin['skills'] == ['./']
+    assert codex['interface']['display_name']
+    assert skill_dir.name in codex['interface']['default_prompt']
+
+
+def test_marketplace_lists_every_skill() -> None:
+    """The marketplace manifest points at every shipped skill."""
+    marketplace = json.loads(_MARKETPLACE.read_text())
```
vyhuholl:
```suggestion
    marketplace = json.loads(_MARKETPLACE.read_text(encoding='utf-8'))
```

tests/test_unit/test_skills/test_skills.py
```diff
@@ -0,0 +1,136 @@
...
+        assert (_REPO_ROOT / plugin['source'] / 'SKILL.md').is_file()
+        assert plugin['keywords']
+        assert plugin['license'] == 'MIT'
+
+
+def test_upgrade_prompts_are_indexed() -> None:
+    """Every migration prompt is known to the codemod, in version order."""
+    upgrade_dir = _SKILLS_DIR / 'dmr-upgrade'
+    releases = json.loads(
+        (upgrade_dir / 'references' / 'renames.json').read_text(),
```
vyhuholl:
```suggestion
        (upgrade_dir / 'references' / 'renames.json').read_text(
            encoding='utf-8',
        ),
```

tests/test_unit/test_skills/test_skills.py
```diff
@@ -0,0 +1,136 @@
...
+    ]
+
+    assert versions == sorted(versions)
+    assert prompts == {
+        path.name for path in (upgrade_dir / 'references').glob('*-to-*.md')
+    }
+    for prompt in prompts:
+        assert (
+            (upgrade_dir / 'references' / prompt)
+            .read_text()
```
vyhuholl:
```suggestion
            .read_text(encoding='utf-8')
```

tests/test_unit/test_skills/test_skills.py
```diff
@@ -0,0 +1,136 @@
...
+                '# Upgrade django-modern-rest from',
+            )
+        )
+
+
+def test_docs_mention_every_skill() -> None:
+    """The docs advertise every skill by its ``$name``."""
+    docs = (
+        _REPO_ROOT / 'docs' / 'pages' / 'ai' / 'agent-skills.rst'
+    ).read_text()
```
vyhuholl:
```suggestion
    ).read_text(encoding='utf-8')
```

CHANGELOG.md
```diff
@@ -741,14 +755,8 @@ for both sync and async controllers at the same time.
...
-### Migrations prompt
-
-User-facing changes:
-
-```md
-Change all existing `dmr.files.FileResponseSpec` usages
-to include `as_attachment=True` parameter.
-```
+Migration prompt: see the `dmr-upgrade` agent skill,
+[`references/0.9-to-0.10.md`](https://github.com/wemake-services/django-modern-rest/blob/master/dmr/.agents/skills/dmr-upgrade/references/0.9-to-0.10.md).
```
vyhuholl, line 759:
```suggestion
[`references/0.9-to-0.10.md`](dmr/.agents/skills/dmr-upgrade/references/0.9-to-0.10.md).
```

CHANGELOG.md
```diff
@@ -798,17 +806,8 @@ to include `as_attachment=True` parameter.
...
-
-```md
-Apply this change to the code that uses `django-modern-rest`:
-1. Replace `dmr.response.APIRedirectError` with `dmr.response.RedirectTo`
-2. Replace `dmr.throttling.backend.DjangoCache`
-   with `dmr.throttling.backend.SyncDjangoCache` for sync throttles
-   and with `dmr.throttling.backend.AsyncDjangoCache` for async throttles
-```
+Migration prompt: see the `dmr-upgrade` agent skill,
+[`references/0.7-to-0.8.md`](https://github.com/wemake-services/django-modern-rest/blob/master/dmr/.agents/skills/dmr-upgrade/references/0.7-to-0.8.md).
```
vyhuholl, line 810:
```suggestion
[`references/0.7-to-0.8.md`](dmr/.agents/skills/dmr-upgrade/references/0.7-to-0.8.md).
```

CHANGELOG.md
```diff
@@ -865,20 +864,8 @@ Apply this change to the code that uses `django-modern-rest`:
...
-2. Change `dmr.security.jwt.auth.get_jwt` function
-   to use `dmr.security.jwt.auth.request_jwt` instead, if user expects
-   to always get a token back, add `strict=True` argument
-3. Change `provide_response_specs` class method to be instance method,
-   replace all `cls` usage with `self`
-4. Add `router: Router` parameter to `Endpoint.get_schema`
-   and `Controller.get_path_item` methods
-```
+Migration prompt: see the `dmr-upgrade` agent skill,
+[`references/0.6-to-0.7.md`](https://github.com/wemake-services/django-modern-rest/blob/master/dmr/.agents/skills/dmr-upgrade/references/0.6-to-0.7.md).
```
vyhuholl, line 868:
```suggestion
[`references/0.6-to-0.7.md`](dmr/.agents/skills/dmr-upgrade/references/0.6-to-0.7.md).
```

CHANGELOG.md
```diff
@@ -1017,143 +1004,8 @@ AKA "The first version that I enjoy".
...
-          parsed_headers: HeaderModel,
-      ) -> AsyncIterator[SSEvent[str]]:
-          if parsed_headers.last_event_id is None:
-              yield SSEvent('starting from scratch')
-          else:
-              yield SSEvent(f'starting from {parsed_headers.last_event_id}')
-  ```
-5. Replace old `dmr.sse` imports with new `dmr.streaming.sse` alternatives
+Migration prompt: see the `dmr-upgrade` agent skill,
+[`references/0.3-to-0.4.md`](https://github.com/wemake-services/django-modern-rest/blob/master/dmr/.agents/skills/dmr-upgrade/references/0.3-to-0.4.md).
```
vyhuholl, line 1008:
```suggestion
[`references/0.3-to-0.4.md`](dmr/.agents/skills/dmr-upgrade/references/0.3-to-0.4.md).
```

vyhuholl:
The other 2 `lychee` 404s are the `metadata.homepage` of the two new skills:
```
[404] .../en/latest/pages/ai/dmr-upgrade.html   (dmr-upgrade/SKILL.md:7)
[404] .../en/latest/pages/ai/agent-skills.html  (dmr/SKILL.md:7)
```
Genuinely chicken-and-egg — they exist only after RTD rebuilds `latest`. Either exclude them, or point `homepage` at the repo tree (add this to `exclude` part of `.github/lychee.tom`:
```toml
  "^https://t\\.me/.*$",
  # These docs pages exist only after this branch is merged and deployed:
  "^https://django-modern-rest\\.readthedocs\\.io/en/latest/pages/ai/(agent-skills|dmr-upgrade)\\.html$",
```

dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
```diff
@@ -0,0 +1,314 @@
+#!/usr/bin/env python3
+# /// script
```
vyhuholl, line 2:
```suggestion
# /// script
```

vyhuholl: `just lint` currently fails on this file — shebang is present but the file is not executable. Since PEP 723 metadata works without the shebang, `uv run dmr_upgrade.py` still resolves the deps and no other scripts in the repo contain it, I propose removing it.

dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
vyhuholl:
```suggestion
    """Convert ``'0.16'`` or ``'0.16.0rc1'`` into a comparable tuple."""
    numbers = re.findall(r'\d+', text.split('+')[0])[:3]
    return tuple(int(part) for part in numbers + ['0'] * (3 - len(numbers)))
```

vyhuholl: Currently this crashes on any non-numeric version segment. `detect_current_version()` reads installed metadata, and we ship `0.16.0 WIP`, so anyone on an `rc`/`.dev` install hits a raw traceback:
```
$ dmr_upgrade.py --current 0.16.0rc1 --target 0.16.0 .
ValueError: invalid literal for int() with base 10: '0rc1'
```
Separately, `--target 0.16` silently selects zero releases and prints `Nothing to do`, because `(0, 16, 0) <= (0, 16)` is `False`. Both fixed by normalising to a 3-tuple.

dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
vyhuholl, line 305:
```suggestion
    for release in releases:
        manual_hits += report_manual_checks(path, source, release)
        if not (release['symbols'] or release['keywords']):
            continue
        try:
            source = rewrite_module(source, str(path), release)
        except cst.ParserSyntaxError as exc:
            print(f'{path}: cannot parse, skipped: {exc}', file=sys.stderr)
            return False, manual_hits
```

vyhuholl: Two bugs found here:

**1. The codemod flags its own output.** Manual checks run on the *rewritten* source. `0.13.0` renames `ConvertedSchema` → `DumpedSchema` and also lists `dmr.openapi.views.base.DumpedSchema` under removed, so:
```
demo/views.py:1: [0.13.0] dmr.openapi.views.base.DumpedSchema was removed
demo/views.py:7: [0.13.0] dmr.openapi.views.base.DumpedSchema was removed
```
Both lines are code the tool had just written correctly. The checks describe *old* API, so they belong before that release's rewrite.
**2. One unparsable file aborts the whole run**, after earlier files are already written — a half-migrated tree plus a `libcst._exceptions.ParserSyntaxError` traceback. Any project has at least one file `libcst` won't parse.
With the suggestion above applied the same run reports `0 manual check(s)` and `broken.py: cannot parse, skipped: ...` instead of dying.

dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
```diff
@@ -0,0 +1,314 @@
...
+    source: str,
+    release: Release,
+) -> int:
+    """Print every line that matches a manual check, return the count."""
+    checks = [
+        (re.compile(check['pattern']), check['message'])
+        for check in release['manual']
+    ]
+    removed = [
+        (re.compile(re.escape(name.rsplit('.', 1)[-1]) + r'\b'), name)
```
vyhuholl:
```suggestion
        (re.compile(r'\b' + re.escape(name.rsplit('.', 1)[-1]) + r'\b'), name)
```

vyhuholl: Add leading words boundary so, for example, `dmr.test.types` won't match any word ending in `types`.

dmr/.agents/skills/dmr-upgrade/references/renames.json
```diff
@@ -0,0 +1,144 @@
...
+        {"pattern": "\\bprovide_response_specs\\b", "message": "provide_response_specs is now an instance method: replace `cls` with `self`"},
+        {"pattern": "\\bget_schema\\(|\\bget_path_item\\(", "message": "Endpoint.get_schema and Controller.get_path_item need a new `router` parameter"}
+      ]
+    },
+    {
+      "to": "0.8.0",
+      "prompt": "0.7-to-0.8.md",
+      "symbols": [
+        {"old": "dmr.response.APIRedirectError", "new": "dmr.response.RedirectTo"},
+        {"old": "dmr.throttling.backend.DjangoCache", "new": "dmr.throttling.backend.SyncDjangoCache"}
```
vyhuholl:
```suggestion
        {"old": "dmr.throttling.backends.DjangoCache", "new": "dmr.throttling.backends.SyncDjangoCache"}
```

vyhuholl: Fixing a typo. Other `new` targets in this file are resolving successfully.

dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
```diff
@@ -0,0 +1,314 @@
...
+    ) -> cst.Call:
+        """Rewrite keyword names on matching calls."""
+        callee = _callee_name(updated_node.func)
+        renames = {
+            keyword['old']: keyword['new']
+            for keyword in self._keywords
+            if keyword['callee'] == callee
+        }
+        if not renames:
+            return updated_node
```
vyhuholl: Keyword renames match on the bare callee name — `_callee_name` returns `func.attr.value` for attribute calls — so `Schema(then=...), from_python(rebuild_namespace=...)` and `FileResponseSpec(file_body=...)` get rewritten no matter which library they come from. Worth resolving through `QualifiedNameProvider` (you already depend on `libcst`) so the rename is scoped to `dmr`. No one-line fix, so leaving it as a note.

dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py
```diff
@@ -0,0 +1,314 @@
...
+    changed_files = 0
+    manual_hits = 0
+    for path in iter_python_files(args.paths):
+        changed, hits = upgrade_file(path, releases, dry_run=args.dry_run)
+        changed_files += int(changed)
+        manual_hits += hits
+        if changed:
+            print(f'{path}: rewritten')  # noqa: WPS421
+    print(  # noqa: WPS421
+        f'{changed_files} file(s) rewritten, {manual_hits} manual check(s). '
```
vyhuholl:
```suggestion
    action = 'would be rewritten' if args.dry_run else 'rewritten'
    for path in iter_python_files(args.paths):
        changed, hits = upgrade_file(path, releases, dry_run=args.dry_run)
        changed_files += int(changed)
        manual_hits += hits
        if changed:
            print(f'{path}: {action}')  # noqa: WPS421
    print(  # noqa: WPS421
        f'{changed_files} file(s) {action}, {manual_hits} manual check(s). '
```

vyhuholl:
`dmr/.agents/skills/dmr-upgrade/scripts/dmr_upgrade.py` contains 314 lines that rewrite users' source in place, with no tests. It escapes `--cov-fail-under=100` only because `dmr/.agents` has no `__init__.py`, and it escapes every type checker because mypy, pyright and pyrefly all skip dot-directories — I confirmed with mypy's own `create_source_list(['dmr'])`, which returns 0 files under `dmr/.agents`. Ruff is the only tool that sees this file.

A few golden-file tests over `renames.json` would have caught the `backends` typo and the `DumpedSchema` self-match.

dmr/.agents/skills/dmr-upgrade/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+and the type checker sees the new API.
```
vyhuholl: **The most important blocker**

Following steps 2 and 3 in order always produces `Nothing to do:` after the bump, `detect_current_version()` returns the target, so `select_releases()` returns `[]`. `--current` is presented as the exception ("when `django-modern-rest` is not importable"), but it's actually required on the happy path.
```suggestion
### 2. Record the current version, then bump the dependency

Write the installed version down first: after the bump the codemod
detects the target version and selects no releases at all.
Then update `pyproject.toml` (or the requirements file) to the target
version and install it, so the type checker sees the new API.
```
Alternative worth considering: make `--current` required in the parser, or read it from the lock file / `git` instead of installed metadata — then the docs don't have to carry the caveat.

dmr/.agents/skills/dmr-upgrade/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+
+### 3. Run the codemod
+
+From the root of the project being upgraded:
+
+```bash
+uv run --with libcst python <path-to-this-skill>/scripts/dmr_upgrade.py \
+  --target <target-version> .
```
vyhuholl:
```suggestion
uv run --with libcst python <path-to-this-skill>/scripts/dmr_upgrade.py \
  --current <version-before-the-bump> --target <target-version> .
```

dmr/.agents/skills/dmr-upgrade/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+Pass `--current <version>` when `django-modern-rest` is not importable
+in that environment and `--dry-run` to preview. The script:
```
vyhuholl:
```suggestion
Pass `--dry-run` to preview. The script:
```

docs/pages/ai/dmr-upgrade.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Running the codemod by hand
+---------------------------
+
+The codemod does not need an agent. From the root of your project:
+
+.. code-block:: bash
+
+   uv run --with libcst python \
+     .agents/skills/dmr-upgrade/scripts/dmr_upgrade.py --target 0.16.0 .
```
vyhuholl:
```suggestion
   uv run --with libcst python \
     .agents/skills/dmr-upgrade/scripts/dmr_upgrade.py \
     --current 0.14.0 --target 0.16.0 .
```

docs/pages/ai/dmr-upgrade.rst
```diff
@@ -0,0 +1,91 @@
...
+
+It detects the installed version (pass ``--current`` otherwise),
```
vyhuholl:
```suggestion
Pass ``--current`` with the version you had *before* the bump — the installed version is already the target one at this point. It
```

justfile
```diff
@@ -38,6 +38,15 @@ lint:
...
     uv run import-linter lint
+    just skills
+
+# Validate agent skills against https://agentskills.io/specification
+[group('dev')]
+skills:
+    #!/usr/bin/env bash
+    for skill in dmr/.agents/skills/*/; do
+      uv run agentskills validate "$skill"
+    done
```
vyhuholl, line 50: The loop swallows failures: no `set -e`, so the recipe's exit code is the last iteration's. A broken skill that isn't last passes silently (`bash -c 'for i in 1 2; do [ $i = 2 ]; done'` → exit 0). The workflow gets -`e` from Actions, `just lint` doesn't.
```suggestion
    #!/usr/bin/env bash
    set -euo pipefail
    for skill in dmr/.agents/skills/*/; do
      uv run agentskills validate "$skill"
    done
```
(All six skills do pass `agentskills validate` today.)

setup.cfg
```diff
@@ -28,6 +28,9 @@ ignore =
 # Excluding some directories:
 extend-exclude =
   .venv
+  # Agent skill scripts are standalone tools that run in user projects,
+  # they are checked with `ruff` and `mypy` only:
```
vyhuholl:
```suggestion
  # Agent skill scripts are standalone tools that run in user projects,
  # they are checked with `ruff` only (`mypy` skips dot-directories):
```

dmr/.agents/skills/dmr/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+
+Test with `dmr_rf` (`DMRRequestFactory`) for unit tests and `dmr_client`
+(`DMRClient`) for full-stack tests, generate payloads with `polyfactory`,
+and use `schemathesis` against the OpenAPI schema.
+
+## Quick reference
+
+| Topic | Rules | Read |
+| --- | --- | --- |
+| Controllers, `@modify` vs `@validate`, serializers, responses, components, redirects | 10 | [references/controllers.md](references/controllers.md) |
```
vyhuholl:
```suggestion
| Controllers, `@modify` vs `@validate`, serializers, responses, components, redirects | 9 | [references/controllers.md](references/controllers.md) |
```
`controllers.md` has 9 `###` rules, not 10.

dmr/.agents/skills/dmr/evals/error-response/graders/no-json-dumps.md
```diff
@@ -0,0 +1,6 @@
+---
+type: regex
+pattern: 'json\.dumps'
```
vyhuholl: The prompt literally hands the model `HttpResponse(json.dumps({'detail': 'missing'}), status=404)` and asks it to fix that. A good answer that quotes the bad line while explaining it fails this grader.
```suggestion
pattern: 'return[^\n]*json\.dumps\('
```

dmr/.agents/skills/dmr/evals/create-endpoint/graders/no-http-response.md
```diff
@@ -0,0 +1,6 @@
+---
+type: regex
+pattern: '(HttpResponse|JsonResponse)\('
```
vyhuholl: Same: "don't return `HttpResponse(...)`" in the explanation fails the check.
```suggestion
pattern: 'return\s+(HttpResponse|JsonResponse)\('
```

dmr/.agents/skills/dmr/evals/unrelated-request/prompt.md
```diff
@@ -0,0 +1,8 @@
+---
+max_turns: 10
+allowed_tools: [Read, Glob, Grep, Skill]
+tags: [routing]
```
vyhuholl: This case is "the skill must not fire", nothing to do with routing — looks like copy-paste.
```suggestion
tags: [negative]
```

vyhuholl:
Also, add this to `tool.hatch.build.targets.wheel` section of `pyproject.toml`:
```toml
only-include = [ "dmr", "dmr_pytest.py" ]
exclude = [ "dmr/.agents/skills/*/evals" ]
```
I built the wheel: all 53 skill files land in `dmr/.agents/`, which is the point — but that includes `evals/` (48K of prompts and graders) shipped to every end user.

docs/conf.py
```diff
@@ -291,6 +293,14 @@ def _get_project_meta() -> dict[str, str]:
         },
     ],
     'accent_color': 'green',
+    # "Copy page" button and "Open in ..." links for LLM chats,
+    # see `_templates/components/copy-page-button.html`:
+    'show_ai_links': True,
```
vyhuholl:
```suggestion
    # "Copy page" button and "Open in ..." links for LLM chats,
    # see `_templates/components/copy-page-button.html`
    # (`show_ai_links` is already `True` in Shibuya's `theme.conf`):
```

docs/conf.py
```diff
@@ -315,7 +325,30 @@ def _get_project_meta() -> dict[str, str]:
...
+llms_txt_summary = (
+    f'Documentation for django-modern-rest version {release}. '
+    'Every page listed below is also available as Markdown: '
+    'replace `.html` with `.md` in its URL. '
+    'The complete documentation in one file is `llms-full.txt` '
+    'next to this file.'
+)
+markdown_anchor_sections = True
+markdown_http_base = html_baseurl.rstrip('/')
+markdown_uri_doc_suffix = '.html'
```
vyhuholl: Links inside the `.md` twins point at `.html`, while `llms_txt_summary` right above tells the reader to swap `.html` for `.md`. An agent that follows a link leaves the Markdown set. Intentional?
```suggestion
markdown_uri_doc_suffix = '.md'
```

vyhuholl (dismissed):
Reviewed the whole diff and ran the new code (codemod, tests, ruff, `agentskills validate`, wheel build). Structural direction is sound — skills in the package, validator + unit tests + evals, docs Markdown twins. The problems are concentrated in the codemod and in a few CI/doc details.

vyhuholl (commented):
Reviewed the whole diff and ran the new code (codemod, tests, ruff, `agentskills validate`, wheel build). Structural direction is sound — skills in the package, validator + unit tests + evals, docs Markdown twins. The problems are concentrated in the codemod and in a few CI/doc details.

sobolevn:
Feel free to commit right into this branch, because I am a bit overwhelmed with CSRF and auth fixes :)

vyhuholl:
Fixed everything and added new tests in https://github.com/wemake-services/django-modern-rest/commit/876eb8ae6916064490cd896c135d6be016e02bd7

sobolevn (commented):
let's remove all codemods and tests for them :)
I don't think that this is really helpful / not buggy.

vyhuholl:
> let's remove all codemods and tests for them :) I don't think that this is really helpful / not buggy.

Removed codemods, their tests and the `libcst` dev dependency in https://github.com/wemake-services/django-modern-rest/commit/354a7c10afe6435ec1bd30e9db8ffaf00d999d3f

pyproject.toml
```diff
@@ -475,6 +479,16 @@ ignore_missing_imports = true
 module = [ "allauth.*" ]
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
```
sobolevn: I propose to use inline type ignores for just several imports in a single file

.readthedocs.yml
```diff
@@ -4,6 +4,15 @@ build:
...
   tools:
     python: "3.12"
+  jobs:
+    post_build:
+      # Markdown twin of every page for LLMs and agents, served next
+      # to the HTML page: `pages/routing.md` for `pages/routing.html`.
+      # Reuses the doctrees of the HTML build, so examples are not re-run.
+      # `--no-sync` keeps the environment installed above: a plain `uv run`
+      # would re-sync it to the default groups and drop the `docs` one.
+      - cd docs && uv run --no-sync python -m sphinx -M markdown . _build -W
```
sobolevn: Can we somehow configure what builds do we run with standard DSL? It does not look correct to run manual builds in RTD.

.gitignore
```diff
@@ -267,3 +267,6 @@ wheelhouse/
 # AI slop:
 .claude/
 .codex/
+
+# Claude Code plugin evals write their reports here:
+dmr/.agents/skills/*/evals/results/
```
sobolevn: This needs to be removed as well.

docs/tools/sphinx_ext/markdown.py
```diff
@@ -0,0 +1,149 @@
+"""
```
sobolevn: This seems overly complex. Why do we have it? Are we sure that there's no default md builder that "just works"?

.github/lychee.toml
```diff
@@ -62,4 +62,12 @@ exclude = [
   # Telegram terminates automated requests even when public channel links are
   # reachable in regular browsers.
   "^https://t\\.me/.*$",
+  # We link our own agent skills and their documentation from `CHANGELOG.md`
```
sobolevn: It is fine that this job will fail here once. But, we don't want this configuration to stay and allow failures in the future.

sobolevn (commented):
I also propose to remove:
- new CI job
- evals
- old versions upgrade notes, let's keep only 3 last ones
- `tests/test_unit/test_skills/test_skills.py` and `tests/test_unit/test_skills/__init__.py`

PR #1566, opened 2026-09-19
Add `security` OpenAPI requirements

### LINKED ISSUE #1499: Fix unused `security` field in `EndpointMetadata` and `_BasePayload`
We don't use these two things:
1. `security` field documentation here: https://github.com/wemake-services/django-modern-rest/blob/f822e35bd2db672dcc43fbd4967a22835a616971/dmr/metadata.py#L429-L430
2. https://github.com/wemake-services/django-modern-rest/blob/f822e35bd2db672dcc43fbd4967a22835a616971/dmr/validation/payload.py#L44

However, I don't think that this is right. Users must have an option to configure this directly. 
Let's create this rule:
- Any user provided `security` objects must be saved to the metadata
- It must be merged with explicit `auth=` providers

It might be useful when people define some external security mechanims, for example that work on HTTP proxy level, or in other microservices.

New snapshot test is required for this change.

nikolaysm:
## What changes

`security` can now be declared on three levels, the same way `auth` is:
`Settings.security`, the `Controller.security` attribute, and `security=`
on `@modify` / `@validate`. The levels are merged endpoint → controller →
settings, duplicates are dropped, and the result is stored on
`EndpointMetadata.security` and added after the requirements generated
from `auth`, so a client can satisfy any of them.

`security=None` disables the requirements for an endpoint or a whole
controller (a controller's `None` also drops what its endpoints declare),
`security=[]` adds nothing on its level. Schemes used in `security` are not
registered automatically — they must be declared in `components` of
`OpenAPIConfig` — and they must not reuse the names that `auth` generates,
`EndpointMetadataError` is raised during the schema generation when they do.

Docs: a new "Customizing security" section with a runnable example, the new
setting in `configuration.rst`, and the note that changing an `auth`-generated
schema is done by subclassing the auth class.

## Follow-ups

I think two follow-up tickets are worth creating:

1. `build_schema` does not reject scheme names that are missing from
   `components.securitySchemes`, and the spec validator does not catch it
   either, so a typo silently ships an invalid document.
2. `ConfigMerger` raises a rather cryptic
   `ValueError: Trying to merge components with shared keys` when a user
   declares a scheme that an auth also registers.

## AI Policy

## Checklist

## Related issues

Closes #1499


nikolaysm:
Hi @sobolevn, I've pushed a fix for the failing lint job and rebased on top of master, so the conflict is gone too. Could you approve the workflow run one more time, please?

